PR #222, opened 2018-01-31
Upgrade CodeCompass to use LLVM 6.0 and fix as many warnings as possible

whisperity:
Updated the build information to download LLVM 6, and also fixed some various missing package issues that arise when the user tries to install on a bare Ubuntu 16.04(.1) LTS machine.

Upgraded the client code to use the interface available in LLVM 6.0. This required subtle changes as to how 3 or 4 functions are invoked.

Fixed as many warnings as possible reported by the compiler, such as missing `override` keywords.

`-Wreturn-type-c-linkage` is a Clang-specific warning. This is not appropriate for us as we don't call the library initialisers from pure C code, so it has been quelched.

Updated the blocks of `extern "C" {}` to have the same code layout.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.9)
+cmake_minimum_required(VERSION 3.4.3)
 project(CodeCompass)
 
 # Common config variables and settings
--- a/doc/deps.md
+++ b/doc/deps.md
@@ -3,18 +3,19 @@ We build CodeCompass under Linux. It is recommended to use a 64-bit operating
 system.
 
 # Dependencies
-CodeCompass uses some 3rd party dependencies. These can be installed from the
+CodeCompass uses some third-party dependencies. These can be installed from the
 official repository of the given Linux distribution. In Ubuntu 16.04 LTS the
 following packages are necessary for building CodeCompass:
 
 - **git**: For fetching and managing CodeCompass source code.
-- **cmake**: For building CodeCompass.
-- **make**: For building CodeCompass.
-- **g++**: For compiling CodeCompass. We use C++14 features, so the compiler
-  has to be able to compile it.
+- **cmake** and **make**: For building CodeCompass.
+- **g++**: For compiling CodeCompass. A version which supports C++14 features
+  is required. (Alternatively, you can compile with Clang.)
 - **libboost-all-dev**: Boost can be used during the development.
-- **llvm-3.8-dev**, **libclang-3.8-dev**: C++ parser uses LLVM/Clang for
-  parsing the source code. Version 3.8 is required.
+- **llvm-6.0**, **clang-6.0**: For compiling CodeCompass with Clang instead of
+  G++.
+- **llvm-6.0-dev**, **libclang-6.0-dev**: C++ parser uses LLVM/Clang for
+  parsing the source code. Version 6.0 or newer is required.
   ***See [Known issues](#known-issues)!***
 - **odb**, **libodb-dev**, **libodb-sqlite-dev**, **libodb-pgsql-dev**: For
   persistence ODB can be used which is an Object Relation Mapping (ORM) system.
@@ -22,12 +23,11 @@ following packages are necessary for building CodeCompass:
 - **openjdk-8-jdk**: For search parsing CodeCompass uses an indexer written in
   Java.
 - **libssl-dev**: OpenSSL libs are required by Thrift.
-- **libsqlite3-dev**: SQLite can be used for persistence. This is needed only if
-  you choose SQLite as a database system.
 - **libgraphviz-dev**: GraphViz is used for generating diagram visualitaions.
 - **libmagic-dev**: For detecting file types.
 - **libgit2-dev**: For compiling Git plugin in CodeCompass.
-- **npm**: For handling JavaScript dependencies for CodeCompass web GUI.
+- **npm** and **nodejs-legacy**: For handling JavaScript dependencies for
+  CodeCompass web GUI.
 - **ctags**: For search parsing.
 - **libgtest-dev**: For testing CodeCompass.
   ***See [Known issues](#known-issues)!***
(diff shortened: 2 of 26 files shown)


whisperity:
**PLEASE DO NOT MERGE**

I have an extra idea to put into this patch quelching more warnings!

whisperity:
@csordasmarton Okay, you can remove the `WIP` now I think.

whisperity:
Actually... it looks like I can't get the binaries to actually load.

```
  what():  Failed to load "/large/whisperity/CodeCompass/CodeCompass-install/bin/../lib/parserplugin/libsearchparser.so": /large/whisperity/CodeCompass/CodeCompass-install/bin/../lib/parserplugin/libsearchparser.so: undefined symbol: _ZN2cc6parser13SourceManager7getFileERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE

  what():  Failed to load "/large/whisperity/CodeCompass/CodeCompass-install/bin/../lib/serviceplugin/libmetricsservice.so": /large/whisperity/CodeCompass/CodeCompass-install/bin/../lib/serviceplugin/libmetricsservice.so: undefined symbol: mg_send_header
```

Okay, this needs more looking at.

whisperity:
Okay, this should properly work now. As an addition, the current code compiles and runs without a hiccup if done with Clang 6.0!

@csordasmarton *HOPEFULLY* no more toggle WIP play here...

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.9)
+cmake_minimum_required(VERSION 3.4.3)
```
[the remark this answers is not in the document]
whisperity: LLVM Version 6 requires this same version of CMake, so the user won't be able to compile their environment anyways if their CMake is older than this.

doc/deps.md
```diff
@@ -109,20 +123,28 @@ cd libodb-2.4.0
...
-# If you use SQLite
+#
+# If you use SQLite:
+#
 wget http://www.codesynthesis.com/download/odb/2.4/libodb-sqlite-2.4.0.tar.gz
 tar -xvf libodb-sqlite-2.4.0.tar.gz
 cd libodb-sqlite-2.4.0
-./configure --prefix=<odb_install_dir>
+./configure CPPFLAGS="-I<odb_install_dir>/include" \
+  LDFLAGS="-L<odb_install_dir>/lib" \
```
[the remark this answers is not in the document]
whisperity: Without these flags, `configure` will say `libodb` is not found, and will refuse to create a makefile. This fixes this issue.

(Steps: download libodb, configure, make and install it to a prefix, and then try to configure libodb-pgsql to the same install prefix.)

Boris said this fix on the ODB mailing list with regards to MinGW, but the same fix applies on Ubuntu.

The install guide was done on a brand new installed Ubuntu 16.04.3 LTS.

whisperity:
Well, back to the `WIP` status for now. I'm getting these errors when parsing Xerces:

```
[INFO] (1/978) Parsing /home/whisperity/Downloads/xerces-c-3.2.1/src/xercesc/util/Base64.cpp
CodeCompass_parser: /large/whisperity/LLVM/llvm/tools/clang/include/clang/Basic/SourceLocation.h:297: const char *clang::PresumedLoc::getFilename() const: Ass
ertion `isValid()' failed.
```

whisperity:
Alright, I think I've got it now. The macro handling wasn't only changed in the interface, it was changed in behaviour too. Now calling `getFilename()` on a `PresumedLocation` asserts if there isn't a file behind it, as opposed to just returning a null pointer. With this change, CodeCompass now works flawlessly... or rather with only as many flaws as with LLVM 3.8.

doc/deps.md
```diff
@@ -3,42 +3,44 @@ We build CodeCompass under Linux. It is recommended to use a 64-bit operating
...
 - **openjdk-8-jdk**: For search parsing CodeCompass uses an indexer written in
   Java.
 - **libssl-dev**: OpenSSL libs are required by Thrift.
-- **libsqlite3-dev**: SQLite can be used for persistence. This is needed only if
-  you choose SQLite as a database system.
 - **libgraphviz-dev**: GraphViz is used for generating diagram visualitaions.
 - **libmagic-dev**: For detecting file types.
 - **libgit2-dev**: For compiling Git plugin in CodeCompass.
-- **npm**: For handling JavaScript dependencies for CodeCompass web GUI.
+- **npm** and **nodejs-legacy**: For handling JavaScript dependencies for
```
bruntib: What is `nodejs-legacy`? Why is it needed?

doc/deps.md
```diff
@@ -53,40 +55,52 @@ sudo apt-get install byacc flex
...
+# Python, PHP and such other Thrift builds are NOT required by CodeCompass.
+
 make install
 ```
 
 ## Known issues
 - In Ubuntu 16.04 LTS the LLVM/Clang has some packaging issues, i.e. some libs
   are searched in `/usr/lib` however the package has these in
-  `/usr/lib/llvm-3.8/lib` (see http://stackoverflow.com/questions/38171543/error-when-using-cmake-with-llvm).
+  `/usr/lib/llvm-3.8/lib` (see
```
bruntib: LLVM 3.8 related packaging issues are not relevant any more.

doc/deps.md
```diff
@@ -109,28 +123,34 @@ cd libodb-2.4.0
...
 cd libodb-pgsql-2.4.0
-./configure --prefix=<odb_install_dir>
+./configure --prefix=<odb_install_dir> \
+  --with-libodb="$(readlink -f ../libodb-2.4.0)"
 make install
 cd ..
 
-sudo apt-get install gcc-<version>-plugin-dev
-sudo apt-get install libcutl-dev
+sudo apt-get install gcc-<version>-plugin-dev libcutl-dev libexpat1-dev
```
bruntib: Why is `libexpat1-dev` needed?

whisperity: `libexpat1-dev` is the header/development package for `libexpat1` which is a dependency of `libcutl` which is the header/development package for `libcutl-dev`. Without installing this, some compilation fails due to missing headers or linker errors.

Most likely, considering this install command has been added to this line, it's the ODB command-line executable that cannot compile without it.

doc/deps.md
whisperity: They are relevant in the sense that there is no indication that the packaging will change once LLVM 6.0's release version hits the upstream repositories, if they do.

whisperity:
It looks like a patch mentioned in previous comments went missing in a force push alongside the lengthy review process. Do **not** be surprised, CodeCompass isn't able to parse anything right now that contains preprocessor directives.

PR #236, opened 2018-04-15
Fixing ubuntu packaged LLVM compilation

dutow:
Some libraries contain global variables registering command line options.
Accidentally linking to these libraries multiple times results in llvm errors:

: CommandLine Error: Option '<some-option>` registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options

As these libraries aren't used directly by the parser, bur are only used by clang, they can be safely removed.

With these changes, codecompass can be built and run with the Clang binaries from the llvm apt repository / from the Ubuntu repository in case of bionic.

See docker repository for example: https://github.com/dutow/CodeCompass-Docker/commit/8b693880140809a19068e205eee591beb62ee999

### DIFF
--- a/plugins/cpp/parser/CMakeLists.txt
+++ b/plugins/cpp/parser/CMakeLists.txt
@@ -1,7 +1,5 @@
 find_package(Clang REQUIRED CONFIG)
 
-llvm_map_components_to_libnames(llvm_libs support core irreader)
-
 include_directories(
   include
   ${CMAKE_SOURCE_DIR}/parser/include
@@ -41,9 +39,7 @@ target_link_libraries(cppparser
   clangLex
   clangBasic
   clang
-  LLVMMCParser
-  LLVMOption
-  ${llvm_libs})
+  )
 
 target_compile_options(cppparser PUBLIC -Wno-unknown-pragmas)
 
--- a/plugins/cpp_reparse/service/CMakeLists.txt
+++ b/plugins/cpp_reparse/service/CMakeLists.txt
@@ -1,7 +1,5 @@
 find_package(Clang REQUIRED CONFIG)
 
-llvm_map_components_to_libnames(llvm_libs support core irreader)
-
 include_directories(
   include
   gen-cpp
@@ -75,7 +73,7 @@ target_link_libraries(cppreparseservice
   clangBasic
   clangAST
   clang
-  ${llvm_libs})
+  )
 
 install(TARGETS cppreparseservice DESTINATION ${INSTALL_SERVICE_DIR})
 install_js_thrift(gen-js/)


whisperity:
Is there any functionality loss (maybe something that's not immediately apparent?) by not compiling with RTTI?

dutow:
I'm still trying to check if there's any difference in the output  - so far I don't see any.

(also, even with these changes, I think it's still build-able with a dynamically linked clang, but I'm checking that too now. If it doesn't, I'll fix that and update the PR)

whisperity:
Also, make sure to test dynamic linkage with `-DBUILD_SHARED_LIBS=ON` in LLVM's cmake invocation.

Self-compiling Clang7 **without RTTI** and then using that with the current master CodeCompass results in the following errors when I attempt to start the parser:

    lib/parserplugin/libcppparser.so: undefined symbol: _ZTIN5clang7tooling21FrontendActionFactoryE

dutow:
You are right, RTTI is required.

Turns out that the clang/llvm package has RTTI enabled:

```
/usr/lib/llvm-6.0/cmake/LLVMConfig.cmake
167:set(LLVM_ENABLE_RTTI ON)
```

So:
* manually built clang requires RTTI=ON
* ubuntu packaged clang has RTTI ON too
* for the ubuntu packages, the libraries I removed cause linkage errors, most likely because of some other flag/setting.

whisperity:
Oh, nice. So it means we can roll without manually building Clang, and no functionality is lost at runtime?

Does Ubuntu 16.04 package manager have the LLVM code in their upstream package list?

We'll try a virgin computer with these changes and then merge this patch.

(Although it'll most likely be broken almost instantly, as my upcoming changes depend on LLVM 7...)

dutow:
Yes, exactly. We have to symlink two files manually, as the ClangConfig.cmake is placed in an incorrect location, but I'll submit a patch for that to the llvm-toolchain repository.

16.04 only has llvm/clang4, 18.04 has it up to 6 by default.
And there's also the llvm ppa: http://apt.llvm.org/
Which provides even llvm 7 nightly builds as deb packages, most projects requiring modern clang on older ubuntus depend on that.

whisperity:
@dutow Once you have the patch ready, please mention its link here in a comment.

mcserep:
Maybe we can warm this pull request up, since Ubuntu Bionic has LLVM 7 / Clang 7 in its standard package repository now and it would be nice if CodeCompass could be compiled without the manual compilation of these dependencies on Ubuntu 18.04.

I have tested @dutow 's changes with my manually compiled LLVM/Clang and it seems to work correctly. I will also test it with LLVM from the Ubuntu package repository later.

mcserep:
I have tested and with the appropriate modifications in `cpp_reparse` (added to the PR) is works properly with LLVM7 and Clang7 installed from the Ubuntu package repository under bionic.

PR #242, opened 2018-04-24
Fix 'corethriftjava' failing to build when JAR is compiled earlier than Thrift generation

whisperity:
According to [CMake's documentation](https://cmake.org/cmake/help/v3.0/command/add_custom_command.html), the `add_custom_command`:

> This defines a command to generate specified OUTPUT file(s). A target created in the same directory (CMakeLists.txt file) that specifies any output of the custom command as a source file is given a rule to generate the file using the command at build time.

There is no indication anywhere that `OUTPUT` can be set to a folder and that it will be meaningful.

In certain environments (seemingly dependent of system language options or collation?), especially when multithreaded build is done, building CodeCompass can fail:

```
make[2]: *** No rule to make target '/mnt/storage/jenkins/workspace/Build master CodeCompass/Source/service/project/gen-java/cc/service/core/Timeout.java', needed by 'service/project/CMakeFiles/corethriftjava.dir/java_compiled_corethriftjava'.  Stop.
CMakeFiles/Makefile2:2040: recipe for target 'service/project/CMakeFiles/corethriftjava.dir/all' failed
make[1]: *** [service/project/CMakeFiles/corethriftjava.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

This is because CMake did not know of the dependency between the Thrift-generated Java file and the fact that this class is to be packaged by `add_jar`.

### DIFF
--- a/service/project/CMakeLists.txt
+++ b/service/project/CMakeLists.txt
@@ -9,6 +9,8 @@ include_directories(SYSTEM
   ${THRIFT_LIBTHRIFT_INCLUDE_DIRS}
   ${ODB_INCLUDE_DIRS})
 
+# The folders here are added so that executing `make clean` removes all
+# generated files.
 add_custom_command(
   OUTPUT
     ${CMAKE_CURRENT_SOURCE_DIR}/gen-cpp/common_constants.cpp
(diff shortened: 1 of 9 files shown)


service/project/CMakeLists.txt
```diff
@@ -67,6 +69,8 @@ add_jar(corethriftjava
   ${CMAKE_CURRENT_SOURCE_DIR}/gen-java/cc/service/core/Timeout.java
   OUTPUT_NAME corethrift)
 
+add_dependencies(corethriftjava commonthrift)
```
whisperity: I found an easier solution. This will make the Java compilation wait until the C++ compilation of the Thrift types is done, but is *much easier* to express in the Makefile.

Issue #225, opened 2018-02-02
Keepalive script unreliable(?) in a headless environment


whisperity:
If CodeCompass Web server is started with `daemon` (the Unix standard start-stop-daemon wrapper), the `keepalive` script automatically exits after the process is started and becomes a zombie:

![image](https://user-images.githubusercontent.com/1969470/35733542-19026dbc-081e-11e8-8fb2-102dc87ffa37.png)

This results in the server crashing for any reason, be it an internal core or an external signal, the keepalive script does not anything, just gets reaped alongside the invoker `daemon` call.

**This might affect the parser too since it's the same script!**

Unfortunately, this zombie state of the shell that executes the keepalive script prevents the otherwise standard-looking `--respawn` option of `daemon`.

whisperity:
Actually, it looks rather situational. Sometimes the keepalive script actually stays running, sometimes it zombies out.

whisperity:
It should also not infinitely flood stuff like "cannot load shared library" as obviously such things won't get fixed by infinitely retrying. (And it can also chew away server storage space resulting in the entire thing tumbling down.)

bruntib:
We don't have a keepalive script anymore.

PR #243, opened 2018-04-24
Make keepalive script also trap SIGTERM

whisperity:
Certain background process wrappers, such as `daemon(1)` (and by default, the `kill` command) send `SIGTERM` (15) as opposed to `SIGINT` (2, generated by a `Ctrl-C` in the runner shell). Without trapping on this signal, only the keepalive script is terminated, which zombifies the underlying shell and also prevents (without manual intervention or looking at the process table...) killing the CodeCompass server.

> Related to #225 but does not necessarily fix the issue.

### DIFF
--- a/scripts/keepalive
+++ b/scripts/keepalive
@@ -8,7 +8,7 @@ ending()
   kill $pid
 }
 
-trap ending SIGINT
+trap ending SIGINT SIGTERM
 
 ulimit -c unlimited &> /dev/null
 


bruntib:
The main goal of this script is to set the environment for the 3rd parties. Keeping CodeCompass alive should be done by more appropriate tools, like `daemon` or something. This current temporary solution restarts CodeCompass for any non-zero exit code which also happens for providing bad command line arguments which makes CodeCompass print the error message in an infinite loop. But anyway CodeCompass shouldn't fail :smile: Originally this was a safety action for the server.

PR #245, opened 2018-05-04
Massively speed up navigation by not sending source text for every AST node

### LINKED ISSUE #231: Clicking into a function in C++ code automatically sends every reference's file's FULL source code in the background
If I click on a random symbol and then click into the whitespace area in the enclosing function, compass auto-selects the function itself, indicated by the orange highlight.

In the background, a `getReferences` call is issued, which sends every reference to the function I'm currently clicking in. These references include **whole source files**. (Check with `--loglevel debug` which shows every Thrift communication in the output.)

This can make execution really slow (and waste a lot of data sent through the network!) if I'm clicking in a function that's widely used. Every click, even if I click from whitespace to whitespace results in **massive data** being transmitted.

whisperity:
> Fixes #231.

As I can see from the source code, there was only **one** location where the source code was actually used (in CodeBites), so it was really, really superfluous to send the source text every time.

Conceptually, the abstract language information system had to be modified. A new API call is introduced to query the full source text, if and when needed.

While I don't have any precise measurings for this on large projects, even on a small synthesised example (a few types and whatnot) the whole interface feels much, much faster. This is especially prominent for large codebases like LLVM/Clang where the removal of sending **(almost) entire headers** (for types) in the API really helps.

### DIFF
--- a/plugins/cpp/service/src/cppservice.cpp
+++ b/plugins/cpp/service/src/cppservice.cpp
@@ -71,6 +71,9 @@ namespace
     {
     }
 
+    /**
+     * Returns the Thrift object for this C++ AST node.
+     */
     cc::service::language::AstNodeInfo operator()(
       const cc::model::CppAstNode& astNode_)
     {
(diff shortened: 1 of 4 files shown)


plugins/cpp/service/src/cppservice.cpp
```diff
@@ -88,24 +91,32 @@ namespace
...
       if (it != _tags.end())
         ret.__set_tags(it->second);
 
       return ret;
     }
 
+    /**
+     * Returns the source code text for this AST node.
+     */
+    std::string getSourceText(
```
bruntib: I think we don't need this function here. Its implementation could go to the place of its usage. (P.s.: Watch the indentation ;)

PR #249, opened 2018-05-08
Reparse: Allow generation of colourful HTML for individual Decls' or statements in the code

whisperity:
> Depends on #248. The patch needs to be rebased once that is merged.

This patch extends the previous one with the ability to right-click on individual AST nodes in the source text, and generate the HTML for the subtree whose root is the clicked AST node.

The same view is used as in the previous patch, just the tree is smaller.


plugins/cpp_reparse/service/src/astcache.h
```diff
@@ -0,0 +1,118 @@
...
+  std::shared_ptr<clang::ASTUnit> getAST(const core::FileId& id_);
+
+  /**
+   * Store the AST for the given file in the cache. The AST Cache takes
+   * ownership over the ASTUnit. The method returns the shared pointer that
+   * wraps the ASTUnit argument given.
+   */
+  std::shared_ptr<clang::ASTUnit> storeAST(
+    const core::FileId& id_,
+    std::unique_ptr<clang::ASTUnit>&& AST_);
```
bruntib: Why are `std::unique_ptr`s passed by `&&`?

doc/deps.md
```diff
@@ -85,22 +86,19 @@ make install
...
 ```bash
-wget http://github.com/llvm-mirror/llvm/archive/release_60.zip
-unzip release_60.zip
-rm release_60.zip
-cd llvm-release_60/tools
-wget http://github.com/llvm-mirror/clang/archive/release_60.zip
-unzip release_60.zip
-mv clang-release_60 clang
-rm release_60.zip
+git clone http://github.com/llvm-mirror/llvm.git --depth 25 llvm
```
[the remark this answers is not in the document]
whisperity: Will be fixed when this patch is rebased on its dependency. I think it was to ensure users can have a bit of wiggle space if the current `master` LLVM contains a build error?

plugins/cpp_reparse/service/include/service/cppreparseservice.h
```diff
@@ -0,0 +1,66 @@
...
+namespace language
+{
+
+class CppReparseServiceHandler: virtual public CppReparseServiceIf
+{
+public:
+  CppReparseServiceHandler(
+    std::shared_ptr<odb::database> db_,
+    std::shared_ptr<std::string> datadir_,
+    const boost::program_options::variables_map& config_
```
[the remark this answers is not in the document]
whisperity: Will be fixed at rebase too.

plugins/cpp_reparse/service/src/astcache.h
whisperity: `std::unique_ptr` cannot be copy constructed, thus cannot be passed by value.

Clang gives you (the client code) unique-ptrs of ASTs. The extra requirement for the user to **move the AST *into* the method** is to signify that `storeAST` *destroys* the unique ownership and internally transforms it into a `shared_ptr` (see `ASTCacheEntry`). Thus, methods that create the ASTs explicitly release the unique_ptr they had in their stack by turning the unique-ptr on the stack into a *moved from* state.

Simply passing the `unique_ptr` by reference wouldn't make this rule explicit enough.

(Later code uses the cache's get method to obtain the same AST back, but that obtained AST is through a shared pointer to ensure ownership is kept as long as the cache can store the AST or some thread is using it.)

bruntib: Yes, the idea behind the mechanism is understandable with ownership, shared pointers, etc. But the parameter passing method is not justified. For moving a `std::unique_ptr` you don't need `&&` as formal parameter, since it has a move constructor:
```.cpp
struct S {};

void f(std::unique_ptr<S>) {}

void g()
{
  std::unique_ptr<S> p;
  f(std::move(p));
}
```

whisperity: I'll update the previous patch, as these things were brought in by that patch, not this one. Unfortunately GitHub can't cherry patches incrementally.

whisperity: Has been fixed in #248.

plugins/cpp_reparse/service/src/astnodelocator.h
```diff
@@ -0,0 +1,88 @@
...
+      // This is merely an optimisation so that no extra calculations are made.
+      // If the start line of the Decl does not match the filter, it can't be
+      // the ASTNode we are searching for.
+      clang::SourceLocation start = node_->getLocStart();
+      if (_context.getSourceManager().getPresumedLineNumber(start) !=
+          _dbNode->location.range.start.line)
+        return false;
+
+      clang::PresumedLoc presumedLoc =
+        _context.getSourceManager().getPresumedLoc(start);
```
bruntib: Isn't it slightly faster to query this presumed location instead of the presumed line number in the previous "if" statement? This way there was only one query for the location, and that contains the line number anyway.

plugins/cpp_reparse/service/src/cppreparseservice.cpp
```diff
@@ -92,6 +92,58 @@ void CppReparseServiceHandler::getAsHTML(
...
+  _transaction([&, this](){
+    astNode = _db->query_one<model::CppAstNode>(
+      AstQuery::id == std::stoull(nodeId_));
+    astNode->location.file.load();
+  });
+
+  if (!astNode)
+  {
+    LOG(error) << "No location file found for AST node #" << nodeId_;
+    return_ = "Invalid AST node given, no related file could be found.";
```
bruntib: We have a `common.InvalidId` exception in `common.thrift` for signaling this error.

plugins/cpp_reparse/service/src/cppreparseservice.cpp
```diff
@@ -92,6 +92,58 @@ void CppReparseServiceHandler::getAsHTML(
...
+  model::CppAstNodePtr astNode;
+  _transaction([&, this](){
+    astNode = _db->query_one<model::CppAstNode>(
+      AstQuery::id == std::stoull(nodeId_));
+    astNode->location.file.load();
+  });
+
+  if (!astNode)
+  {
+    LOG(error) << "No location file found for AST node #" << nodeId_;
```
bruntib: This should be a debug or warning level log. Error is given if it makes CodeCompass stop running.

plugins/cpp_reparse/service/src/cppreparseservice.cpp
whisperity: Handling errors, headers and such is added as a `TODO` in the Thrift file. It will be done in a following patch some day.

plugins/cpp_reparse/service/src/cppreparseservice.cpp
whisperity: Shouldn't `CRITICAL` be used for that, though? I'll change this to a warning nonetheless.

bruntib: We don't have critical level.

plugins/cpp_reparse/service/src/asthtml.cpp
```diff
@@ -146,6 +149,62 @@ class ColourTag : public UnformattedString
...
+    public RecursiveASTVisitor<ASTSubtreeHTMLPrinter>
+{
+  typedef RecursiveASTVisitor<ASTSubtreeHTMLPrinter> Base;
+
+public:
+  ASTSubtreeHTMLPrinter(std::unique_ptr<raw_ostream> out_,
+                        ASTContext& context_,
+                        model::CppAstNodePtr node_)
+    : _out(*out_),
+      _ownedOut(std::move(out_)),
```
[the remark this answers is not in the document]
whisperity: Left-behind from the times when this class was inheriting from some LLVM type.

PR #250, opened 2018-05-08
Reparse: Create a Syntax InfoTree that allows browsing the AST on-demand by the user

whisperity:
> Depends on #249. Patch needs to be rebased once it is merged.

This patch introduces a new *Info Tree* implementation which allows the user to browse the AST in the accordion. This can be used on whole translation units (by right-clicking in the file browser), or to a particular AST Node by clicking in the source code view (similar to how HTML generation is added in #248 and #249).

For each AST node, the tree is traversed quickly and only for the necessary parts, making the iteration on the tree a good user experience. As the user goes inwards on the tree, fewer nodes are visited each time (this is because the nodes are visited in a preorder labelling manner to make sure individual inner nodes can be located over the API, even if the AST is eliminated from the cache).

The InfoTree that is sent to the user persist through cache flushes, i.e. if the user is viewing a subtree and the AST for this translation unit is released from cache, at the next query, it is re-parsed (and cached), and the traversal of the tree identifies the node the user was viewing.

For each node, a "dummy" output is generated as an "info" :information_source:. This is a naive dump of the "first" line for the particular node in the AST HTML output, but without colouring.

This is to be extended *later* with tailored information for the particular node based on its type.

An example over a small `main()` function:

reparse 3

### DIFF
--- a/plugins/cpp/webgui/js/cppInfoTree.js
+++ b/plugins/cpp/webgui/js/cppInfoTree.js
@@ -292,6 +292,7 @@ function (model, viewHandler, util) {
   }
 
   var cppInfoTree = {
+    id : 'cpp-infotree',
     render : function (elementInfo) {
       var ret = [];
 
--- a/plugins/cpp_reparse/service/CMakeLists.txt
+++ b/plugins/cpp_reparse/service/CMakeLists.txt
@@ -58,6 +58,7 @@ add_library(cppreparseservice SHARED
   src/astcache.cpp
   src/asthtml.cpp
   src/databasefilesystem.cpp
+  src/infotree.cpp
   src/reparser.cpp)
 
 target_compile_options(cppreparseservice PUBLIC -Wno-unknown-pragmas)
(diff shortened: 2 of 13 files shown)


whisperity:
@bruntib There were some weird issues with compiling the reparse service with G++ 5.4 (Ubuntu 16.04 stock supplied), because the "old" G++'s type deduction seems to be missing a few things that Clang could handle. The fix has been added to this PR.

bruntib:
I'd like to anticipate that I haven't seen the implementation so far, and my opinion is based only on the screenshot.

I don't think that InfoTree would be a proper place for displaying the AST. Personally I think it is not really usable, searchable and useful from code comprehension point of view. Not to mention that even in the most simple cases the AST can be quite big. So the AST display should stay in the central view as in the preceding pull requests. In CodeMirror the folding feature can collapse the irrelevant parts of the tree, like the body of a function, loop, etc. https://codemirror.net/demo/folding.html

As far as I understand the main goal was that the user should be able to see the code and its AST at the same time. I think this should rather be solved with a split view like in case of diagrams. The already existing AST displaying central module should be used directly in this split view, like Text module is used directly in the diagram module.

whisperity:
There are multiple issues with building CodeCompass as of now so I'm withdrawing the latest commit and instead will put together a patch that will hopefully fix every issue with building. (E.g. release builds with GCC even with the "move to Optional<T> return value" fix applied crashes on linking a boost symbol.)

PR #254, opened 2018-05-14
Show size of C++ types in the InfoTree

### LINKED ISSUE #240: Show the size of types in the Info Tree with respect to the environmental things
> via @Xazax-hun:
It would be great to display the size for classes and structs in the info tree.

Later it would be great to be able to use it in some metrics, e.g find the biggest data structures.

There was an implementation for this in the Earhart release, which *might have never been merged* to the `Earhart` branch in the current repository.

whisperity:
> Closes #240.

**Applies to the C++ plugin only.** 

Store the type's size and alignment in the database. This is obtained from Clang at parse time, so this value respects compilation options.

Four new icons are added to the icon set at code points `e071` &ndash; `e074`. The latter two are used for size and alignment, respectively.

image

An alignment of `1` means no special alignment (every memory address is divisible by 1). To easy the results on the API, this "non-special" value is not sent.

(Fun fact: Xerces contains a type named `XMLReader` which is around 160000 bytes large. It's instances contain a number of arrays in themselves.)


plugins/cpp/service/src/cppservice.cpp
```diff
@@ -307,6 +307,11 @@ void CppServiceHandler::getProperties(
           return_["Abstract type"] = "true";
         if (type.isPOD)
           return_["POD type"] = "true";
+        if (type.size > 0)
+          return_["Size"] = std::to_string(type.size);
```
csordasmarton: It should be documented somewhere that the unit of this value is in bytes (e.g.: someone writes a new client should know this information) or you can set the unit here not in the client side.
```cpp
return_["Size"] = std::to_string(type.size) + "bytes";
```

plugins/cpp/service/src/cppservice.cpp
```diff
@@ -307,6 +307,11 @@ void CppServiceHandler::getProperties(
...
+        if (type.alignment > 1)
+          // Only send alignment if it is meaningful alignment.
+          return_["Alignment"] = std::to_string(type.alignment);
```
csordasmarton: Same as above.

plugins/cpp/model/include/model/cpptype.h
```diff
@@ -50,6 +50,8 @@ struct CppType : CppEntity
 {
   bool isAbstract = false;
   bool isPOD = false;
+  std::size_t size;
+  std::size_t alignment;
```
csordasmarton: Update the `toString` method to write out these values too.

plugins/cpp/service/src/cppservice.cpp
whisperity: I didn't want to send it as a string via the api because maybe someone wants to say a different suffix. I'll add this to the docs.

whisperity:
I have updated the C++ test: We can't really test for the value of `Size` but we are sure a value must be presented. We also can't really test for `Alignment` and it might not even exist. The *Properties service test* code has been enabled to handle these extra cases.

plugins/cpp/model/include/model/cpptype.h
```diff
@@ -51,6 +51,11 @@ struct CppType : CppEntity
   bool isAbstract = false;
   bool isPOD = false;
 
+  // Size and alignment is expressed in C++ standard bytes (sizeof(char)).
+  // Default value 0 means it couldn't have been obtained from Clang.
+  std::size_t size;
+  std::size_t alignment;
```
bruntib: In the comment you are writing about 0 as a default value, however these members are not initialized.

bruntib: What is the alignment of a record?

plugins/cpp/test/src/cpppropertiesservicetest.cpp
```diff
@@ -35,7 +37,9 @@ class CppPropertiesServiceTest : public ::testing::Test
   }
 
   void checkProperties(model::FileId fid_, int line_, int col_,
-    std::map<std::string, std::string>& expectedLines_)
+    std::map<std::string, std::string>& expectedValues_,
+    std::vector<std::string> ignoredValues_ = {},
+    std::vector<std::string> optionalProperties_ = {})
```
bruntib: Input variables should be passed by const reference.

plugins/cpp/test/src/cpppropertiesservicetest.cpp
```diff
@@ -88,7 +114,8 @@ TEST_F(CppPropertiesServiceTest, ClassPropertiesTest)
       {"Qualified name", "cc::test::SimpleClass"}
     };
 
-    checkProperties(_simpleClassHeader, 9, 10, expected);
+    checkProperties(_simpleClassHeader, 9, 10, expected,
+                    {"Size"}, {"Alignment"});
```
bruntib: We should consider testing the actual value of "size" property by using the `sizeof` operator.

plugins/cpp/model/include/model/cpptype.h
whisperity: An alignment of `x` bytes means that instances of the type should have their first byte at a memory address that is divisible by `x`. It is to ensure that member data accesses are optimal. See http://en.cppreference.com/w/c/language/object

bruntib: Is there any advantage of storing this information regarding code comprehension? Is there any reasonable use-case why somebody may be interested in the alignment? I'm not sure if any use-case would be worth for this storage.

plugins/cpp/test/src/cpppropertiesservicetest.cpp
bruntib: That's not true. This works for me:
```.cpp
void f(const std::vector<int>& v = {}) {}
int main() { f({}); }
```
Const reference keeps these temporaries alive.

plugins/cpp/model/include/model/cpptype.h
whisperity: (This should have went as reply to the other comment but the phone client is behaving strange.) It was @zporky's idea to store and present these two "memory model related" labels for the types. For alignment, his example was custom allocators.

whisperity:
I have updated the code with properly setting the initialiser for the new fields, and also updating the test to take `const`s.

plugins/cpp/model/include/model/cpptype.h
bruntib: I still don't think it would be worth extending database size with storing this info, not to mention that considering "reparse" as a future feature this will be reproducible. It is not really related to code comprehension.

whisperity:
Will look into how we can make this thing user-friendly with on-demand stuff with reparse, as opposed to dumping it into the database. Will need infrastructure stuff to allow for a `cpp rparse infotree` that can on-demand request single pieces of information.

bruntib:
My general concern about this question is the database size. The storage consumption and runtime efficiency has always been enemies and so far we always chose speed. I think we are fast enough for the base use cases, and we should not increase database size of not necessary. Actually it would be nice to see if we can reduce the current database size by throwing out potentially unnecessary and unused things. There is a related ticket #21 to store the file contents compressed in the database as CodeChecker does.

zporky:
A stable working reparse may reduce the database size and therefore
implicitly increase even the speed. Current increment parsing experiences
show that the database is the single most critical bottleneck. As an
(unlikely) extremity, I can imagine an experience where we have no parsed
database at all only some index, and we reparse on demand. The issue is
with go to def and all references: we still have to index which names occur
in which files.

gsd

PR #256, opened 2018-05-16
Update usage guide

### LINKED ISSUE #239: Usage guide relies on self-compiled Postgres, does not explain SQLite connection string
The `postgres` and `initdb` commands are only available if someone self-compiles postgres, `sudo su`s into the `postgresql` user, or does massive environmental hacking to make these commands available to them.

The guide should be changed to focus that on a standard Ubuntu install *with postgres installed from apt* how can one set up the database properly to the system-managed background database service.

whisperity:
- Details on how to use PostgreSQL if it was installed from package
   manager.
 - Added details on how to specify a SQLite database.
 - Extended the "Basic usage example" with giving the project directory
   so executing the example properly runs the search parser too.
 - Also added pointer and reference token style used in project to the coding conventions.

> Closes #239.

### DIFF
--- a/doc/coding_conventions.md
+++ b/doc/coding_conventions.md
@@ -26,8 +26,8 @@ Formatting
   namespace parser
   {
   /* ... */
-  }
-  }
+  } // parser
+  } // cc
   ```
 - **Blocks** Related opening and closing brackets should be placed to the same
   column (do not follow Java style). This rule holds for namespaces, classes,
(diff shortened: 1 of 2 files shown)


whisperity:
Need to also mention that if apt-get postgres is used and the database is to be created manually and not via a superuser, the following line has to be used:

    CREATE DATABASE mydatabase WITH ENCODING 'SQL_ASCII' LC_CTYPE 'C' LC_COLLATE 'C' TEMPLATE template0;

Because the default layout if UTF-8 which will result in an error at parsing.

PR #260, opened 2018-06-04
Adding Boost library to the linked directories

mcserep:
This merge request aims to solve a bug related to environments with multiple versions of the Boost library installed.

**Assume the following scenario:**
1.  Boost v1.58 is installed from the package repository to the default location, e.g. `/usr/lib/x86_64-linux-gnu/`;
2. Boost v1.60 is compiled and installed to another (standard) location, e.g. `/usr/local/lib/`.
3. Execute CMake for CodeCompass configured with Boost v1.60 (or actually it will find 1.60 by default in this case)
4. When compiling CodeCompass, it will try to link against the v1.58 libraries, resulting in a linking error. 

**Resolution:**
Specify `Boost_LIBRARY_DIRS` as a linked directory in CMake configurations where Boost is used.
I am not aware of a more elegant solution.


whisperity:
The documentation for [`link_directories`](https://cmake.org/cmake/help/v3.0/command/link_directories.html) says:

> Library locations returned by find_package() and find_library() are absolute paths. Pass these absolute library file paths directly to the target_link_libraries() command. CMake will ensure the linker finds them.

Could you please check that using this works or not:
 * Create a custom function or variable that exports the neccessary Boost libraries (`boost_thread`, `boost_filesystem`, `boost_system`, `boost_program_options`, `boost_log`) from the absolute paths returned by the `find_package` call in `CMakeLists.txt`, and use this in the `target_link_libraries` call?
 * [`FindBoost`'s docs](https://cmake.org/cmake/help/v3.0/module/FindBoost.html) shows an example line like this: `find_package(Boost 1.36.0 COMPONENTS date_time filesystem system ...)`. This makes me believe that the configuration line could be altered in the main CMakeLists file, and then use the `Boost_LIBRARIES` variable returned by the `find_package`... assuming the previously quoted piece is right.

Due to different plugins relying on different parts of Boost, perhaps a custom function that creates the absolutely pathed variable list is more expressive in this case.

mcserep:
It worked out, updated the merge request as you suggested.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ enable_testing()
 
 find_package(Threads REQUIRED)
 find_package(LLVM    REQUIRED CONFIG)
-find_package(Boost   REQUIRED)
+find_package(Boost   REQUIRED COMPONENTS thread filesystem system program_options log regex)
```
whisperity: In case the custom function works without manually specifying the `COMPONENTS [...]` here, I'd rather have these removed, as this just introduces an extra place where stuff needs to be maintained.

parser/CMakeLists.txt
```diff
@@ -14,15 +14,17 @@ add_executable(CodeCompass_parser
 set_target_properties(CodeCompass_parser
   PROPERTIES ENABLE_EXPORTS 1)
 
-target_link_libraries(CodeCompass_parser
-  util
-  model
+find_boost_library(
```
whisperity: I know originally they weren't in a good order, but in case changing the order does **not** actually break the install, we should order these things alphabetically.

parser/CMakeLists.txt
```diff
@@ -14,15 +14,17 @@ add_executable(CodeCompass_parser
...
   boost_thread
   boost_filesystem
   boost_system
   boost_program_options
-  boost_log
-  ${ODB_LIBRARIES}    
+  boost_log)
```
whisperity: Maybe the parameters here should be just `filesystem log program_options system thread`? (See my other comment in the newly introduced function.)

Functions.cmake
```diff
@@ -87,3 +87,17 @@ function(install_webplugin _dir)
   file(GLOB _userguides "${_dir}/userguide/*.md")
   set_property(GLOBAL APPEND PROPERTY USERGUIDES "${_userguides}")
 endfunction(install_webplugin)
+
+# Finds the absolute paths for the given Boost libraries
+# Use variable arguments for the Boost libraries to link
+function(find_boost_library)
+  foreach(_lib ${ARGV})
+    foreach(_path ${Boost_LIBRARIES})
+      if(_path MATCHES ".*${_lib}\.so$")
```
whisperity: Because every Boost library seemingly has a `boost_` prefix, this could be hardcoded here.

Functions.cmake
```diff
@@ -87,3 +87,17 @@ function(install_webplugin _dir)
   file(GLOB _userguides "${_dir}/userguide/*.md")
   set_property(GLOBAL APPEND PROPERTY USERGUIDES "${_userguides}")
 endfunction(install_webplugin)
+
+# Finds the absolute paths for the given Boost libraries
+# Use variable arguments for the Boost libraries to link
+function(find_boost_library)
```
whisperity: To keep ourselves to CMake's DSL's naming convention, the function should have a plural name: `find_boost_libraries`.

whisperity:
@mcserep That is good to hear! Are you using 16.04 or 18.04 to develop CodeCompass as of now?

I've ran into issues with the linkage too on a build server earlier because it got auto-updated, but I could solve it by simply `purge`ing the previous things and doing a full rebuild.

Functions.cmake
mcserep: Updated accordingly.

Functions.cmake
mcserep: Updated accordingly.

parser/CMakeLists.txt
mcserep: Updated accordingly.

parser/CMakeLists.txt
mcserep: It's OK to order them alphabetically, updated accordingly.

mcserep:
@whisperity I use Ubuntu 16.04.

util/CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ add_library(util STATIC
   src/pipedprocess.cpp)
 
 target_compile_options(util PUBLIC -fPIC)
+
+find_boost_library(
+  boost_regex)
 target_link_libraries(util
-  boost_regex
+  ${Boost_LINK_LIBRARIES}
   sqlite3)
```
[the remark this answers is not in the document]
mcserep: Looks suspicious. Might be the reason why CodeCompass requires SQLite on build even when configured to use PostgreSQL. Added the suggested comment.

CMakeLists.txt
whisperity: Uh-huh. Okay, I understand it now. Weird either way, I believe it was more in line with common sense that if no `COMPONENTS` is given, then it means "every component" and thus I could fetch what is installed on the system...

whisperity:
I have a lot of ideas on how to make the entire build ecosystem a bit more expressive and user-friendly, thank you for running into this issue and fixing it. :wink:

(P.s.: You submitted the pull request from your `master` branch! Just something to watch out for later...)

PR #266, opened 2018-07-15
Incremental parsing

mcserep:
Initial support for incremental parsing. Incremental parsing mode is carried out automatically in case both the specified database and the project workspace directory exist.
The main objectives were the following.

### Refactor the parser architecture
The `parse()` method of `AbstractParser` was separated into 2 phases: `preparse()` and `parse()`.
First, the `preparse()`  method for all parsers are evaluated (in reverse topological order), which are responsible for the database and project workspace maintenance for incremental parsing. Then, parsing on a cleaned up environment is carried out by the `parse()` methods for each parser in a same way as a "normal", initial parse.

Deleted and modified files are detected by the `ParserContext` and this is information is made available towards each parser.

### Implement incremental support for the parsers
- **CppParser**: all information regarding the deleted and modified files are cleaned up from the database, these files are treated as new files and parsed accordingly. Files along the header inclusion chains are also marked modified if a header file was edited.
- **MetricsParser:** all information regarding the deleted and modified files are cleaned up from the database, these files are treated as new files and parsed accordingly.
- **SearchParser:** the `search` directory in the project workspace is deleted and a new, complete parsing is carried out.
- **GitParser:** the repository directory under the `version` directory in the project workspace is deleted and a new, complete parsing is carried out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # IDEs and editors
 
-nbproject
+nbproject/
+.idea/
 *.swp
 
 # Generated files and directories
@@ -17,3 +18,4 @@ nbproject
 # Other
 
 /build/
+/build_*/
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,4 @@ nbproject/
 
 # Other
 
-/build/
-/build_*/
+/build*/
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
   -j 4
 ```
 
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a **complete** compilation database, 
+therefore the build commands for only the modified files are not sufficient.
+In case of CMake, using the result of the `CMAKE_EXPORT_COMPILE_COMMANDS=ON` argument, the 
+compilation database will always contain all files.
+Currently the C++ and metrics parsers support incremental parsing, while other parsers
+just execute a forced reparse.
+
 ## 3. Start the web server
 You can start the CodeCompass webserver with `CodeCompass_webserver` binary in
 the CodeCompass installation directory.
--- a/model/include/model/buildsourcetarget.h
+++ b/model/include/model/buildsourcetarget.h
@@ -25,6 +25,7 @@ struct BuildSource
   FilePtr file;
 
   #pragma db not_null
+  #pragma db on_delete(cascade)
   std::shared_ptr<BuildAction> action;
 };
 
@@ -40,6 +41,7 @@ struct BuildTarget
   FilePtr file;
 
   #pragma db not_null
+#pragma db on_delete(cascade)
   std::shared_ptr<BuildAction> action;
 };
 
--- a/model/include/model/buildsourcetarget.h
+++ b/model/include/model/buildsourcetarget.h
@@ -41,7 +41,7 @@ struct BuildTarget
   FilePtr file;
 
   #pragma db not_null
-#pragma db on_delete(cascade)
+  #pragma db on_delete(cascade)
   std::shared_ptr<BuildAction> action;
 };
 
--- a/parser/include/parser/abstractparser.h
+++ b/parser/include/parser/abstractparser.h
@@ -36,6 +36,7 @@ public:
   /**
    * Maintains and cleans up the database in preparation of
    * incremental parsing.
+   * @return Return true if the preparse was success, false otherwise.
    */
   virtual bool preparse()
   {
--- a/parser/include/parser/abstractparser.h
+++ b/parser/include/parser/abstractparser.h
@@ -33,7 +33,14 @@ public:
    */
   virtual std::vector<std::string> getDependentParsers() const = 0;
 
-  virtual void preparse(){}
+  /**
+   * Maintains and cleans up the database in preparation of
+   * incremental parsing.
+   */
+  virtual bool preparse()
+  {
+    return true;
+  }
 
   /**
    * Method parses a path or a compilation database
--- a/parser/include/parser/abstractparser.h
+++ b/parser/include/parser/abstractparser.h
@@ -33,6 +33,8 @@ public:
    */
   virtual std::vector<std::string> getDependentParsers() const = 0;
 
+  virtual void preparse(){}
+
   /**
    * Method parses a path or a compilation database
    * @return Return true if the parse was success, false otherwise
--- a/parser/include/parser/parsercontext.h
+++ b/parser/include/parser/parsercontext.h
@@ -36,8 +36,7 @@ struct ParserContext
       std::shared_ptr<odb::database> db_,
       SourceManager& srcMgr_,
       std::string& compassRoot_,
-      po::variables_map& options_,
-      std::unordered_map<std::string, IncrementalStatus> fileStatus_);
+      po::variables_map& options);
 
   std::shared_ptr<odb::database> db;
   SourceManager& srcMgr;
--- a/parser/include/parser/parsercontext.h
+++ b/parser/include/parser/parsercontext.h
@@ -1,20 +1,13 @@
 #ifndef CC_PARSER_PARSERCONTEXT_H
 #define CC_PARSER_PARSERCONTEXT_H
 
+#include <fstream>
 #include <memory>
 #include <unordered_map>
-#include <fstream>
 
 #include <boost/program_options.hpp>
-#include <boost/filesystem.hpp>
 
 #include <odb/database.hxx>
-
-#include <model/file.h>
-#include <model/file-odb.hxx>
-
-#include <util/hash.h>
-#include <util/odbtransaction.h>
 
 namespace po = boost::program_options; 
 
@@ -25,8 +18,10 @@ namespace parser
 
 class SourceManager;
 
- /**
+/**
  * Defines file status categories for incremental parsing.
+ *
+ * State in database VERSUS state on disk at parse time.
  */
 enum class IncrementalStatus
 {
--- a/parser/include/parser/parsercontext.h
+++ b/parser/include/parser/parsercontext.h
@@ -2,8 +2,19 @@
 #define CC_PARSER_PARSERCONTEXT_H
 
 #include <memory>
+#include <unordered_map>
+#include <fstream>
+
 #include <boost/program_options.hpp>
+#include <boost/filesystem.hpp>
+
 #include <odb/database.hxx>
+
+#include <model/file.h>
+#include <model/file-odb.hxx>
+
+#include <util/hash.h>
+#include <util/odbtransaction.h>
 
 namespace po = boost::program_options; 
 
@@ -14,24 +25,30 @@ namespace parser
 
 class SourceManager;
 
+ /**
+ * Defines file status categories for incremental parsing.
+ */
+enum class IncrementalStatus
+{
+  MODIFIED,
+  ADDED,
+  DELETED
+};
+
 struct ParserContext 
 {  
   ParserContext(
-    std::shared_ptr<odb::database> db_,
-    SourceManager& srcMgr_,
-    std::string& compassRoot_,
-    po::variables_map& options_) :
-      db(db_),
-      srcMgr(srcMgr_),
-      compassRoot(compassRoot_),
-      options(options_)
-  {
-  }
+      std::shared_ptr<odb::database> db_,
+      SourceManager& srcMgr_,
+      std::string& compassRoot_,
+      po::variables_map& options_,
+      std::unordered_map<std::string, IncrementalStatus> fileStatus_);
 
   std::shared_ptr<odb::database> db;
   SourceManager& srcMgr;
   std::string& compassRoot;
   po::variables_map& options;
+  std::unordered_map<std::string, IncrementalStatus> fileStatus;
 };
 
 } // parser
--- a/parser/include/parser/parsercontext.h
+++ b/parser/include/parser/parsercontext.h
@@ -1,7 +1,6 @@
 #ifndef CC_PARSER_PARSERCONTEXT_H
 #define CC_PARSER_PARSERCONTEXT_H
 
-#include <fstream>
 #include <memory>
 #include <unordered_map>
 
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -216,7 +216,7 @@ int main(int argc, char* argv[])
   //--- Check database and project directory existence ---//
   
   bool isNewDb = cc::util::connectDatabase(
-    vm["database"].as<std::string>(), false) ? false : true;
+    vm["database"].as<std::string>(), false) == nullptr;
   bool isNewProject = !checkProjectDir(vm);
 
   if (isNewProject ^ isNewDb)
@@ -258,7 +258,7 @@ int main(int argc, char* argv[])
   //--- Start parsers ---//
 
   cc::parser::SourceManager srcMgr(db);
-  cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm, fileStatus);
+  cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm);
   pHandler.createPlugins(ctx);
 
   // TODO: Handle errors returned by preparse().
@@ -266,7 +266,7 @@ int main(int argc, char* argv[])
   for (auto it = topologicalOrder.rbegin(); it != topologicalOrder.rend(); ++it)
   {
     LOG(info) << "[" << *it << "] preparse started!";
-    if(!pHandler.getParser(*it)->preparse())
+    if (!pHandler.getParser(*it)->preparse())
     {
       LOG(error) << "[" << *it << "] preparse failed!";
       return 2;
@@ -274,7 +274,7 @@ int main(int argc, char* argv[])
   }
 
   // TODO: Consider whether there is a better place for this code.
-  (cc::util::OdbTransaction(ctx.db))([&]
+  cc::util::OdbTransaction {ctx.db} ([&]
   {
     for (auto& item : ctx.fileStatus)
     {
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -219,14 +219,14 @@ int main(int argc, char* argv[])
     vm["database"].as<std::string>(), false) ? false : true;
   bool isNewProject = !checkProjectDir(vm);
 
-  if (isNewProject && !isNewDb || !isNewProject && isNewDb)
+  if (isNewProject ^ isNewDb)
   {
     LOG(error)
       << "Database and working directory existence are inconsistent. Use -f for reparsing!";
     return 1;
   }
 
-  if(!isNewDb)
+  if (!isNewDb)
     LOG(info) << "Project already exists, incremental parsing in action.";
 
   //--- Prepare workspace and project directory ---//
@@ -261,20 +261,22 @@ int main(int argc, char* argv[])
   cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm, fileStatus);
   pHandler.createPlugins(ctx);
 
-  std::vector<std::string> reversedTopologicalOrder = pHandler.getTopologicalOrder();
-  std::reverse(reversedTopologicalOrder.begin(), reversedTopologicalOrder.end());
-
   // TODO: Handle errors returned by preparse().
-  for(const std::string& parserName : reversedTopologicalOrder)
+  std::vector<std::string> topologicalOrder = pHandler.getTopologicalOrder();
+  for (auto it = topologicalOrder.rbegin(); it != topologicalOrder.rend(); ++it)
   {
-    LOG(info) << "[" << parserName << "] preparse started!";
-    pHandler.getParser(parserName)->preparse();
+    LOG(info) << "[" << *it << "] preparse started!";
+    if(!pHandler.getParser(*it)->preparse())
+    {
+      LOG(error) << "[" << *it << "] preparse failed!";
+      return 2;
+    }
   }
 
   // TODO: Consider whether there is a better place for this code.
   (cc::util::OdbTransaction(ctx.db))([&]
   {
-    for(auto& item : ctx.fileStatus)
+    for (auto& item : ctx.fileStatus)
     {
       switch (item.second)
       {
@@ -306,7 +308,7 @@ int main(int argc, char* argv[])
 
   //--- Add indexes to the database ---//
 
-  if(vm.count("force") || isNewDb)
+  if (vm.count("force") || isNewDb)
     cc::util::createIndexes(db, SQL_DIR);
 
   //--- Create project config file ---//
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -211,8 +211,8 @@ int main(int argc, char* argv[])
     LOG(error) << "Error in command line arguments: " << e.what();
     return 1;
   }
-
-  //--- Check database and project directory existsence ---//
+  
+  //--- Check database and project directory existence ---//
   
   bool isNewDb = cc::util::connectDatabase(
     vm["database"].as<std::string>(), false) ? false : true;
@@ -224,6 +224,9 @@ int main(int argc, char* argv[])
       << "Database and working directory existence are inconsistent. Use -f for reparsing!";
     return 1;
   }
+
+  if(!isNewDb)
+    LOG(info) << "Project already exists, incremental parsing in action.";
 
   //--- Prepare workspace and project directory ---//
   
@@ -260,15 +263,17 @@ int main(int argc, char* argv[])
   std::vector<std::string> reversedTopologicalOrder = pHandler.getTopologicalOrder();
   std::reverse(reversedTopologicalOrder.begin(), reversedTopologicalOrder.end());
 
+  // TODO: Handle errors returned by preparse().
   for(const std::string& parserName : reversedTopologicalOrder)
   {
+    LOG(info) << "[" << parserName << "] preparse started!";
     pHandler.getParser(parserName)->preparse();
   }
 
   // TODO: Handle errors returned by parse().
   for (const std::string& parserName : pHandler.getTopologicalOrder())
   {
-    LOG(info) << "[" << parserName << "] started!";
+    LOG(info) << "[" << parserName << "] parse started!";
     pHandler.getParser(parserName)->parse();
   }
 
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -16,6 +16,7 @@
 #include <util/dbutil.h>
 #include <util/filesystem.h>
 #include <util/logutil.h>
+#include <util/odbtransaction.h>
 
 #include <parser/parsercontext.h>
 #include <parser/pluginhandler.h>
@@ -269,6 +270,32 @@ int main(int argc, char* argv[])
     LOG(info) << "[" << parserName << "] preparse started!";
     pHandler.getParser(parserName)->preparse();
   }
+
+  // TODO: Consider whether there is a better place for this code.
+  (cc::util::OdbTransaction(ctx.db))([&]
+  {
+    for(auto& item : ctx.fileStatus)
+    {
+      switch (item.second)
+      {
+        case cc::parser::IncrementalStatus::MODIFIED:
+        case cc::parser::IncrementalStatus::DELETED:
+        {
+          LOG(info) << "Database cleanup: " << item.first;
+
+          // Fetch file from SourceManager by path
+          cc::model::FilePtr delFile = srcMgr.getFile(item.first);
+
+          // Delete File and FileContent (only when no other File references it)
+          srcMgr.removeFile(*delFile);
+          break;
+        }
+        case cc::parser::IncrementalStatus::ADDED:
+          // Empty deliberately
+          break;
+      }
+    }
+  });
 
   // TODO: Handle errors returned by parse().
   for (const std::string& parserName : pHandler.getTopologicalOrder())
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -82,7 +82,7 @@ po::options_description commandLineArguments()
 /**
  * This function checks the existence of the workspace and project directory
  * based on the given command line arguments.
- * @return The path of the project directory.
+ * @return Whether the project directory exists or not.
  */
 bool checkProjectDir(const po::variables_map& vm_)
 {
@@ -90,7 +90,7 @@ bool checkProjectDir(const po::variables_map& vm_)
     = vm_["workspace"].as<std::string>() + '/'
       + vm_["name"].as<std::string>();
 
-  return fs::exists(projDir) && fs::is_directory(projDir);
+  return fs::is_directory(projDir);
 }
 
 /**
@@ -276,7 +276,7 @@ int main(int argc, char* argv[])
   // TODO: Consider whether there is a better place for this code.
   cc::util::OdbTransaction {ctx.db} ([&]
   {
-    for (auto& item : ctx.fileStatus)
+    for (const auto& item : ctx.fileStatus)
     {
       switch (item.second)
       {
@@ -300,7 +300,7 @@ int main(int argc, char* argv[])
   });
 
   // TODO: Handle errors returned by parse().
-  for (const std::string& parserName : pHandler.getTopologicalOrder())
+  for (const std::string& parserName : topologicalOrder)
   {
     LOG(info) << "[" << parserName << "] parse started!";
     pHandler.getParser(parserName)->parse();
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -79,10 +79,23 @@ po::options_description commandLineArguments()
 }
 
 /**
+ * This function checks the existence of the workspace and project directory
+ * based on the given command line arguments.
+ * @return The path of the project directory.
+ */
+bool checkProjectDir(const po::variables_map& vm_)
+{
+  const std::string projDir
+    = vm_["workspace"].as<std::string>() + '/'
+      + vm_["name"].as<std::string>();
+
+  return fs::exists(projDir) && fs::is_directory(projDir);
+}
+
+/**
  * This function prepares the workspace and project directory based on the given
  * command line arguments. If the project directory can't be created because of
- * permission issues or the directory already exists and "--force" flag is not
- * provided then empty string returns which indicates the problem.
+ * permission issues then empty string returns which indicates the problem.
  * @return The path of the project directory.
  */
 std::string prepareProjectDir(const po::variables_map& vm_)
@@ -121,11 +134,6 @@ std::string prepareProjectDir(const po::variables_map& vm_)
       LOG(error) << "Permission denied to create " + projDir;
       return std::string();
     }
-  }
-  else
-  {
-    LOG(error) << projDir << " already exists. Use -f for reparsing!";
-    return std::string();
   }
 
   return projDir;
@@ -204,7 +212,20 @@ int main(int argc, char* argv[])
     return 1;
   }
 
-  //--- Check workspace and project directory ---//
+  //--- Check database and project directory existsence ---//
+
+  bool isNewDb = cc::util::connectDatabase(
+    vm["database"].as<std::string>(), false) ? false : true;
+  bool isNewProject = !checkProjectDir(vm);
+
+  if (isNewProject && !isNewDb || !isNewProject && isNewDb)
+  {
+    LOG(error)
+      << "Database and working directory existence are inconsistent. Use -f for reparsing!";
+    return 1;
+  }
+
+  //--- Prepare workspace and project directory ---//
 
   std::string projDir = prepareProjectDir(vm);
   if (projDir.empty())
@@ -212,8 +233,8 @@ int main(int argc, char* argv[])
 
   //--- Create and init database ---//
 
-  std::shared_ptr<odb::database> db = cc::util::createDatabase(
-    vm["database"].as<std::string>());
+  std::shared_ptr<odb::database> db = cc::util::connectDatabase(
+      vm["database"].as<std::string>(), true);
 
   if (!db)
   {
@@ -225,7 +246,8 @@ int main(int argc, char* argv[])
   if (vm.count("force"))
     cc::util::removeTables(db, SQL_DIR);
 
-  cc::util::createTables(db, SQL_DIR);
+  if (vm.count("force") || isNewDb)
+    cc::util::createTables(db, SQL_DIR);
 
   //--- Start parsers ---//
 
@@ -242,7 +264,8 @@ int main(int argc, char* argv[])
 
   //--- Add indexes to the database ---//
 
-  cc::util::createIndexes(db, SQL_DIR);
+  if(vm.count("force") || isNewDb)
+    cc::util::createIndexes(db, SQL_DIR);
 
   //--- Create project config file ---//
 
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -213,7 +213,7 @@ int main(int argc, char* argv[])
   }
 
   //--- Check database and project directory existsence ---//
-
+  
   bool isNewDb = cc::util::connectDatabase(
     vm["database"].as<std::string>(), false) ? false : true;
   bool isNewProject = !checkProjectDir(vm);
@@ -226,7 +226,7 @@ int main(int argc, char* argv[])
   }
 
   //--- Prepare workspace and project directory ---//
-
+  
   std::string projDir = prepareProjectDir(vm);
   if (projDir.empty())
     return 1;
@@ -235,6 +235,8 @@ int main(int argc, char* argv[])
 
   std::shared_ptr<odb::database> db = cc::util::connectDatabase(
       vm["database"].as<std::string>(), true);
+
+  std::unordered_map<std::string, cc::parser::IncrementalStatus> fileStatus;
 
   if (!db)
   {
@@ -252,8 +254,16 @@ int main(int argc, char* argv[])
   //--- Start parsers ---//
 
   cc::parser::SourceManager srcMgr(db);
-  cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm);
+  cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm, fileStatus);
   pHandler.createPlugins(ctx);
+
+  std::vector<std::string> reversedTopologicalOrder = pHandler.getTopologicalOrder();
+  std::reverse(reversedTopologicalOrder.begin(), reversedTopologicalOrder.end());
+
+  for(const std::string& parserName : reversedTopologicalOrder)
+  {
+    pHandler.getParser(parserName)->preparse();
+  }
 
   // TODO: Handle errors returned by parse().
   for (const std::string& parserName : pHandler.getTopologicalOrder())
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -1,3 +1,5 @@
+#include <fstream>
+
 #include <boost/filesystem.hpp>
 
 #include <model/file.h>
@@ -30,14 +32,13 @@ ParserContext::ParserContext(
 
   (util::OdbTransaction(this->db))([&]
    {
-     // Fetch all files from SourceManager
-     std::vector<model::FilePtr> files = this->srcMgr.getAllFiles();
-     files.erase(std::remove_if(files.begin(), files.end(),
-                                [](const auto &item)
-                                {
-                                  return item->type == model::File::DIRECTORY_TYPE ||
-                                         item->type == model::File::BINARY_TYPE;
-                                }), files.end());
+     // Fetch directory and binary type files from SourceManager
+     auto func = [](const auto& item)
+     {
+       return item->type != model::File::DIRECTORY_TYPE &&
+              item->type != model::File::BINARY_TYPE;
+     };
+     std::vector<model::FilePtr> files = this->srcMgr.getFiles(func);
 
      for (model::FilePtr file : files)
      {
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -12,7 +12,7 @@
 namespace po = boost::program_options;
 
 namespace cc
-{,
+{
 namespace parser
 {
 
@@ -43,12 +43,13 @@ ParserContext::ParserContext(
      {
        if (boost::filesystem::exists(file->path))
        {
-         if (!this->fileStatus.count(file->path))
+         if (!fileStatus.count(file->path))
          {
-           auto content = file->content.load();
-           fileHashes[file->path] = content != nullptr ? content->hash : "";
-           if (content == nullptr)
+           model::FileContentPtr content = file->content.load();
+           if (!content)
              continue;
+
+           fileHashes[file->path] = content->hash;
 
            std::ifstream fileStream(file->path);
            std::string fileContent(
@@ -58,16 +59,16 @@ ParserContext::ParserContext(
 
            if (content->hash != util::sha1Hash(fileContent))
            {
-             this->fileStatus.insert(
-               std::make_pair(file->path, cc::parser::IncrementalStatus::MODIFIED));
+             this->fileStatus.emplace(
+               file->path, cc::parser::IncrementalStatus::MODIFIED);
              LOG(debug) << "File modified: " << file->path;
            }
          }
        }
        else
        {
-         this->fileStatus.insert(
-             std::make_pair(file->path, cc::parser::IncrementalStatus::DELETED));
+         fileStatus.emplace(
+             file->path, cc::parser::IncrementalStatus::DELETED);
          LOG(debug) << "File deleted: " << file->path;
        }
      }
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -1,0 +1,73 @@
+#include <parser/parsercontext.h>
+#include <parser/sourcemanager.h>
+
+namespace po = boost::program_options;
+
+namespace cc
+{,
+namespace parser
+{
+
+ParserContext::ParserContext(
+    std::shared_ptr<odb::database> db_,
+    SourceManager& srcMgr_,
+    std::string& compassRoot_,
+    po::variables_map& options_,
+    std::unordered_map<std::string, IncrementalStatus> fileStatus_):
+    db(db_),
+    srcMgr(srcMgr_),
+    compassRoot(compassRoot_),
+    options(options_),
+    fileStatus(fileStatus_)
+{
+  std::unordered_map<std::string, std::string> fileHashes;
+
+  (util::OdbTransaction(this->db))([&]
+   {
+     // Fetch all files from SourceManager
+     std::vector<model::FilePtr> files = this->srcMgr.getAllFiles();
+     files.erase(std::remove_if(files.begin(), files.end(),
+                                [](const auto &item)
+                                {
+                                  return item->type == model::File::DIRECTORY_TYPE ||
+                                         item->type == model::File::BINARY_TYPE;
+                                }), files.end());
+
+     for (model::FilePtr file : files)
+     {
+       if (boost::filesystem::exists(file->path))
+       {
+         if (!this->fileStatus.count(file->path))
+         {
+           auto content = file->content.load();
+           fileHashes[file->path] = content != nullptr ? content->hash : "";
+           if (content == nullptr)
+             continue;
+
+           std::ifstream fileStream(file->path);
+           std::string fileContent(
+               (std::istreambuf_iterator<char>(fileStream)),
+               (std::istreambuf_iterator<char>()));
+           fileStream.close();
+
+           if (content->hash != util::sha1Hash(fileContent))
+           {
+             this->fileStatus.insert(
+               std::make_pair(file->path, cc::parser::IncrementalStatus::MODIFIED));
+             LOG(debug) << "File modified: " << file->path;
+           }
+         }
+       }
+       else
+       {
+         this->fileStatus.insert(
+             std::make_pair(file->path, cc::parser::IncrementalStatus::DELETED));
+         LOG(debug) << "File deleted: " << file->path;
+       }
+     }
+   });
+}
+
+}
+}
+
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -65,6 +65,8 @@ ParserContext::ParserContext(
          LOG(debug) << "File deleted: " << file->path;
        }
      }
+
+     // TODO: detect ADDED files
    });
 }
 
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -1,3 +1,11 @@
+#include <boost/filesystem.hpp>
+
+#include <model/file.h>
+#include <model/file-odb.hxx>
+
+#include <util/hash.h>
+#include <util/odbtransaction.h>
+
 #include <parser/parsercontext.h>
 #include <parser/sourcemanager.h>
 
@@ -12,13 +20,11 @@ ParserContext::ParserContext(
     std::shared_ptr<odb::database> db_,
     SourceManager& srcMgr_,
     std::string& compassRoot_,
-    po::variables_map& options_,
-    std::unordered_map<std::string, IncrementalStatus> fileStatus_):
-    db(db_),
-    srcMgr(srcMgr_),
-    compassRoot(compassRoot_),
-    options(options_),
-    fileStatus(fileStatus_)
+    po::variables_map& options_):
+      db(db_),
+      srcMgr(srcMgr_),
+      options(options_),
+      compassRoot(compassRoot_)
 {
   std::unordered_map<std::string, std::string> fileHashes;
 
--- a/parser/src/sourcemanager.cpp
+++ b/parser/src/sourcemanager.cpp
@@ -221,6 +221,31 @@ void SourceManager::updateFile(const model::File& file_)
     });
 }
 
+void SourceManager::removeFile(const model::File& file_)
+{
+  bool removeContent = false;
+
+  // Delete File and FileContent (only when no other File references it)
+  _transaction([&, this]() {
+    auto relFiles = _db->query<model::File>(
+        odb::query<model::File>::content == file_.content.object_id());
+    if(std::distance(relFiles.begin(), relFiles.end()) == 1)
+    {
+      removeContent = true;
+      _db->erase<model::FileContent>(file_.content.object_id());
+    }
+    _db->erase<model::File>(file_.id);
+  });
+
+  // Maintain cache
+  _createFileMutex.lock();
+  _files.erase(file_.path);
+  _persistedFiles.erase(file_.id);
+  if(removeContent)
+    _persistedContents.erase(file_.content.object_id());
+  _createFileMutex.unlock();
+}
+
 void SourceManager::persistFiles()
 {
   std::lock_guard<std::mutex> guard(_createFileMutex);
--- a/parser/src/sourcemanager.cpp
+++ b/parser/src/sourcemanager.cpp
@@ -229,7 +229,7 @@ void SourceManager::updateFile(const model::File& file_)
   _createFileMutex.unlock();
 
   if (find)
-    _transaction([&, this]() {
+    _transaction([&]() {
       _db->update(file_);
     });
 }
@@ -239,10 +239,10 @@ void SourceManager::removeFile(const model::File& file_)
   bool removeContent = false;
 
   // Delete File and FileContent (only when no other File references it)
-  _transaction([&, this]() {
+  _transaction([&]() {
     auto relFiles = _db->query<model::File>(
         odb::query<model::File>::content == file_.content.object_id());
-    if(std::distance(relFiles.begin(), relFiles.end()) == 1)
+    if (relFiles.size() == 1)
     {
       removeContent = true;
       _db->erase<model::FileContent>(file_.content.object_id());
@@ -251,18 +251,20 @@ void SourceManager::removeFile(const model::File& file_)
   });
 
   // Maintain cache
-  std::lock_guard<std::mutex> guard(_createFileMutex);
-  _files.erase(file_.path);
-  _persistedFiles.erase(file_.id);
-  if (removeContent)
-    _persistedContents.erase(file_.content.object_id());
+  {
+    std::lock_guard<std::mutex> guard(_createFileMutex);
+    _files.erase(file_.path);
+    _persistedFiles.erase(file_.id);
+    if (removeContent)
+      _persistedContents.erase(file_.content.object_id());
+  }
 }
 
 void SourceManager::persistFiles()
 {
   std::lock_guard<std::mutex> guard(_createFileMutex);
 
-  _transaction([&, this]() {
+  _transaction([&]() {
     for (const auto& p : _files)
     {
       if (_persistedFiles.find(p.second->id) == _persistedFiles.end())
--- a/parser/src/sourcemanager.cpp
+++ b/parser/src/sourcemanager.cpp
@@ -251,12 +251,11 @@ void SourceManager::removeFile(const model::File& file_)
   });
 
   // Maintain cache
-  _createFileMutex.lock();
+  std::lock_guard<std::mutex> guard(_createFileMutex);
   _files.erase(file_.path);
   _persistedFiles.erase(file_.id);
-  if(removeContent)
+  if (removeContent)
     _persistedContents.erase(file_.content.object_id());
-  _createFileMutex.unlock();
 }
 
 void SourceManager::persistFiles()
--- a/parser/src/sourcemanager.cpp
+++ b/parser/src/sourcemanager.cpp
@@ -149,6 +149,19 @@ model::FilePtr SourceManager::getCreateFileEntry(
   return file;
 }
 
+std::vector<model::FilePtr> SourceManager::getAllFiles()
+{
+  std::lock_guard<std::mutex> guard(_createFileMutex);
+
+  std::vector<model::FilePtr> values;
+  values.reserve(_files.size());
+
+  std::transform(_files.begin(), _files.end(), std::back_inserter(values),
+                 [](const auto &pair) { return pair.second; });
+
+  return values;
+}
+
 model::FilePtr SourceManager::getFile(const std::string& path_)
 {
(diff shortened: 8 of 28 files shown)


mcserep:
Resolved the conflicts with `Ericsson:master` and re-tested functionality.

parser/include/parser/parsercontext.h
```diff
@@ -14,24 +25,29 @@ namespace parser
 
 class SourceManager;
 
+ /**
+ * Defines file status categories for incremental parsing.
+ */
+enum class IncrementalStatus
+{
+  MODIFIED,
```
whisperity: What do these values mean?

parser/include/parser/parsercontext.h
```diff
@@ -2,9 +2,20 @@
 #define CC_PARSER_PARSERCONTEXT_H
 
 #include <memory>
+#include <unordered_map>
```
whisperity: The includes should be ordered here.

parser/src/parser.cpp
```diff
@@ -204,16 +213,32 @@ int main(int argc, char* argv[])
...
+  bool isNewProject = !checkProjectDir(vm);
+
+  if (isNewProject && !isNewDb || !isNewProject && isNewDb)
+  {
+    LOG(error)
+      << "Database and working directory existence are inconsistent. Use -f for reparsing!";
+    return 1;
+  }
+
+  if(!isNewDb)
```
whisperity: Space missing between if and condition.

parser/src/parser.cpp
```diff
@@ -204,16 +213,32 @@ int main(int argc, char* argv[])
...
   }
 
-  //--- Check workspace and project directory ---//
+  //--- Check database and project directory existence ---//
+
+  bool isNewDb = cc::util::connectDatabase(
+    vm["database"].as<std::string>(), false) ? false : true;
+  bool isNewProject = !checkProjectDir(vm);
+
+  if (isNewProject && !isNewDb || !isNewProject && isNewDb)
```
whisperity: `&&` within `||` will trigger compiler warnings, please put the subformulae of `||` in parentheses.

whisperity: Also, I think this condition is the same as `isNewProject XOR isNewDb`.

parser/src/parser.cpp
```diff
@@ -204,16 +213,32 @@ int main(int argc, char* argv[])
     return 1;
   }
 
-  //--- Check workspace and project directory ---//
+  //--- Check database and project directory existence ---//
+
+  bool isNewDb = cc::util::connectDatabase(
```
whisperity: `!cc::util::connectDatabase(...)`?

parser/src/parser.cpp
```diff
@@ -225,24 +250,62 @@ int main(int argc, char* argv[])
...
+  if (vm.count("force") || isNewDb)
+    cc::util::createTables(db, SQL_DIR);
 
   //--- Start parsers ---//
 
   cc::parser::SourceManager srcMgr(db);
   cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm);
   pHandler.createPlugins(ctx);
 
+  std::vector<std::string> reversedTopologicalOrder = pHandler.getTopologicalOrder();
```
whisperity: The original variable does not contain anything reversed. This can be confusing as there is a direct dependency for these two lines to always exist together.

parser/src/parser.cpp
```diff
@@ -225,24 +250,62 @@ int main(int argc, char* argv[])
...
+  std::reverse(reversedTopologicalOrder.begin(), reversedTopologicalOrder.end());
+
+  // TODO: Handle errors returned by preparse().
+  for(const std::string& parserName : reversedTopologicalOrder)
```
whisperity: Space missing

parser/src/parser.cpp
```diff
@@ -225,24 +250,62 @@ int main(int argc, char* argv[])
...
+  {
+    LOG(info) << "[" << parserName << "] preparse started!";
+    pHandler.getParser(parserName)->preparse();
+  }
+
+  // TODO: Consider whether there is a better place for this code.
+  (cc::util::OdbTransaction(ctx.db))([&]
+  {
+    for(auto& item : ctx.fileStatus)
```
whisperity: Space missing

parser/src/parser.cpp
```diff
@@ -225,24 +250,62 @@ int main(int argc, char* argv[])
...
   {
-    LOG(info) << "[" << parserName << "] started!";
+    LOG(info) << "[" << parserName << "] parse started!";
     pHandler.getParser(parserName)->parse();
   }
 
   //--- Add indexes to the database ---//
 
-  cc::util::createIndexes(db, SQL_DIR);
+  if(vm.count("force") || isNewDb)
```
whisperity: Space missing.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+namespace parser
+{
+
+ParserContext::ParserContext(
+    std::shared_ptr<odb::database> db_,
+    SourceManager& srcMgr_,
+    po::variables_map& options_,
+    std::unordered_map<std::string, IncrementalStatus> fileStatus_):
+    db(db_),
```
whisperity: Indent the initialisers two more spaces.

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,31 @@ void SourceManager::updateFile(const model::File& file_)
...
+    if(std::distance(relFiles.begin(), relFiles.end()) == 1)
+    {
+      removeContent = true;
+      _db->erase<model::FileContent>(file_.content.object_id());
+    }
+    _db->erase<model::File>(file_.id);
+  });
+
+  // Maintain cache
+  _createFileMutex.lock();
```
whisperity: Shouldn't it be better to use the lock guard (scoped guard) initialized here, in case an exception or something happens, rather than to manual lock?

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,31 @@ void SourceManager::updateFile(const model::File& file_)
...
+  _files.erase(file_.path);
+  _persistedFiles.erase(file_.id);
+  if(removeContent)
```
whisperity: Space missing.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -1,6 +1,9 @@
 #include <algorithm>
 #include <memory>
 #include <unordered_map>
+#include <fstream>
```
whisperity: Ordering wrong here.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,137 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+  {
+    for(const model::FilePtr file : filePtrs)
+    {
+      if(file)
+      {
+        markAsModified(file);
+      }
+    }
+
+    for(auto& item : _ctx.fileStatus)
```
whisperity: Spaces missing around these parts, again.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,137 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+    {
+      switch (item.second)
+      {
+        case IncrementalStatus::MODIFIED:
+        case IncrementalStatus::DELETED:
+        {
+          LOG(info) << "[cppparser] Database cleanup: " << item.first;
+
+          // Fetch file from SourceManager by path
```
whisperity: These comments here are useless. I'd rather have some higher level overview as to why exactly these things need to be queried and deleted.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,137 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+          // Empty deliberately
+          break;
+      }
+    }
+  }); // end of transaction
+}
+
 bool CppParser::parse()
 {
+  if(_ctx.options.count("incremental"))
```
whisperity: This is a leftover from the previous iteration where incrementality needed to be explicitly toggled, right? Mark a TODO here, please.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -358,6 +479,73 @@ bool CppParser::parse()
...
+    {
+      _ctx.fileStatus.insert(std::make_pair(loaded->path, IncrementalStatus::MODIFIED));
+      LOG(debug) << "File modified: " << loaded->path;
+
+      markAsModified(loaded);
+    }
+  }
+}
+
+std::set<model::CppNodeId> CppParser::collectNodeSet(model::CppNodeId node_) const
```
whisperity: What's a node set? Also, spaces and general formatting mismatch in this code too.

plugins/git/parser/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 include_directories(
   include
   ${PROJECT_SOURCE_DIR}/util/include
-  ${PROJECT_SOURCE_DIR}/parser/include)
+  ${PROJECT_SOURCE_DIR}/parser/include
+  ${PROJECT_SOURCE_DIR}/model/include)
```
whisperity: Why was this include added?

parser/include/parser/abstractparser.h
```diff
@@ -33,6 +33,8 @@ class AbstractParser
    */
   virtual std::vector<std::string> getDependentParsers() const = 0;
 
+  virtual void preparse(){}
```
whisperity: A documentation for this function should be given (and I agree `parse()`'s one is lacking too...). I'm not too confident a `void` return value is a good choice here (perhaps `bool` could indicate that the subclass parser failed to maintain its datasets and thus should not be called --- perhaps add this as a TODO).

Also, the default implementation of just *do nothing* feels weird.

util/src/dbutil.cpp
```diff
@@ -324,11 +335,16 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
       updateConnectionString(connStr_, "database", "postgres");
     std::string dbName = connStrComponent(connStr_, "database");
 
-    createPsqlDatbase(defaultPsqlConnStr, dbName);
-
-    int size = odbOpts.size();
-    db.reset(new odb::pgsql::database(size, cStyleOptions),
-      [](odb::database*){});
+    if(checkPsqlDatbase(defaultPsqlConnStr, dbName, create_))
```
whisperity: Space missing here.

util/src/dbutil.cpp
```diff
@@ -324,11 +335,16 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
...
+    {
+      db.reset(new odb::pgsql::database(optionsSize, cStyleOptions),
+               [](odb::database *) {});
```
whisperity: Extra space in type.

mcserep:
Thanks for remarks, we will look into them. Regarding the stress test of the cascaded deletion: it depends on the number of modified and deleted files. What is considered an average use case?

Regarding the commit sizes, I strongly believe that contribution of multiple authors through a version control system is most efficient with small (but consistent) commits and not a few large (major) commits. Large commits raises otherwise avoidable merge conflicts, and slows down collaboration between authors as they are unable to see modifications made by others.
I propose a discussion on project and version control management to further elaborate on this if needed.

plugins/cpp/parser/src/cppparser.cpp
mcserep: There was no previous iteration, this is the first, therefore contains some changes that were later dropped according to the project meeting discussions. E.g. the `--incremental` flag was first decided to be introduced, but later to be removed as CodeCompass should detect this automatically.
I advise to see the final changeset, not the single commits, the overall amount of changed codelines is not that much.

So, to clarify it: this is a `feature branch` (managed according to the [Git Feature Branching Model](https://nvie.com/posts/a-successful-git-branching-model/)), which is about to get merged into the `develop` branch (`master` in this case, because there is no `develop`).
Therefore it contains all changes of the feature branch, as normal.

mcserep: Yes, must have remained there from the previous iteration, when the `--incremental` flag was introduced, but later it was decided to be dropped and to automatically perform an incremental parsing when required.
The whole condition can be removed, as it just prints out debug information.

parser/src/parser.cpp
mcserep: The whole instruction goes like this:
```cpp
bool isNewDb = cc::util::connectDatabase(
  vm["database"].as<std::string>(), false) ? false : true;
```
which is basically the same. Could be simplified to your suggestion, I just didn't want to negate the returned shared pointer, as that would seem like a boolean value is returned and could be confusing.

plugins/git/parser/CMakeLists.txt
mcserep: It was added because some new includes (e.g. `model/File.h`) was added to `parsercontext.h` and therefore *gitparser* required the new include directory. Later `parsercontext.h` was separated into a header and body file to eliminate this added dependency, but it was left there. Good find!
Fixed in eaae9fdac3faadd46558e484a28a7930a32d3591.

parser/src/parsercontext.cpp
mcserep: Fixed in eaae9fdac3faadd46558e484a28a7930a32d3591.

whisperity:
First the measurement case should be something where every file is marked as `modified`. This would show the absolute extreme of how painful all those `ON DELETE CASCADE`s are. (Because previously a "full new parse" was done by throwing tables or throwing the entire database, which is a much easier operation.)

plugins/cpp/parser/src/cppparser.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

plugins/cpp/parser/src/cppparser.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/src/sourcemanager.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/src/parser.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/src/parser.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/src/parser.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/include/parser/parsercontext.h
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

util/src/dbutil.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

util/src/dbutil.cpp
intjftw: Solved in 4b52c300da8e3f8309356cf71e5f9c70207cc051

parser/src/parser.cpp
mcserep: Well, it is only 2 lines of (simple) code, so I don't think it was that much confusing; naming the original variable `topologicalOrder` would have been much more confusing, since an *in-place* reversing is performed afterwards. Anyway, refactored it to use revere iterators in 4bfa0e3011222b14b1e78e95eee12c253c7edbf8.

parser/src/sourcemanager.cpp
mcserep: Theoretically you are correct (and refactored in 4bfa0e3011222b14b1e78e95eee12c253c7edbf8), but I would be surprised to have an exception thrown there and even it is thrown, the `SourceManager`'s cache remains in an inconsistent state, so termination of the CodeCompass parser is advised.

parser/include/parser/parsercontext.h
whisperity: I think what's missing here is that this thing is "state in database VERSUS state on disk at parse time". The individual enumconstants are clean enough.

plugins/cpp/parser/src/cppparser.cpp
mcserep: After reconsideration we still think those comments are useful, because they guide through the reader what database tables will be affected by direct deletion without the need to read through the ~80 lines of code.
A higher level remark was added about these deletions in 1acae06bb70d50831c4887cfad6ef9df459b99cb.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Not needed after all, removed in 1acae06bb70d50831c4887cfad6ef9df459b99cb.

parser/include/parser/parsercontext.h
mcserep: Added in 7100014093aed0a1ae46a02ef6c1bff82a641139.

parser/src/parser.cpp
mcserep: Fixed in 4b52c300da8e3f8309356cf71e5f9c70207cc051.

parser/include/parser/abstractparser.h
mcserep: Good point, implemented in 7100014093aed0a1ae46a02ef6c1bff82a641139.

parser/src/parser.cpp
whisperity: This isn't fixed yet. Still, the tertiary operator feels too verbose here. How would you like the following as a compromise?

```cpp
bool isNewDb = cc::util::...(...) == false;
```

parser/src/parser.cpp
```diff
@@ -204,16 +213,32 @@ int main(int argc, char* argv[])
...
+  //--- Check database and project directory existence ---//
+
+  bool isNewDb = cc::util::connectDatabase(
+    vm["database"].as<std::string>(), false) ? false : true;
+  bool isNewProject = !checkProjectDir(vm);
+
+  if (isNewProject ^ isNewDb)
+  {
+    LOG(error)
+      << "Database and working directory existence are inconsistent. Use -f for reparsing!";
```
whisperity: are -> is

Specify '--force' to do a clean parse!

(Yet again, trying to eliminate the *reparse* phrase from here.)

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
   cc::parser::SourceManager srcMgr(db);
   cc::parser::ParserContext ctx(db, srcMgr, compassRoot, vm);
   pHandler.createPlugins(ctx);
 
+  // TODO: Handle errors returned by preparse().
+  std::vector<std::string> topologicalOrder = pHandler.getTopologicalOrder();
+  for (auto it = topologicalOrder.rbegin(); it != topologicalOrder.rend(); ++it)
+  {
+    LOG(info) << "[" << *it << "] preparse started!";
+    if(!pHandler.getParser(*it)->preparse())
```
whisperity: Space missing between if and paren.

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
+    {
+      LOG(error) << "[" << *it << "] preparse failed!";
+      return 2;
+    }
+  }
+
+  // TODO: Consider whether there is a better place for this code.
+  (cc::util::OdbTransaction(ctx.db))([&]
```
whisperity: Instead of trying to solve vexparse with Lisp, use uniform initialization.

```cpp
cc::util::OdbTransaction {ctx.db} ([&] {
```

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
+    for (auto& item : ctx.fileStatus)
+    {
+      switch (item.second)
+      {
+        case cc::parser::IncrementalStatus::MODIFIED:
+        case cc::parser::IncrementalStatus::DELETED:
+        {
+          LOG(info) << "Database cleanup: " << item.first;
+
+          // Fetch file from SourceManager by path
```
whisperity: No dot at the end of sentence.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+      db(db_),
+      srcMgr(srcMgr_),
+      options(options_),
+      compassRoot(compassRoot_)
+{
+  std::unordered_map<std::string, std::string> fileHashes;
+
+  (util::OdbTransaction(this->db))([&]
+   {
+     // Fetch all files from SourceManager
```
whisperity: .

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
+          cc::model::FilePtr delFile = srcMgr.getFile(item.first);
+
+          // Delete File and FileContent (only when no other File references it)
```
whisperity: .

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+                                {
+                                  return item->type == model::File::DIRECTORY_TYPE ||
+                                         item->type == model::File::BINARY_TYPE;
+                                }), files.end());
+
+     for (model::FilePtr file : files)
+     {
+       if (boost::filesystem::exists(file->path))
+       {
+         if (!this->fileStatus.count(file->path))
```
whisperity: `this->` not needed.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+         {
+           auto content = file->content.load();
```
whisperity: I take it that this *content* is the actual source code stored in the database. Do we really have to load this into memory just to compare if the local disk contains the same hash? The foreign key IIRC maps file content hashes to the CLOB. Perhaps `file->content.object_id()` works here, as the hash is the primary key on the content table.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+           fileHashes[file->path] = content != nullptr ? content->hash : "";
+           if (content == nullptr)
+             continue;
+
+           std::ifstream fileStream(file->path);
+           std::string fileContent(
+               (std::istreambuf_iterator<char>(fileStream)),
```
whisperity: Vexparse parentheses can be replaced with uniform initialization again:

```cpp
std::string fileContent(
  std::istreambuf_iterator<char>{fileStream},
  std::istreambuf_iterator<char>{});
```

(And indentation is off.)

----
I say this stort of string initialization is **ugly**, because if you mess up the vexparse you don't get compile error, but rather some nice converting constructor and weirdo results.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+               (std::istreambuf_iterator<char>()));
+           fileStream.close();
+
+           if (content->hash != util::sha1Hash(fileContent))
+           {
+             this->fileStatus.insert(
```
whisperity: If we are sure that every file path is only added once, to prevent string memory operations, let's use forwarding and in-place construction with [`emplace`](https://en.cppreference.com/w/cpp/container/map/emplace) instead of `insert`:

    fileStatus.emplace(file->path, ...);

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+               std::make_pair(file->path, cc::parser::IncrementalStatus::MODIFIED));
+             LOG(debug) << "File modified: " << file->path;
+           }
+         }
+       }
+       else
+       {
+         this->fileStatus.insert(
```
whisperity: Same here about `emplace`, and the `this->` is not needed.

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,30 @@ void SourceManager::updateFile(const model::File& file_)
     });
 }
 
+void SourceManager::removeFile(const model::File& file_)
+{
+  bool removeContent = false;
+
+  // Delete File and FileContent (only when no other File references it)
```
whisperity: .

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,30 @@ void SourceManager::updateFile(const model::File& file_)
...
+  _transaction([&, this]() {
```
whisperity: Why is `this` captured here, if everything also captured by reference?

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,30 @@ void SourceManager::updateFile(const model::File& file_)
...
+    auto relFiles = _db->query<model::File>(
+        odb::query<model::File>::content == file_.content.object_id());
+    if(std::distance(relFiles.begin(), relFiles.end()) == 1)
```
whisperity: Space missing before paren. Also, [`odb::result<T>`](https://www.codesynthesis.com/products/odb/doc/manual.xhtml#4.4) has a `size()` getter.

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,30 @@ void SourceManager::updateFile(const model::File& file_)
...
+    {
+      removeContent = true;
+      _db->erase<model::FileContent>(file_.content.object_id());
+    }
+    _db->erase<model::File>(file_.id);
+  });
+
+  // Maintain cache
```
whisperity: .

parser/src/sourcemanager.cpp
```diff
@@ -221,6 +234,30 @@ void SourceManager::updateFile(const model::File& file_)
...
+  std::lock_guard<std::mutex> guard(_createFileMutex);
+  _files.erase(file_.path);
```
whisperity: For visual indication of the critical section, because it's not the whole function which should be *synchronized*, please put the statements into a new block:

```cpp
  std::lock_guard<std::mutex> guard(_createFileMutex);
  {
    _files.erase(file_.path);
    // ...
  }
}

File: plugins/cpp/parser/src/cppparser.cpp [hunk 38]
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+{
+  std::vector<model::FilePtr> filePtrs(_ctx.fileStatus.size());
+
+  std::transform(_ctx.fileStatus.begin(),
+                 _ctx.fileStatus.end(),
+                 filePtrs.begin(),
+                 [this](const auto& item)
+                 {
+                   if (item.second == IncrementalStatus::MODIFIED)
+                   {
```
whisperity: Please remove the brackets from these one-liner branches.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+                   }
+                   else
+                   {
+                     return std::make_shared<model::File>();
+                   }
+                 });
+
+  try
+  {
+    (util::OdbTransaction(_ctx.db))([&]
```
whisperity: Same comment about the vexparse parens, turn them into uniform init.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+    {
+      for (const model::FilePtr file : filePtrs)
+      {
+        if(file)
```
whisperity: Space missing, statement has redundant brackets both for `for` and `if`.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+                   {
+                     return std::make_shared<model::File>();
+                   }
+                 });
+
+  try
+  {
+    (util::OdbTransaction(_ctx.db))([&]
+    {
+      for (const model::FilePtr file : filePtrs)
```
whisperity: Capturing this shared ptr via copy is way too expensive for this simple operation! The `filePtrs` lives on the local stack so it's impossible for the last reference to go out of scope and destroy these objects. And because the type of the elements is self-explanatory, and only a few lines above, you can safely use `auto` without sacrificing readability.

    for (const auto& file : filePtrs)

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+        {
+          markAsModified(file);
+        }
+      }
+
+      for (auto& item : _ctx.fileStatus)
+      {
+        switch (item.second)
+        {
+          /*
```
whisperity: Multiline comments should be reserved for documentation comments and high level (such as namespace or whole file) stuff. Also, please ensure you use your space until the 80 column barrier before doing the line break. These 5 lines I think can safely be reduced to 2.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+           * In case the file status is MODIFIED or DELETED,
+           * delete the related records to the File entity
+           * from the C++ parser's model.
```
whisperity: It's not just the parser's model, but the whole plugin's.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+           */
+          case IncrementalStatus::MODIFIED:
+          case IncrementalStatus::DELETED:
+          {
+            LOG(info) << "[cppparser] Database cleanup: " << item.first;
+
+            // Fetch file from SourceManager by path
```
whisperity: .

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            model::FilePtr delFile = _ctx.srcMgr.getFile(item.first);
+
+            // Query CppAstNode
```
whisperity: .

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            auto defCppAstNodes = _ctx.db->query<model::CppAstNode>(
+              odb::query<model::CppAstNode>::location.file == delFile->id &&
+              odb::query<model::CppAstNode>::astType == model::CppAstNode::AstType::Definition);
+            for (const model::CppAstNode &astNode : defCppAstNodes)
```
whisperity: `const auto& astNode : defCppAstNodes` works here too.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            {
+              // Delete CppEntity
+              auto delCppEntities = _ctx.db->query<model::CppEntity>(
+                odb::query<model::CppEntity>::astNodeId == astNode.id);
+              for (const model::CppEntity &entity : delCppEntities)
```
whisperity: `const auto&`

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+              {
```
whisperity: No need for bracket.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            // Fetch file from SourceManager by path
+            model::FilePtr delFile = _ctx.srcMgr.getFile(item.first);
+
+            // Query CppAstNode
+            auto defCppAstNodes = _ctx.db->query<model::CppAstNode>(
+              odb::query<model::CppAstNode>::location.file == delFile->id &&
+              odb::query<model::CppAstNode>::astType == model::CppAstNode::AstType::Definition);
+            for (const model::CppAstNode &astNode : defCppAstNodes)
+            {
+              // Delete CppEntity
```
whisperity: Comment is redundant.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+                odb::query<model::CppEntity>::astNodeId == astNode.id);
+              for (const model::CppEntity &entity : delCppEntities)
+              {
+                _ctx.db->erase<model::CppEntity>(entity.id);
+              }
+
+              // Delete CppInheritance
+              auto delCppInheritance = _ctx.db->query<model::CppInheritance>(
+                odb::query<model::CppInheritance>::derived == astNode.mangledNameHash);
+              for (const model::CppInheritance &inheritance : delCppInheritance)
```
whisperity: `const auto& variable`, the `&` is in the wrong spot. No need for the comment, no need for the brackets, this applies to the whole bigger loop.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+
+              // Delete CppNode (connected to CppAstNode)
+              auto delNodes = _ctx.db->query<model::CppNode>(
+                odb::query<model::CppNode>::domainId == std::to_string(astNode.id) &&
+                odb::query<model::CppNode>::domain == model::CppNode::CPPASTNODE);
+              for (model::CppNode &node : delNodes)
+              {
+                _ctx.db->erase<model::CppNode>(node.id);
+                // TODO: handle that RelationCollector will try to reinsert the other end of the relation
+                // Raises "[WARNING] object already persistent"
```
whisperity: Has it been checked that this warning is only a nuisance but does not break actual functionality?

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            for (model::CppNode &node : delNodes)
+            {
+              _ctx.db->erase<model::CppNode>(node.id);
+              // TODO: handle that RelationCollector will try to reinsert the other end of the relation
+              // Raises "[WARNING] object already persistent"
+            }
+          }
+
+          case IncrementalStatus::ADDED:
+            // Empty deliberately
```
whisperity: .

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+            break;
+        }
+      }
+    }); // end of transaction
```
whisperity: E
.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+  }
+  catch(odb::database_exception&)
```
whisperity: space

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -358,6 +484,32 @@ bool CppParser::parse()
   return success;
 }
 
+void CppParser::initBuildActions()
+{
+  (util::OdbTransaction(_ctx.db))([&, this] {
```
whisperity: Vexing parse parens -> uniform, again.

whisperity: The `this` need not be captured, I think.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -358,6 +484,32 @@ bool CppParser::parse()
...
+    for (const model::BuildAction& ba : _ctx.db->query<model::BuildAction>())
```
whisperity: As much as possible, don't repeat ourselves: `const auto& action`

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -358,6 +484,32 @@ bool CppParser::parse()
...
+
+void CppParser::markAsModified(const model::FilePtr file_)
+{
+  auto inclusions = _ctx.db->query<model::CppHeaderInclusion>(
+    odb::query<model::CppHeaderInclusion>::included == file_->id);
+
+  for (auto inc : inclusions)
+  {
+    model::FilePtr loaded = inc.includer.load();
+    if(_ctx.fileStatus.count(loaded->path) == 0)
```
whisperity: space, and `!....count(...)` should express the same thing

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -358,6 +484,32 @@ bool CppParser::parse()
...
+    {
+      _ctx.fileStatus.insert(std::make_pair(loaded->path, IncrementalStatus::MODIFIED));
```
whisperity: Emplace

plugins/metrics/parser/src/metricsparser.cpp
```diff
@@ -51,6 +51,40 @@ std::vector<std::string> MetricsParser::getDependentParsers() const
   return std::vector<std::string>{};
 }
 
+bool MetricsParser::preparse()
+{
+  if(!_fileIdCache.empty())
```
whisperity: space

plugins/metrics/parser/src/metricsparser.cpp
```diff
@@ -51,6 +51,40 @@ std::vector<std::string> MetricsParser::getDependentParsers() const
...
+  {
+    try
+    {
+      (util::OdbTransaction(_ctx.db))([this] {
```
whisperity: Vexparse -> uniform

plugins/metrics/parser/src/metricsparser.cpp
```diff
@@ -51,6 +51,40 @@ std::vector<std::string> MetricsParser::getDependentParsers() const
...
+        for(const model::File& file
```
whisperity: space, `auto`

plugins/metrics/parser/src/metricsparser.cpp
```diff
@@ -51,6 +51,40 @@ std::vector<std::string> MetricsParser::getDependentParsers() const
...
+          {
+            LOG(info) << "[metricsparser] Database cleanup: " << file.path;
+
+            _ctx.db->erase_query<model::Metrics>(odb::query<model::Metrics>::file == file.id);
+            _fileIdCache.erase(file.id);
+          }
+        }
+      });
+    }
+    catch(odb::database_exception&)
```
whisperity: space

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -329,19 +332,142 @@ int CppParser::worker(const clang::tooling::CompileCommand& command_)
...
+          case IncrementalStatus::ADDED:
+            // Empty deliberately
+            break;
+        }
+      }
+    }); // end of transaction
+  }
+  catch(odb::database_exception&)
+  {
+    LOG(fatal) << "Transaction failed!";
```
whisperity: `failed in C++ parser` instead, if `metrics` also mentions who they are.

plugins/metrics/parser/src/metricsparser.cpp
```diff
@@ -51,6 +51,40 @@ std::vector<std::string> MetricsParser::getDependentParsers() const
...
+  {
+    try
+    {
+      (util::OdbTransaction(_ctx.db))([this] {
+        for(const model::File& file
+          : _ctx.db->query<model::File>(
+          odb::query<model::File>::id.in_range(_fileIdCache.begin(), _fileIdCache.end())))
+        {
+          auto it = _ctx.fileStatus.find(file.path);
+          if(it != _ctx.fileStatus.end() &&
```
whisperity: space

util/include/util/dbutil.h
```diff
@@ -5,6 +5,7 @@
 #include <string>
 
 #include <odb/database.hxx>
+#include <odb/sqlite/exceptions.hxx>
```
whisperity: This include should be guarded with `DATABASE_SQLITE`. There are no guarantees for this header to exist, if someone is only compiling for Postgres, they need not have the SQLite connector library's headers.

util/src/dbutil.cpp
```diff
@@ -281,39 +290,41 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
...
+        SQLITE_UTF8,
+        nullptr,
+        &sqliteRegexImpl,
+        nullptr,
+        nullptr,
+        nullptr);
+    }
+    catch(odb::database_exception& e)
+    {
+      // Could not connect to DB
```
whisperity: .

util/src/dbutil.cpp
```diff
@@ -281,39 +290,41 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
...
+        sqlitePtr,
+        "regexp", 2, // regexp function with one argument
+        SQLITE_UTF8,
+        nullptr,
+        &sqliteRegexImpl,
+        nullptr,
+        nullptr,
+        nullptr);
+    }
+    catch(odb::database_exception& e)
```
whisperity: Space

util/src/dbutil.cpp
```diff
@@ -324,11 +335,16 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
       updateConnectionString(connStr_, "database", "postgres");
     std::string dbName = connStrComponent(connStr_, "database");
 
-    createPsqlDatbase(defaultPsqlConnStr, dbName);
-
-    int size = odbOpts.size();
-    db.reset(new odb::pgsql::database(size, cStyleOptions),
-      [](odb::database*){});
+    if (checkPsqlDatbase(defaultPsqlConnStr, dbName, create_))
+    {
```
whisperity: not needed `{`

util/src/dbutil.cpp
```diff
@@ -324,11 +335,16 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
...
+      db.reset(new odb::pgsql::database(optionsSize, cStyleOptions),
+               [](odb::database*) {});
+    }
+    else
+    {
+      // DB does not exists
```
whisperity: .

util/src/dbutil.cpp
```diff
@@ -324,11 +335,16 @@ std::shared_ptr<odb::database> createDatabase(const std::string& connStr_)
...
-    int size = odbOpts.size();
-    db.reset(new odb::pgsql::database(size, cStyleOptions),
-      [](odb::database*){});
+    if (checkPsqlDatbase(defaultPsqlConnStr, dbName, create_))
+    {
+      db.reset(new odb::pgsql::database(optionsSize, cStyleOptions),
+               [](odb::database*) {});
+    }
+    else
+    {
```
whisperity: No need for brackets.

parser/src/parser.cpp
mcserep: In that case I think checking the equality against `nullptr` would still be better, otherwise it would still ambiguously imply that `connectDatabase()` returns with a boolean value.

.gitignore
```diff
@@ -17,3 +18,4 @@ nbproject
 # Other
 
 /build/
+/build_*/
```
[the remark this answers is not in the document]
mcserep: IDE specific notations were already present in the .gitignore file (there is a whole *"IDEs and editors"* section), so I don't really get the first part of this change request.

I added the `build_*` folder to the .gitignore file, because `build` was already there, but a separate folder for SQLite and PgSQL configuration was useful during the development process, since we had to test against both.
The `build` and `build_*` folders are not related to any IDE, but a developer-specific stuff, simply the most common names for the CMake build folders and are widely ignored in Git repositories.

Can be merged into a single line, will do it later.

doc/usage.md
```diff
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
...
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a complete compilation database, 
+therefore the build instructions for the modified files are not sufficient.
+In case of CMake, the CMAKE_EXPORT_COMPILE_COMMANDS module's output always contains all files.
+
+Currently `cppparser` and `metricsparser` support incremental parsing, while other parsers
+just execute a forced reparse.
```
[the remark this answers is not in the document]
mcserep: I know about another developed feature named *reparse*, but right now the **forced parse is named reparse at multiple locations**, so I aimed to be consistent. This name collision should be solved sooner or later, but that is related to the  *reparse* feature and not the *incremental parsing*.

parser/src/parser.cpp
mcserep: I just copied the "*Use -f for reparsing!*" part from other error messages present in the project, to be consistent with them. Again, I think this name collision issue regarding the *reparse* phrase should not be part of this pull request.

parser/src/parser.cpp
mcserep: This syntax is present at 8 other locations in the codebase (authored by project **members**!) , so I assumed it is acceptable. Still think so and this level of coding style should not be target for any change request.

parser/src/parser.cpp
mcserep: You are joking now, right? Aside that requesting a change at a such a low level (comment style) is comedic; single line, short comments nowhere end in punctuation dot in the CodeCompass codebase last time I checked.

parser/src/parsercontext.cpp
mcserep: Guess you are not joking. Still should not be a change request at the very first place.

parser/src/parser.cpp
mcserep: See above.

parser/src/parsercontext.cpp
mcserep: Good point :thumbsup:

parser/src/sourcemanager.cpp
mcserep: See above.

parser/src/sourcemanager.cpp
mcserep: See above,

parser/src/parser.cpp
whisperity: Thank you for counting it. You might also want to *randomly push a change that resolves line endings* like some [other guy did to the LLVM repository recently](http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20180730/236784.html)?

You assumed it was acceptable. Sure, I can understand it.

But now you were told it is not.

Also, project members' knowledge and code styles evolve over time, it's not like everyone is born with *x* amount of knowledge and then it stays constant. Why should **more** technical debt be added that is to be weeded out later.

Code is getting touched right now, so it should be ensured that the code that is getting touched *now* is up to some standards.
(Some standard that, may yet evolve in the future, agreeably.)

parser/src/parser.cpp
whisperity: Style-based fixup requests are common, see the comment above about what usually happens in the LLVM community too.

parser/src/sourcemanager.cpp
mcserep: Everything should be synchronized after the creation of the guard, and that is the visual indication. (That is 4 lines of code, it's not like one can miss it visually.)
Also, the visual indication of a critical section should match it functionally, so in that case the guard should be created inside the new block I think.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Same response.

parser/src/sourcemanager.cpp
whisperity: Yeah, do that. Must have had some Java on my mind with the whole `synchronized`.

plugins/cpp/parser/src/cppparser.cpp
mcserep: The [coding convention](https://github.com/Ericsson/CodeCompass/blob/master/doc/coding_conventions.md) states nothing on this matter and the existing codebase is also very inconsistent on the usage of brackets in such cases.

Since no uniform coding style is applied regarding this, a personal coding style preference should not be reflected as change requests.

whisperity: Yeah, because obviously in a project where code reviews hang for **multiple years** and where ***zero** developers exist officially* a code style change could go in any time, right?

parser/src/parser.cpp
mcserep: There is no fixup, your suggestion is **against** the applied coding style convention of the project.

parser/src/parser.cpp
mcserep: I am not about to comply with a coding style convention you made up yourself and:
- the existing codebase does not comply with it,
- the other project members do not comply with it,
- the official coding convention of the project does not mention it.

These kind of nonsense change requests are just burning valuable research and development time for the sake of nothing.
Should be discussed on next project meeting if you disagree.

plugins/cpp/parser/src/cppparser.cpp
mcserep: We are not a mind-readers @whisperity , we cannot comply with coding conventions which are not mentioned anywhere and cannot be deduced from the codebase, because it is inconsistent.

If you think that the current coding convention is not broad or strict enough, please suggest its extension (we have weekly meetings!), but requesting changes on a developed feature based on your **personal** preferences is both unprofessional and unacceptable.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

plugins/cpp/parser/src/cppparser.cpp
mcserep: See above.

util/src/dbutil.cpp
mcserep: See above.

util/src/dbutil.cpp
mcserep: See above.

util/src/dbutil.cpp
mcserep: See above.

util/src/dbutil.cpp
mcserep: See above.

mcserep:
Unfortunately I have no surprises you had difficulties with rebasing it, that is the reason I was against it in the first place.

Right now @whisperity , you have made 80 change requests for the ~600 lines of changed code. Most of them are about code formatting, styling and other coding conventions. I understand that code styling is important to keep the codebase clean and nice, but you cannot expect that other contributors will comply with coding conventions which they do not know about, and are not agreed upon. Again, you shouldn't enforce your personal coding convention preferences (even if they are good) without discussing and accepting them with the other project members, because simply it won't work out.
You must understand that refactoring/rewriting the code submitted for pull request in multiple iterations, just because a reviewer clings on non-documented, non-agreed coding convention rules is not only frustrating for the authors, but is also a highly inefficient development mechanism, requiring a lot of extra time.

Let's elaborate on this on the next weekly meeting, because further pull requests are coming, and these situations will just repeat. And we are still not mind-readers.

util/src/dbutil.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

util/include/util/dbutil.h
mcserep: Must have gotten there accidentally, because all SQLite specific things are in the `.cpp` file. Removed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/metrics/parser/src/metricsparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/metrics/parser/src/metricsparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/metrics/parser/src/metricsparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Yes. Also, just for the later record, with #274 this will need some refactoring. (But first one of the pull request has to be accepted, so the fix can go into the other one.)

parser/src/sourcemanager.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/sourcemanager.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/sourcemanager.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/parsercontext.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/parsercontext.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/parsercontext.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/parser.cpp
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/include/parser/abstractparser.h
```diff
@@ -33,6 +33,15 @@ class AbstractParser
    */
   virtual std::vector<std::string> getDependentParsers() const = 0;
 
+  /**
+   * Maintains and cleans up the database in preparation of
+   * incremental parsing.
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

model/include/model/buildsourcetarget.h
```diff
@@ -40,6 +41,7 @@ struct BuildTarget
   FilePtr file;
 
   #pragma db not_null
+#pragma db on_delete(cascade)
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

doc/usage.md
```diff
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
...
 
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a complete compilation database, 
+therefore the build instructions for the modified files are not sufficient.
+In case of CMake, the CMAKE_EXPORT_COMPILE_COMMANDS module's output always contains all files.
+
+Currently `cppparser` and `metricsparser` support incremental parsing, while other parsers
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

doc/usage.md
```diff
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
   -j 4
 ```
 
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a complete compilation database, 
+therefore the build instructions for the modified files are not sufficient.
+In case of CMake, the CMAKE_EXPORT_COMPILE_COMMANDS module's output always contains all files.
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

doc/usage.md
```diff
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
   -j 4
 ```
 
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a complete compilation database, 
+therefore the build instructions for the modified files are not sufficient.
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

doc/usage.md
```diff
@@ -165,6 +165,17 @@ keepalive CodeCompass_parser \
   -j 4
 ```
 
+### Incremental parsing
+
+As an experimental feature CodeCompass supports incremental parsing, updating an
+existing database and project workspace by detecting the added, deleted and modified files.  
+Incremental parsing depends on that the build tooling generates a complete compilation database,
```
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

.gitignore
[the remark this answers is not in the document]
mcserep: Fixed in 64cb407581662e78675baf916a95dba2b765b3e4.

parser/src/parsercontext.cpp
mcserep: Fixed in 52b7b72b31bdc585c43440a29071f0842e4570b6.

parser/src/sourcemanager.cpp
mcserep: Fixed in 52b7b72b31bdc585c43440a29071f0842e4570b6.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Thanks for the remark on how the stack and references in C++ work, it was just a typo. Fixed in 52b7b72b31bdc585c43440a29071f0842e4570b6.

bruntib:
My opinion about the styling issues: me too, I'm quite compulsive about the coding standards and it was one of the most important goals when rewriting CodeCompass, to keep good code quality. However, it is true, that the rules are not written down anywhere, so requiring a style change can easily seem to be an ad-hoc requirement which comes out of nowhere. Even now there are some style issues in the code (longer lines than 80 characters, indentation mismatch) which I'd like to be fixed, but we should save them until a linter tool is introduced to a CI system, which I would totally support. I know that clang-format can even rewrite code fragments, but I think that would be too strong. Can it be set to work in check-only mode? Or does anyone know similar tools like pep8 for Python?

bruntib:
One more note: here is a coding style guide: https://github.com/Ericsson/CodeCompass/blob/master/doc/coding_conventions.md. This could be extended with the contested cases and checked by an automatized tool.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 4301dea623f06401b259435c3999d35a3f646dd6.

plugins/cpp/parser/src/cppparser.cpp
mcserep: Fixed in 4301dea623f06401b259435c3999d35a3f646dd6.

plugins/metrics/parser/src/metricsparser.cpp
mcserep: Fixed in 4301dea623f06401b259435c3999d35a3f646dd6.

parser/src/parser.cpp
mcserep: Vexparse fixed in 4301dea623f06401b259435c3999d35a3f646dd6.

mcserep:
Squashed the commits of the change requests.

parser/src/parser.cpp
mcserep: Fixed in 1c4b8809a7139dc3519037d42d650b2e0e339686.

mcserep:
Thanks for the review @bruntib!
Please modify the code accordingly @intjftw at your earliest convenience.

parser/src/parsercontext.cpp
```diff
@@ -0,0 +1,81 @@
...
+      options(options_),
+      compassRoot(compassRoot_)
+{
+  std::unordered_map<std::string, std::string> fileHashes;
+
+  (util::OdbTransaction(this->db))([&]
+   {
+     // Fetch all files from SourceManager
+     std::vector<model::FilePtr> files = this->srcMgr.getAllFiles();
+     files.erase(std::remove_if(files.begin(), files.end(),
```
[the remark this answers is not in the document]
mcserep: Proposed method signature:
```cpp
template <typename Filter = AllFilesFilter>
std::vector<model::FilePtr> SourceManager::getFiles(const Filter& beta = Filter())
```
where `AllFilesFilter` could be a nested functor inside `SourceManager`, like:
```cpp
struct AllFilesFilter
{
  bool operator()(model::FilePtr file) { return true; }
};
```

Any remarks @bruntib ?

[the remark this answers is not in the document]
bruntib: Yes, it looks good. Note that the `operator()` has to be a const member function if the functor is passed by constant reference, but besides that it's fine.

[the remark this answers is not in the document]
intjftw: Method added in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/src/parser.cpp
```diff
@@ -78,11 +79,24 @@ po::options_description commandLineArguments()
...
+ * based on the given command line arguments.
+ * @return The path of the project directory.
+ */
+bool checkProjectDir(const po::variables_map& vm_)
+{
+  const std::string projDir
+    = vm_["workspace"].as<std::string>() + '/'
+      + vm_["name"].as<std::string>();
+
+  return fs::exists(projDir) && fs::is_directory(projDir);
```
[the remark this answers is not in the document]
intjftw: Fixed in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/include/parser/parsercontext.h
```diff
@@ -1,8 +1,12 @@
 #ifndef CC_PARSER_PARSERCONTEXT_H
 #define CC_PARSER_PARSERCONTEXT_H
 
+#include <fstream>
```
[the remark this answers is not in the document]
intjftw: Fixed in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/src/parser.cpp
```diff
@@ -78,11 +79,24 @@ po::options_description commandLineArguments()
   return desc;
 }
 
+/**
+ * This function checks the existence of the workspace and project directory
+ * based on the given command line arguments.
+ * @return The path of the project directory.
```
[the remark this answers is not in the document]
intjftw: Fixed in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
+        }
+        case cc::parser::IncrementalStatus::ADDED:
+          // Empty deliberately
+          break;
+      }
+    }
+  });
+
   // TODO: Handle errors returned by parse().
   for (const std::string& parserName : pHandler.getTopologicalOrder())
```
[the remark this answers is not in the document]
intjftw: Fixed in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/src/parser.cpp
```diff
@@ -225,24 +250,64 @@ int main(int argc, char* argv[])
...
+    {
+      LOG(error) << "[" << *it << "] preparse failed!";
+      return 2;
+    }
+  }
+
+  // TODO: Consider whether there is a better place for this code.
+  (cc::util::OdbTransaction(ctx.db))([&]
+  {
+    for (auto& item : ctx.fileStatus)
```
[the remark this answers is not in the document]
intjftw: Fixed in 79bf56576af8f61faff59097ec450fff1bc48137.

parser/src/parsercontext.cpp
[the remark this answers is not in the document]
intjftw: Use of multiple loops eliminated in 686c8225fcce4ecfa52230a3af14229bc164175a.

PR #270, opened 2018-07-19
Enhance build generation and constrict third-party library scope

whisperity:


whisperity:
Other PRs superseded this.

whisperity:
Not yet completely superseded, however. What about the `add_thrift_library` CMake function, @mcserep? Just pinging here to set a reminder that the other patches not yet fully covver the functionality of this proposal.

mcserep:
That's right. The `add_thrift_library` function reduces the code redundancy in `CMakeLists.txt` files a little, and adds a dependency chain feature; however that is rarely used, since there is only 2 layers of dependency right now. Therefore for me it seems that while the goal of this new CMake function is clear, this refactoring requires more work than the benefits it provides.

Right now I tried to focus on the issues hindering the development process (like generating ODB and Thrift files into the wrong directory, see #323 and #341), but of course it can be addressed in a separate PR later.

steakhal:
Hi all, I'm failing to build (using the  `Ninja` generator) due to the poor dependency handling in `CMakeLists.txt`s where the sources are generated by `thrift`.

To fix this, we have to specify all the generated files.
This would also benefit from the imaginary `add_thrift_library` command.
I would like to continue @whisperity's work if you all think that this would be the right direction.

PR #272, opened 2018-07-23
Show parameters for member functions, cross out explicit deletes, don't show implicits "-1" line

### LINKED ISSUE #302: Location of an implicit node
The Info Tree presents all usages of a type. In case of a compiler generated copy and move constructor the parameter of these functions are also the usages of the given type. Do we present them in the Info Tree? If we do, then where do we jump when clicking on them? Their position in the database is at -1 line and column.

whisperity:
> Closes #302.

 - Fixed an issue with the InfoTree for a C++ function returned the name of the function as *signature*, as opposed to the real signature stored in the database.
 - The listing of methods for a type listed only the name and return value of methods, without their parameters. This made distinguishing overloads impossible: from a glance at the method list, one could not deduce what overloads exist for the type.
 - The "source range" for implicit methods is invalid (`-1` is saved and returned), and thus isn't a useful information to show on the API.
 - If the user explicitly `= delete`s an implicit generated function or an overload, show the delete in the info tree with a new special marker.

Current behaviour with the problems:
image
image
image

New behaviour introduced by this patch:
image
image
image

### DIFF
--- a/plugins/cpp/model/include/model/common.h
+++ b/plugins/cpp/model/include/model/common.h
@@ -23,7 +23,8 @@ enum Tag
   Static  = 3,
   Implicit = 4,
   Global = 5,
-  Constant = 6
+  Constant = 6,
+  UserDeleted = 7
 };
 
 inline std::string visibilityToString(Visibility v_)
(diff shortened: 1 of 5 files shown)


mcserep:
Feel free to reopen it if further progress has been done.

PR #278, opened 2018-09-03
Follow Clang signature changes

bruntib:
A function signature changed in Clang which must be followed in our
overload.

### DIFF
--- a/plugins/cpp/parser/src/ppincludecallback.cpp
+++ b/plugins/cpp/parser/src/ppincludecallback.cpp
@@ -65,7 +65,8 @@ void PPIncludeCallback::InclusionDirective(
   const clang::FileEntry*,
   clang::StringRef searchPath_,
   clang::StringRef,
-  const clang::Module*)
+  const clang::Module*,
+  clang::SrcMgr::CharacteristicKind)
 {
   if (searchPath_.empty())
     return;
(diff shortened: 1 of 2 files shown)


whisperity:
I'll wait with merging this once Clang has officially released and we can update the user guide for a working download URL.

util/src/dbutil.cpp
```diff
@@ -156,6 +156,7 @@ bool checkPsqlDatbase(const std::string& connStr_, const std::string dbName_, bo
     {
       std::string createCmd = "CREATE DATABASE " + dbName_
         + " ENCODING = 'SQL_ASCII'"
+        + " LC_CTYPE='C'"
```
whisperity: How did this not cause an error before? :hushed:

bruntib: The error may depend on the host machine encoding settings and the PostgreSQL server encoding settings. The error may not occur if you start an own PostgreSQL server, but occurs for me on a common server machine where the database system was installed from Linux repository.
Should I send this modification in a separate pull request while we're waiting for the Clang release?

whisperity: Rather, I think you should remove the interface-specific changes and only keep this one, and then update the commit and the PR's name and title accordingly.

I have already prepared the fix for the LLVM 7.0 release on a branch of mine, just was also waiting for the release. :blush:

whisperity: I remember this `CTYPE` thing, I have added it to the docs in the most recent change when I upgraded the wording and expressivity of the usage guide. Thank you for finding a practical point in the code where it also fails.

(I used to manually drop and create my DBs from a shell script when testing...)

whisperity:
Now that LLVM 7.0 has released tomorrow I'll update my branch where I prepared these changes and submit a pull request.

PR #282, opened 2018-09-22
User shouldn't create DB manually

bruntib:
The database is created by CodeCompass. Since incremental parsing
feature the existence of the database means that an earlier parse
already exists which is now being updated.


mcserep:
Since it's only stated for SQLite that _"embedded library will take care of creating and using the database file"_, I think further clarification could be useful, that the database is always created by CodeCompass.

bruntib:
Done.

PR #291, opened 2018-10-19
Update deps.md

Benzy77:
This change can let developer to more clearly understand how to build and set environmen in their computer.

### DIFF
--- a/doc/deps.md
+++ b/doc/deps.md
@@ -59,7 +59,7 @@ by other processes which could, in extreme cases, make the system very hard or
 impossible to recover. **Please do NOT add a `sudo` in front of any `make` or
 other commands below, unless *explicitly* specified!**
 
-### Thrift
+### Step1 Thrift
 CodeCompass needs [Thrift](https://thrift.apache.org/) which provides Remote
 Procedure Call (RPC) between the server and the client. Thrift is not part of
 the official Ubuntu 16.04 LTS repositories, but you can download it and build
@@ -91,7 +91,7 @@ cd thrift-<version>
 make install
 ```
 
-### LLVM/Clang
+### Step2 LLVM/Clang
 In Ubuntu 16.04 LTS the LLVM/Clang has some packaging issues, i.e. some libs
 are searched in `/usr/lib` however the package has these in
 `/usr/lib/llvm-3.8/lib` (see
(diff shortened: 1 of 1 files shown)


doc/deps.md
```diff
@@ -143,7 +147,7 @@ download and recompile ODB using a new C++ compiler.
 wget http://www.codesynthesis.com/download/odb/2.4/libodb-2.4.0.tar.gz
 tar -xvf libodb-2.4.0.tar.gz
 cd libodb-2.4.0
-./configure --prefix=<odb_install_dir>
+./configure --prefix=<libodb_install_dir>
```
whisperity: These **do not** need to go into separate folders. It'd be much better if all the 3rd-party dependencies went into the same folder (e.g. Clang, Thrift and ODB into the same "virtual root prefix")... This will create even more confusion, as ODB is something that is integrated together and should exist together (so when you need to upgrade or uninstall or change something you don't have to hunt down 7 different folders), I really don't agree with this change...

doc/deps.md
```diff
@@ -122,14 +124,16 @@ cd build
...
   ../llvm
+  
+#This clang_install_dir maybe confuse with above clang.They are different.
 # This make step takes a while. If you have more CPUs then you can compile on
 # several threads with -j<number_of_threads> flag.
 make install
 ```
 
-### ODB
+### Step3 ODB
```
whisperity: Format these lines properly: `Step 3. ODB`

doc/deps.md
```diff
@@ -217,17 +220,17 @@ mkdir <gtest_install_dir>/lib
...
-export GTEST_ROOT=<gtest_install_dir>
-export CMAKE_PREFIX_PATH=<thrift_install_dir>:$CMAKE_PREFIX_PATH
-export CMAKE_PREFIX_PATH=<clang_install_dir>:$CMAKE_PREFIX_PATH
-export CMAKE_PREFIX_PATH=<odb_install_dir>:$CMAKE_PREFIX_PATH
-export PATH=<thrift_install_dir>/bin:$PATH
-export PATH=<odb_install_dir>/bin:$PATH
+export GTEST_ROOT=<gtest_install_dir>                                      #Step5 <gtest_install_dir>
+export CMAKE_PREFIX_PATH=<thrift_install_dir>:$CMAKE_PREFIX_PATH           #Step1 <thrift_install_dir>
+export CMAKE_PREFIX_PATH=<clang_install_dir>:$CMAKE_PREFIX_PATH            #Step2 <clang_install_dir>
+export CMAKE_PREFIX_PATH=<odb_install_dir>:$CMAKE_PREFIX_PATH              #Step4 <odb_install_dir>
```
whisperity: If you put SQLite_ODB and Postgres_ODB and the likes into separate folders, this will not work, as at link time the libraries won't be bound properly.

doc/deps.md
whisperity: These **do not** need to go into separate folders. It'd be much better if all the 3rd-party dependencies went into the same folder (e.g. Clang, Thrift and ODB into the same "virtual root prefix")... This will create even more confusion, as ODB is something that is integrated together and should exist together (so when you need to upgrade or uninstall or change something you don't have to hunt down 7 different folders), I really don't agree with this change...

mcserep:
Closed since important change requests has not been addressed by original contributor.

PR #299, opened 2018-11-25
Incremental parsing fix

### LINKED ISSUE #294: No clickable node in a modified file after incremental parsing
1. Parse a project.
2. Start the server, get an "info tree" on a node. It works fine.
3. Modify a file (by inserting an empty line to the beginning).
4. Do the incremental parsing.
5. Start the server, get an "info tree" on a node in the modified file. I doesn't work now. Whatever I click, there is no info tree or diagram on it. It looks that all nodes are missing from that file.

mcserep:
Fixing incremental parsing:
- deleting all `CppEntity` records for modified files, not only those with a *Definition* AST type.
- caching preexisting `CppEdge` records in `RelationCollector`, so they will not get reinserted, raising a database exception.

Improved the handling of database errors in `OdbTransaction`.

Fixes #294

### DIFF
--- a/util/include/util/odbtransaction.h
+++ b/util/include/util/odbtransaction.h
@@ -179,11 +179,17 @@ void persistAll(Cont& cont_, std::shared_ptr<odb::database> db_)
     }
     catch (const odb::database_exception& ex)
     {
-      LOG(debug)
-        << item->toString();
-      LOG(error)
-        << ex.what() << std::endl;
+      LOG(debug) << item->toString();
 
+#ifdef DATABASE_PGSQL
+      if (std::string(ex.what()).find("25P02") != std::string::npos)
+      {
+        LOG(error) << "The transaction was aborted due to previous error, omitting further changes!";
+        break;
+      }
+#endif
+
+      LOG(error) << ex.what() << std::endl;
       throw;
     }
   }
--- a/util/include/util/odbtransaction.h
+++ b/util/include/util/odbtransaction.h
@@ -175,12 +175,16 @@ void persistAll(Cont& cont_, std::shared_ptr<odb::database> db_)
         << item->toString();
       LOG(warning)
         << ex.what() << std::endl
-        << "AST nodes in this translation unit will be ignored!";
+        << "Further changes in this transaction will might be ignored!";
     }
     catch (const odb::database_exception& ex)
     {
-      // TODO: Error code should be checked and rethrow if it is not unique
-      // constraint error. Error code may be database specific.
+      LOG(debug)
+        << item->toString();
+      LOG(error)
+        << ex.what() << std::endl;
+
+      throw;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


util/include/util/odbtransaction.h
```diff
@@ -175,12 +175,22 @@ void persistAll(Cont& cont_, std::shared_ptr<odb::database> db_)
         << item->toString();
       LOG(warning)
         << ex.what() << std::endl
-        << "AST nodes in this translation unit will be ignored!";
+        << "Further changes in this transaction will might be ignored!";
```
bruntib: "will might be" -> "will be"

util/include/util/odbtransaction.h
```diff
@@ -175,12 +175,22 @@ void persistAll(Cont& cont_, std::shared_ptr<odb::database> db_)
...
     catch (const odb::database_exception& ex)
     {
-      // TODO: Error code should be checked and rethrow if it is not unique
-      // constraint error. Error code may be database specific.
+      LOG(debug) << item->toString();
+
+#ifdef DATABASE_PGSQL
+      if (std::string(ex.what()).find("25P02") != std::string::npos)
+      {
+        LOG(error) << "The transaction was aborted due to previous error, omitting further changes!";
```
bruntib: What does "previous error" mean?

util/include/util/odbtransaction.h
```diff
@@ -175,12 +175,22 @@ void persistAll(Cont& cont_, std::shared_ptr<odb::database> db_)
...
+        << "Further changes in this transaction will might be ignored!";
     }
     catch (const odb::database_exception& ex)
     {
-      // TODO: Error code should be checked and rethrow if it is not unique
-      // constraint error. Error code may be database specific.
+      LOG(debug) << item->toString();
+
+#ifdef DATABASE_PGSQL
+      if (std::string(ex.what()).find("25P02") != std::string::npos)
```
bruntib: It is fine, because it should occur rarely, but `std::strstr()` could also be used for finding substrings in a C style string.

util/include/util/odbtransaction.h
mcserep: I was not sure whether it is true for SQLite also, but will reword it then.

util/include/util/odbtransaction.h
mcserep: A previous exception. In case of PostgreSQL if a database exception is raised during a transaction, the transaction **must be rolled back**, otherwise all further commands will raise a specific exception with code **25P02**, stating that the transaction has been aborted.

util/include/util/odbtransaction.h
mcserep: Fixed in d0099ddf824925de4247256b084ca474631f6fc9.

util/include/util/odbtransaction.h
mcserep: Fixed in d0099ddf824925de4247256b084ca474631f6fc9.

util/include/util/odbtransaction.h
bruntib: I mean does the user get notified about that exception? This is an error level log, and the exception message is not printed earlier.

mcserep: Yes.

PR #301, opened 2018-12-07
Eliminating compiler warnings

bruntib:
In this commit the compiler warnings are eliminated, except for the ones
in the C++ service. Those are unhandled cases in switch structures. Those
are planned to be implemented.

### DIFF
--- a/parser/include/parser/parsercontext.h
+++ b/parser/include/parser/parsercontext.h
@@ -32,10 +32,10 @@ enum class IncrementalStatus
 struct ParserContext 
 {  
   ParserContext(
-      std::shared_ptr<odb::database> db_,
-      SourceManager& srcMgr_,
-      std::string& compassRoot_,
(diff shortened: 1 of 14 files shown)


parser/include/parser/parsercontext.h
```diff
@@ -32,10 +32,10 @@ enum class IncrementalStatus
...
 {  
   ParserContext(
-      std::shared_ptr<odb::database> db_,
-      SourceManager& srcMgr_,
-      std::string& compassRoot_,
-      po::variables_map& options);
+    std::shared_ptr<odb::database> db_,
+    SourceManager& srcMgr_,
+    std::string& compassRoot_,
+    po::variables_map& options);
```
mcserep: There is also a naming issue for argument `options` which could be fixed now, it should be `options_`, according to the coding convention.

PR #322, opened 2019-04-05
Do not use keepalive in tests

mcserep:
Tests should be performed without any error terminating the program and should fail when e.g. an unhandled exception is raised or on a segmentation fault.
Therefore the `keepalive` script should not be used, because it restarts the program and the tests will probably never fail, nor succeed.


mcserep:
> I would even recommend to remove the `keepalive` script completely. Earlier its real purpose was to set the runtime environment of CodeCompass, but it has been solved since then. If the parser crashes for some reason then we should fix it instead of using such a script and hide the real error.

I agree, we can address this in a separate PR.
My personal experience was that the webserver can crash sometimes. While these cases should also be examined and the errors fixed, it is also convenient to automatically restart the webserver in such cases.

bruntib:
There are more appropriate and standard solutions for keeping a process alive, like daemon or maybe Docker in the future.

Issue #326, opened 2019-04-23
CppParserTest fails over sqlite database


mcserep:
2 of 10 test cases fail in `CppParserTest` on the master branch:
1. `CppParserTest.FunctionCall`
2. `CppParserTest.Variable`

All tests are passed with a PostgreSQL database, this issue only occurs with an SQLite database.

Test log:
> 2: [  INFO ] /home/mate/Repos/Ericsson/CodeCompass/plugins/cpp/test/src/cpptest.cpp:24:: Using database for tests: sqlite:database=/home/mate/cc/test_db.sqlite
> 2: [==========] Running 10 tests from 1 test case.
> 2: [----------] Global test environment set-up.
> 2: [----------] 10 tests from CppParserTest
> 2: [ RUN      ] CppParserTest.FilesAreInDatabase
> 2: [       OK ] CppParserTest.FilesAreInDatabase (3 ms)
> 2: [ RUN      ] CppParserTest.SimpleFunction
> 2: [       OK ] CppParserTest.SimpleFunction (1 ms)
> 2: [ RUN      ] CppParserTest.FunctionDeclarationOnly
> 2: [       OK ] CppParserTest.FunctionDeclarationOnly (0 ms)
> 2: [ RUN      ] CppParserTest.FunctionWithMultipleDeclarations
> 2: [       OK ] CppParserTest.FunctionWithMultipleDeclarations (0 ms)
> 2: [ RUN      ] CppParserTest.FunctionCall
> 2: /home/mate/Repos/Ericsson/CodeCompass/plugins/cpp/test/src/cppparsertest.cpp:238: Failure
> 2: Value of: false
> 2:   Actual: false
> 2: Expected: true
> 2: /home/mate/Repos/Ericsson/CodeCompass/plugins/cpp/test/src/cppparsertest.cpp:238: Failure
> 2: Value of: false
> 2:   Actual: false
> 2: Expected: true
> 2: /home/mate/Repos/Ericsson/CodeCompass/plugins/cpp/test/src/cppparsertest.cpp:238: Failure
> 2: Value of: false
> 2:   Actual: false
> 2: Expected: true
> 2: [  FAILED  ] CppParserTest.FunctionCall (1 ms)
> 2: [ RUN      ] CppParserTest.Typedef
> 2: [       OK ] CppParserTest.Typedef (0 ms)
> 2: [ RUN      ] CppParserTest.Record
> 2: [       OK ] CppParserTest.Record (1 ms)
[...]

PR #327, opened 2019-04-23
Add Travis CI

mcserep:
Execute CodeCompass build and run unit & integration tests on Travis CI.

This pull request is ready to be merged, the CI job for SQLite only fails because the master branch itself contains a bug, see #326. It should be merged after that issue is resolved. Commits can be squashed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,22 @@ sudo: require
 
 language: cpp
 
+services:
+  - postgresql
+
 addons:
   apt:
     update: true
     sources:
       - ubuntu-toolchain-r-test
+      - llvm-toolchain-xenial-7
     packages:
       - libboost-all-dev
+      - llvm-7
+      - llvm-7-dev
+      - clang-7
+      - libclang-7-dev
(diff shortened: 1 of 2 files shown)


bruntib:
Is it worth compiling CodeCompass in Debug mode on Travis? We are not going to debug CodeCompass on the Travis server and using Release build may be faster in the CI.

mcserep:
Initially I set it to Debug mode, but you're right. Switched it to Release mode (e92e884be668d33dc77c37a35246638d0430c5e1), will see whether it got significantly faster.

mcserep:
Well, the CI runtime is 10-11 minutes in Release mode and 11-12 minutes in Debug mode, but I think this is acceptable. A minimal improvement could be building CodeCompass on 2 cores, because the environment has 2 cores according to the [docs](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system).

(Note: the runtime for the first job was 23 minutes, because Thrift had to be build, but it was cached after that.)

Issue #328, opened 2019-04-23
Simplify installation guide


mcserep:
During testing with dockerization and continuous integration, I realized that the [compilation and installation guide](https://github.com/Ericsson/CodeCompass/blob/master/doc/deps.md) could be simplified in 2 major points:
- [x] **Manual compilation of ODB is not required.** The default package repository both for Ubuntu Xenial and Bionic contains version 2.4 and CodeCompass compiles and works fine both with `g++-5` and `g++-7`. (The default GCC versions respectively.) Solved by #330.
- [x] **Manual compilation of LLVM/Clang is not required.** For Ubuntu Bionic LLVM 7 and Clang 7 is part of the default package repository, compiled with RTTI. For Xenial, the [official package repository provided by LLVM](https://apt.llvm.org/) can easily be added and then LLVM/Clang version 7 can be installed with `apt-get` (also compiled with RTTI). Solved by #333.

As a proof, recently completed Travis CI configuration also depends on Xenial without manual compilation of these dependencies, see #327.

I believe that the default installation guide for CodeCompass should be kept as simple as possible, so it won't deter newcomer developers and possible users with multiple hours of compilation time for just the dependencies. Manual compilation of these dependencies could be an *advanced guide*, probably a wiki a page.

bruntib:
I completely agree with the simplification of the documentation. I wanted to remove the ODB related parts from the docs for a while, because I saw this dependency in the Ubuntu repository earlier.

I'll check LLVM too, because as far as I remember, it didn't work because of that RTTI thing. However, we had personal discussions about this issue. I don't know yet how RTTI is related to the linkage of a project where it caused the problems. It would be great if LLVM could also be installed from a repo.

mcserep:
LLVM/Clang is compiled with RTTI support in the standard Ubuntu repository, so it shall cause no issue. Maybe it was not the case in the past, but right now it is 😃

At least it worked for me (and in the CI #327) for parsing simple projects, but test it for yourself of course.

PR #330, opened 2019-05-01
ODB known issues in the documentation.

bruntib:
ODB can be installed from Ubuntu package repository.

### DIFF
--- a/doc/deps.md
+++ b/doc/deps.md
@@ -18,10 +18,11 @@ packages are necessary for building CodeCompass:
   parsing the source code. Version 7.0 or newer is required. Clang 7 is not yet
   released, so the project must be compiled manually from source.
   ***See [Known issues](#known-issues)!***
-- **`odb`**, **`libodb-dev`**, **`libodb-sqlite-dev`**, **`libodb-pgsql-dev`**:
-  For persistence ODB can be used which is an Object Relation Mapping (ORM)
-  system.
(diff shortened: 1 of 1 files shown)


mcserep:
So I guess you agree with my observations in #328?
Write a comment there if you think, because I think LLVM/Clang dependency can also be simplified.

bruntib:
I completely agree with the simplification of the documentation. I wanted to remove the ODB related parts from the docs for a while, because I saw this dependency in the Ubuntu repository earlier.

I'll check LLVM too, because as far as I remember, it didn't work because of that RTTI thing. However, we had personal discussions about this issue. I don't know yet how RTTI is related to the linkage of a project where it caused the problems. It would be great if LLVM could also be installed from a repo.

mcserep:
Good, I will check whether the **`libodb-dev`**, **`libodb-sqlite-dev`** and **`libodb-pgsql-dev`** packages are truly not required, then I will merge this pull request.

LLVM/Clang is compiled with RTTI support in the standard Ubuntu repository, so it shall cause no issue. Maybe it was not the case in the past, but right now it is 😃

bruntib:
These packages that you listed are the dependencies of the package `odb`, so they should be installed automatically.

mcserep:
True.

mcserep:
Partially superseded #270

PR #332, opened 2019-06-08
Correct connection string in doc; gitignore build dir

rpoisel:
Following your build documentation was quite straight forward (thanks!). There was just one minor issue, I wanted to correct in this PR.


mcserep:
Thanks @rpoisel , I merged it.
(Please note however, that sqlite support will might be dropped in the future due to performance issues with larger databases.)

PR #334, opened 2019-06-11
Improvement on incremental parsing of C++ files

mcserep:
These changes significantly improves the performance of incremental parsing of C++ files on 2 major points:
- The database cleanup is performed on a per-file basis instead of a single, enormous transaction.
- Cleanup is performed in a parallel manner. For this to be achieved, a graph is constructed from the files to be cleaned up along the inclusion relations. Then the nodes of the graph are sorted topologically. In each iteration the leaf nodes of the graph are cleaned up (paralelly), which guarantees that no file is cleaned up before its dependants.

A new option for the `CodeCompass_parser` binary named `--incremental-threshold` has also been introduced, which sets an upper threshold of change (in the percentage of changed files) for incremental parsing. Above the threshold, a full, forced reparse is performed. The default value is *10%*.


mcserep:
This pull request now also fixes important bug in `RelationCollector` when using multiple parser jobs (see 2b75217aa82b1940edd5854959cf71793ce98812).

`RelationCollector` has a `CppEdge` cache defined as a static member, which is populated by the constructor of the first `RelationCollector` instance. Assuming multiple parser jobs, the first thread will start populating the cache, while the other threads will see that the cache is not empty (however also not completed) and continue to work with an incomplete cache. Thus already existing `CppEdge` instances will be reinserted into the DB, resulting in an exception finally

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -339,7 +343,86 @@ std::vector<std::string> CppParser::getDependentParsers() const
   return std::vector<std::string>{};
 }
 
-bool CppParser::preparse(bool dry_)
+std::vector<std::vector<std::string>> CppParser::createCleanupOrder()
```
bruntib: What is the use of this ordering? Isn't it enough simply removing all records in all tables belonging to the modified/deleted files?

parser/include/parser/sourcemanager.h
```diff
@@ -32,16 +32,12 @@ class SourceManager
     bool operator()(model::FilePtr) const { return true; }
   };
 
-  template<typename Filter = SourceManager::AllFilesFilter>
-  std::vector<model::FilePtr> getFiles(const Filter& beta_ = Filter())
+  /**
+   * This function is a getter for the size of _files.
+   */
+  std::map<std::string, model::FilePtr>::size_type filesSize()
```
bruntib: I think `numberOfFiles()` or something similar would be better. This would also result a better documentation to the purpose of the function. And we should also modify the documentation above, because the user of this function doesn't know what `_files` is.

parser/include/parser/sourcemanager.h
```diff
@@ -50,9 +46,19 @@ class SourceManager
...
    * nullptr.
+   * @param path_ The file path to look up.
    */
   model::FilePtr getFile(const std::string& path_);
 
+  /**
+   * This function returns a pointer to the corresponding model::File objects
+   * based on the given beta_ filter. The objects are read from a cache.
+   * Uncached files are ignored.
+   * @param beta_ A filter functor iterated over the cached model::File objects.
```
bruntib: What are cached files? The `_files` container is filled up when starting the parser, so it always contains all files. Please check this statement, but I think it works like that.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -339,7 +343,86 @@ std::vector<std::string> CppParser::getDependentParsers() const
...
+    }
+
+    ++index;
+  }
+  LOG(debug) << "[cppparser] Topology has " << index << " levels.";
+
+  return order;
+}
+
+void CppParser::markModifiedFiles()
```
bruntib: Why only included files are marked? What about the includers? Shouldn't those be reparsed too?

parser/src/parser.cpp
```diff
@@ -330,36 +333,61 @@ int main(int argc, char* argv[])
...
   }
 
   if (vm.count("dry-run"))
   {
     incrementalList(ctx);
+    return 0;
   }
-  else
+
+  if(ctx.fileStatus.size() > ctx.srcMgr.filesSize() * vm["incremental-threshold"].as<int>() / 100.0)
```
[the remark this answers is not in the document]
mcserep: Deleted files are still required to be cleaned from the database, which is a significant part of the work to be done. Therefore we decided to count also the deleted files against this threshold.

parser/src/parser.cpp
```diff
@@ -176,9 +179,9 @@ void incrementalList(cc::parser::ParserContext& ctx_)
  */
 void incrementalCleanup(cc::parser::ParserContext& ctx_)
 {
-  cc::util::OdbTransaction {ctx_.db} ([&]
+  for (auto& item : ctx_.fileStatus)
```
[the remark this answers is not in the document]
mcserep: While for maintaining the C++ parser database, the rollback log became enormous when handling the whole operation in a single transaction; one may argue this will not be the issue when only the `File`, `FileContent`, `BuildAction`, `BuildLog`, `BuildSource` and `BuildTarget` tables are considered. Yet it seems a more scalable solution.
Shall we do a performance test on it?

webserver/src/webserver.cpp
```diff
@@ -35,7 +35,7 @@ po::options_description commandLineArguments()
       "Logging level of the parser. Possible values are: debug, info, warning, "
       "error, critical")
     ("jobs,j", po::value<int>()->default_value(4),
-      "Number of worker threads.");
+      "Number of parseWorker threads.");
```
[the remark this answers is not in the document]
mcserep: I renamed the `CppParser::worker` method to `CppParser::parseWorker`. Looks like the refactoring feature of CLion was super greedy :smile:

plugins/cpp/parser/src/cppparser.cpp
mcserep: As described in this PR's description:

> Cleanup is performed in a parallel manner. For this to be achieved, a graph is constructed from the files to be cleaned up along the inclusion relations. Then the nodes of the graph are sorted topologically. In each iteration the leaf nodes of the graph are cleaned up (paralelly), which guarantees that no file is cleaned up before its dependants.

So, if e.g. `a.cpp` depends on `b.h` and we clean up `b.h` before `a.cpp`, that would fail the cleanup process.

parser/include/parser/sourcemanager.h
mcserep: > The _files container is filled up when starting the parser, so it always contains all files.
Exactly, that is why we named it a cache. Since it contains all files, which were in the DB when the `SourceManager` was instantiated or were later added through the `SourceManager`.
It does not contain files which were added to the DB directly or thorough another process. That is an in-memory cache of the File table in my understanding.

I think we mean the same thing here. Please suggest a better naming if you think *caching* is confusing here in the documentation.

plugins/cpp/parser/src/cppparser.cpp
mcserep: The `markByInclusion` method marks the **includer** files (recursively).
Please clarify if I misunderstood your concern here.

parser/src/parser.cpp
```diff
@@ -330,36 +333,61 @@ int main(int argc, char* argv[])
 
   //--- Start parsers ---//
 
+  /*
+   * Workflow for incremental parsing:
+   * 1. directly modified files are detected by ParserContext.
+   * 2. all plugin parsers marks the indirectly modified files.
+   * 3. all plugin parsers performs a cleanup operation.
+   * 4. global tables are cleaned up by parser.cpp.
+   * 5. all plugin parsers performs a parsing operation.
```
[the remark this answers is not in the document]
mcserep: Fixed in 80d9d2f2c410aad4feabf80cf21f1403c77f669f

parser/src/parser.cpp
```diff
@@ -330,36 +333,61 @@ int main(int argc, char* argv[])
...
   {
     incrementalList(ctx);
+    return 0;
   }
-  else
+
+  if(ctx.fileStatus.size() > ctx.srcMgr.filesSize() * vm["incremental-threshold"].as<int>() / 100.0)
   {
-    incrementalCleanup(ctx);
+    LOG(info) << "The number of changed files exceeds the given incremental"
```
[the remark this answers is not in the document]
mcserep: Fixed in 80d9d2f2c410aad4feabf80cf21f1403c77f669f

webserver/src/webserver.cpp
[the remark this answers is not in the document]
mcserep: Fixed in 80d9d2f2c410aad4feabf80cf21f1403c77f669f

parser/include/parser/sourcemanager.h
mcserep: Fixed in 80d9d2f2c410aad4feabf80cf21f1403c77f669f

mcserep:
> One more question: why do we distinguish modified and deleted files in `ParserContext`? Isn't a simple set enough for storing them all?

It is for:
- supporting a separate listing of modified/deleted files on a dry-run.
- possible future extension point if this metadata would be ever required.

plugins/cpp/parser/src/cppparser.cpp
intjftw: To reproduce the error that made topological ordering necessary, let's say we have these files:

definitions.h:
```
#ifndef DEFINITIONS_H
#define DEFINITIONS_H

enum class Day
{
  Monday = 1,
  Tuesday,
  Wednesday,
  Thursday,
  Friday,
  Saturday,
  Sunday  
};

enum Permission
{
  Readable = 4,
  Writable = 2,
  Executable = 1
};

#endif // DEFINITIONS_H
```

main.cpp:
```
#include <iostream>

#include "definitions.h"

int main()
{
  Day d = Day::Friday;
  std::cout << "d = " << static_cast<int>(d) << std::endl;
  d = Day::Sunday;
  std::cout << "d = " << static_cast<int>(d) << std::endl;
  
  int p = Readable;
  std::cout << "p = " << p << std::endl;
  p |= Executable;  
  std::cout << "p = " << p << std::endl;
  
  return 0;
}
```

Any modification in _definitions.h_ produces the following error when the arbitrarily ordered cleanup is executed:

> [INFO] (1/1) Parsing /home/efekane/Repos/cc-tests/enums/main.cpp
[DEBUG] CppMacro
id = 0
mangledNameHash = 13147581534377510727
qualifiedName = DEFINITIONS_H
[ERROR] 23505: ERROR:  duplicate key value violates unique constraint "CppEntity_astNodeId_i"
DETAIL:  Key ("astNodeId")=(3956703780563145061) already exists.

bruntib: I still couldn't reproduce this error. I created a branch "no_order" on my fork where I appended a commit in which the topological ordering is skipped. Could you please checkout and run that code? If you can reproduce the error, could you please describe a scenario? Thanks!

mcserep: This error cannot be reproduced and might be a result of the usage of an outdated codebase or workspace database. We can drop this issue.
However we still have issues related to the parallel cleanup of files with include dependency (which can be prevented with the topological sorting), I emailed you the details @bruntib .

PR #337, opened 2019-06-18
Fixes download URLs for ncurses and doxygen.

pbomta:


mcserep:
Thanks for the fixes @pbomta , since version Earhart is not actively developed, it seems these URLs got outdated.

PR #340, opened 2019-07-23
Update Thrift

mcserep:
Update Thrift from version 0.10.0 to 0.12.0.  
CodeCompass is no longer binded to Java 1.8, newer versions can be used. As a default, `default-jdk` is set as a new requirement. (For Xenial it is 1.8, for Bionic it is 1.11)


mcserep:
Updated with new CI script from the master branch to fix merge conflict

whisperity:
Could you please rebase the patch on master, or show which PRs are blocking the merge of this, instead of having a `merge master into topic` kinda commit in the history?

mcserep:
> Could you please rebase the patch on master, or show which PRs are blocking the merge of this, instead of having a `merge master into topic` kinda commit in the history?

Since #342 modified the CI script significantly, a merge conflict had to be solved in `.travis.yml`.

mcserep:
Updated commit message as requested.

bruntib:
Note, that the GUI still uses an earlier version of Thrift:
https://github.com/Ericsson/CodeCompass/blob/f8e89ac3aea27146c502d34003821d8b03445d23/webgui/package.json#L12
We should upgrade it too. Hopefully these different versions at server and client side are compatible. But probably they are, because there was no API change.

mcserep:
> We should upgrade it too. 

I agree. However the [NPM Thrift API](https://www.npmjs.com/package/thrift) depends on NodeJS and uses the `require` directive from version `0.10.0`, thus the update is not trivial. Shall we use RequireJS or what other suggestions you have, @bruntib , @zomen2 ?

> Hopefully these different versions at server and client side are compatible. But probably they are, because there was no API change.

After a quick test everything seems to work on the web GUI, so it is not super important, but would be nice to upgrade.

PR #343, opened 2019-07-29
Restrict LLVM's scope to the C++ parser module

mcserep:
LLVM is only used by the C++ parser module, which should be reflected by the build system, by not importing it in the global `CMakeLists.txt`.
Based on the original work of @whisperity.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,11 +10,11 @@ include(Functions.cmake)
 # Do some sanity check on the testing setup and enable testing if applicable.
 include(Testing.cmake)
 
-find_package(Threads REQUIRED)
 find_package(Boost   REQUIRED COMPONENTS filesystem log program_options regex system thread)
-find_package(Odb     REQUIRED)
-find_package(Thrift  REQUIRED)
 find_package(Java    REQUIRED)
+find_package(Odb     REQUIRED)
+find_package(Threads REQUIRED)
+find_package(Thrift  REQUIRED)
 find_package(GTest)
 
 include(UseJava)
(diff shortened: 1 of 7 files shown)


whisperity:
Supersedes #270.

mcserep:
CI fails on tests, requires further investigation.

mcserep:
Fixed, CI passes now.

CMakeLists.txt
```diff
@@ -11,44 +11,31 @@ include(Functions.cmake)
...
-  message(SEND_ERROR "RTTI is required for LLVM")
-endif()
-
 include(UseJava)
 
 # Set the third-party libraries' path properly so the installation can find them.
 include(Exports.cmake)
 set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH};${RUNENV_LD_LIBRARY_PATH}")
 
+# Modules
```
whisperity: Is there any particular reason as to why the directories are in this order? If not, can we try having them in alphabetical ordering for ease of read?

CMakeLists.txt
```diff
@@ -11,44 +11,31 @@ include(Functions.cmake)
...
 add_subdirectory(util)
 add_subdirectory(parser)
-add_subdirectory(plugins)
 add_subdirectory(webserver)
 add_subdirectory(service)
 add_subdirectory(logger)
 add_subdirectory(scripts)
 add_subdirectory(webgui)
+# Plugins
+add_subdirectory(plugins)
```
whisperity: I understand it could be better if this one is on the bottom.

CMakeLists.txt
mcserep: I only put the plugins directory to the end, because I think plugins should be added after all the core modules for a better CMake output and easier debugging in same cases. The order of these modules I guess is arbitrary.

mcserep: I reordered them, see https://github.com/Ericsson/CodeCompass/pull/343/commits/88acad41d4d1a23f6a7063fd1094ace85198e45f

PR #368, opened 2019-12-13
Thrift13

filbeofITK:
Upgraded thrift's version to the newest (13).
Resolved conflict between thrift and dojo (complications because of dojo's require function).
Changed the travis file to test with thrift's 13 version instead of 12.


scripts/remover.sh
```diff
@@ -0,0 +1,11 @@
+#/bin/bash
+
+FILES=$(ls *.js)
+
+for FILE in $FILES
+do
+	if [ -f $FILE ]
+	then
+		sed -i -e "/if (typeof Int64 === 'undefined' && typeof require === 'function') {/,+2 d" $FILE
```
mcserep: Why is this post-install hacking of Thrift Javascript files required?
- What issue does it solve?
- Is it impossible (or unreasonable) to solve it with another approach? 

Please explain, because modifying generated files or vendor code with a Bash script after installation should be a very last resort as a solution, as it is platform-specific and hard to maintain in the future.

filbeofITK: It is required because the thrift generated files have a condition to determine whether they can use the require function (which is node-js specific) or not. Since CodeCompass's web client uses client version of javascript we can not use it. Unfortunately there is a require function in the dojo software kit , by which an error is generated, causing the web client to be unfunctionable.

I could not find any other approach to this problem. I've looked it up but even though it is specified to the thrift api to generate client side js it generates this condition.

Ther remover script only contains a few lines and is only called from one cmake function, so it is easy to remove it if the problem gets resolved.

CodeCompass contains loads of platform specific scripts, so I believe it will not cause any problem if we add one more.

mcserep: I see the problem now. While it is true, that this script contains only few lines and is only called from one CMake function, so it is easy to remove; but each time we add such a "workaround" solution, we are degrading the software quality and maintainability a little bit. So I cannot agree with the "it will not cause any problem if we add one more" part *in general*.

On the other hand there might be no better solution for this issue you described, so it might be required. (I cannot think of anything better right now, but I will think about it later.)  
 Jumping back a little: why are we upgrading to Thrift 0.13 in the first place? Was there an issue or bug with version 0.12 which we need to solve? Or just to keep the dependencies nice and fresh?

filbeofITK: Upgrading to a newer version solved one minor bug in the web client, but yes the main goal was to keep the dependencies in check. Sadly yes there were several bugs with the 0.12 version. So many, that to upgrade to it would require an additional building script. (I'm talking about the client side js library) Also thrift 0. 12 requires version 8 java, and the default-jdk repository that we cite as dependency contains java 11. So we would eventually have to upgrade either the deps.md or thrift :)

Functions.cmake
```diff
@@ -70,6 +70,8 @@ function(install_js_thrift)
     DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/gen-js/
     DESTINATION ${INSTALL_GEN_DIR}
     FILES_MATCHING PATTERN "*.js")
+  install(
+    CODE "execute_process(COMMAND ${CMAKE_SOURCE_DIR}/scripts/remover.sh WORKING_DIRECTORY ${INSTALL_GEN_DIR})")
```
mcserep:
```suggestion
    CODE "execute_process(COMMAND /bin/bash ${CMAKE_SOURCE_DIR}/scripts/remover.sh WORKING_DIRECTORY ${INSTALL_GEN_DIR})")
```
When omitting `/bin/bash`, the *remover.sh* was not executed for me, even after correcting the shebang (see below).

scripts/remover.sh
```diff
@@ -0,0 +1,11 @@
+#/bin/bash
```
mcserep:
```suggestion
#!/bin/bash
```
Incorrect shebang.

scripts/remover.sh
mcserep: Okay, I have no better idea, so let's go with this. I have tested your branch, see the new review comments for issues to be fixed.

A remark: while Thrift 0.12 may required Java 8, we concluded with @bruntib  and @whisperity  that CodeCompass never really required the Java library to be built for Thrift, **only the generator part was required**. This means, that you can compile Thrift with the `--with-java=NO` option with a modern Java version and CodeCompass will still compile and work properly. (I think I will even add this `--with-java=NO` flag to `deps.md` before this PR is merged as a fix for the Thrift 0.12 issue with newer JDK than version 8.)

mcserep: Okay, I have no better idea, so let's go with this. I have tested your branch, see the new review comments for issues to be fixed.

A remark: while Thrift 0.12 may required Java 8, we concluded with @bruntib  and @whisperity  that CodeCompass never really required the Java library to be built for Thrift, **only the generator part was required**. This means, that you can compile Thrift with the `--with-java=NO` option with a modern Java version and CodeCompass will still compile and work properly. (I think I will even add this `--with-java=NO` flag to `deps.md` before this PR is merged as a fix for the Thrift 0.12 issue with newer JDK than version 8.)

Functions.cmake
filbeofITK: Uh I really missed typing again... I fixed the shebang.
The install work just fine on my PC without the `/bin/bash` , I don't know what makes the difference.

filbeofITK: Maybe it can be because you use a different shell from bash?
If the problem is caused by that I'll add the `/bin/bash` to the script.

mcserep: The default shell for my user is `/bin/bash`, so it should not have been an issue.
I will test it on another computer next week.

mcserep: Well, it worked on another machine now (with Ubuntu Bionic). I will merge this now and if any issue is raised, I will add `/bin/bash` later.

PR #370, opened 2020-01-16
Set the linker to gold, because it is significantly faster.

filbeofITK:
Also the CodeCompass project does not require any of the gnu linker's
extra tools, and gold is located in the binutils package, that needs to be installed
to build CodeCompass, meaning it's safe to change to gold.

### DIFF
--- a/Config.cmake
+++ b/Config.cmake
@@ -68,10 +68,14 @@ set(ODBFLAGS
   --include-with-brackets
   --default-pointer "std::shared_ptr")
 
+# Set CXX flags
 set(CMAKE_CXX_FLAGS "-W -Wall -Wextra -pedantic\
   -std=c++14 \
   -DDATABASE_${DATABASE_U} \
   -DBOOST_LOG_DYN_LINK")
+
+# Set gold linker
+set(CMAKE_LINKER "/usr/bin/gold")
 
 # Cmake module directory (FindOdb, FindThrift etc.)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}")
--- a/Config.cmake
+++ b/Config.cmake
@@ -74,8 +74,14 @@ set(CMAKE_CXX_FLAGS "-W -Wall -Wextra -pedantic\
   -DDATABASE_${DATABASE_U} \
   -DBOOST_LOG_DYN_LINK")
 
-# Set gold linker
-set(CMAKE_LINKER "/usr/bin/gold")
+# Gold is the primary linker 
+if(NOT DEFINED CODECOMPASS_LINKER)
+    set(CODECOMPASS_LINKER "/usr/bin/gold")
+endif()
+
+set(CMAKE_LINKER "${CODECOMPASS_LINKER}")
+
+
 
 # Cmake module directory (FindOdb, FindThrift etc.)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}")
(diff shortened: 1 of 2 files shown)


whisperity:
If a package needs to be installed, please document it in `docs/deps.md`.

I'm conflicted about this change though. It would be better if a sensible command-line option (like LLVM's `LLVM_USE_LINKER`) is introduced.

What if someone wants to use LLD? Many say that is even smarter and faster and can also do link-time optimisation.

People are free to choose their own linker. For example, I never had a problem building CodeCompass with GNU ld (LLVM on the other hand... 😜). This change most likely breaks the project under OS/X-macOS and Windows as the paths could be wholly different here.

filbeofITK:
Yes it would be better to make it a choice, and the default value would be the gold linker, I'm working on it .
Gold is located in the same package as the gnu compiler, which is a dependency of several other packages CodeCompass requires, so there is no need to modify the `docs/deps.md`
I didn't really know that we were aiming to port CodeCompass to windows or macOS either.

mcserep:
Actually `gold` linker is officially still in [beta test](https://www.gnu.org/software/binutils/). How much difference does it make on the compilation of CodeCompass to switch to the gold linker? I mean, does it really worth it to make it the *default* linker?

filbeofITK:
Well on a low spec machine (my laptop) it reduced the time of building to it's quarter, since it uses RAM more efficiently. On a more powerful computer, with more RAM, the install took less than, the half of the time it took with the standard gnu linker.

mcserep:
Well, since this is in the official binutils package, I am not against this change, but just for the record, I could not measure a significant difference.
Computer specification: Ubuntu 16.04 virtual machine, 4 CPU core, 8GB RAM. Compilation on 2 threads. 

Clean compilation time with `ld` linker: 5 minutes 37 sec
 Clean compilation time with `gold` linker: 5 minutes 18 sec

I measured it multiple times, the results were the same. So I assume most time is not spent on linking. But we can switch to gold, if you have better performance with that.

@whisperity, since you have commented on it before, please check whether the modifications of this PR are fine for you, then we can merge.

Config.cmake
```diff
@@ -68,11 +68,21 @@ set(ODBFLAGS
...
 
+# Set CXX flags
 set(CMAKE_CXX_FLAGS "-W -Wall -Wextra -pedantic\
   -std=c++14 \
   -DDATABASE_${DATABASE_U} \
   -DBOOST_LOG_DYN_LINK")
 
+# Gold is the primary linker 
+if(NOT DEFINED CODECOMPASS_LINKER)
+    set(CODECOMPASS_LINKER "/usr/bin/gold")
```
bruntib: I think we shouldn't hard-code this path. We should introduce some CMake magic here to find `gold` binary. Probably there are specific CMake commands for this. Moreover in case this linker is not installed for some reason on a platform we should leave `$CMAKE_LINKER` on its default value.

filbeofITK: Sadly I didn't find any of that kind of magic. Cmake only has compiler auto detection, and then the compiler driver picks the linker if we don't specify it. There is one way where, we make an extrenal tool chain (like with cross compiling) and implement the linker selection in that. Then we pass this tool chain to the compiler driver. I don't believe this should be worth it.

whisperity: Here's the relevant magic from LLVM tree: https://reviews.llvm.org/D28738

mcserep: How is this is any magic? As far as I can see, it simply uses the `lld` as the linker's binary path, which has to be found on the `PATH` I guess. There is no work done on locating this binary.

mcserep:
```suggestion
    set(CODECOMPASS_LINKER "gold")
```

So this would be the "magic", depending on the OS to locate the binary?

filbeofITK: Yes. Yes it was :(

Issue #371, opened 2020-01-19
Skipped parsers are not skipped after all


intjftw:
The `getPluginNames()` method of `pluginhandler.cpp` does not consider skipped parsers but returns all of them. This causes runtime error when a parser is skipped by the -s switch.

PR #373, opened 2020-01-20
Skipped plugins are not started, only loaded ones.

intjftw:
Skipped plugins have been called in `parser.cpp` along with non-skipped ones when the `pluginHandler` started the parsers which caused a runtime error. Now only loaded parsers are being started. Solves #371 .

### DIFF
--- a/parser/include/parser/pluginhandler.h
+++ b/parser/include/parser/pluginhandler.h
@@ -37,6 +37,12 @@ public:
   void loadPlugins(std::vector<std::string>& skipParserList_);
 
   /**
+   * Return the list of loaded plugins based the file names in the dynamic
+   * libraries.
+   */
+  std::vector<std::string> getLoadedPluginNames() const;
+
+  /**
    * Return the list of available plugins based on the file names in the parser
    * lib directory.
    */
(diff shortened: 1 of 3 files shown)


parser/src/pluginhandler.cpp
```diff
@@ -63,6 +63,18 @@ void PluginHandler::loadPlugins(std::vector<std::string>& skipParserList_)
   }
 }
 
+std::vector<std::string> PluginHandler::getLoadedPluginNames() const
+{
+  std::vector<std::string> plugins;
+
+  for (auto pair : _dynamicLibraries)
```
mcserep:
```suggestion
  for (const auto& pair : _dynamicLibraries)
```

parser/include/parser/pluginhandler.h
```diff
@@ -36,6 +36,12 @@ class PluginHandler
    */
   void loadPlugins(std::vector<std::string>& skipParserList_);
 
+  /**
+   * Return the list of loaded plugins based the file names in the dynamic
+   * libraries.
```
mcserep: While the data member behind this functionality is named `_dynamicLibraries`, it is a private data member, so this method description can be confusing for someone just reading the public API.  
I think it should simply state that this methods returns the loaded and not skipped plugin names.

Also, an *"on"* preposition is missing after the word *"based"*.

mcserep:
*(CI only failed because of too early merge.)*

PR #377, opened 2020-02-20
Changed docker RUN wget to ADD.

filbeofITK:
This a more docker like approach.
Also in the thrift build I changed the `make -j 4` to `make -j $(nproc)` so it suits every computer.


mcserep:
Maybe we can even drop the `wget` installation then?

PR #383, opened 2020-04-03
Docker web

filbeofITK:
This is a rework of @csordasmarton 's pull request.


mcserep:
I will close this PR, as it highly overlaps with the just merged #354 and the review was done there, while not yet on this one.

I have added a couple of modifications to the merged PR from this one, but feel free to open a new PR and submit any further suggestions you think shall get merged.

On the other hand I don't think that the production image shall use the development image as a parent, because that way everything which was required to build CodeCompass (and possibly their intermediate files) will be present in the production image. One of the main goals of the production docker image was to keep it small and it shall only contain the runtime dependencies of CodeCompass.

PR #385, opened 2020-04-19
Switched to odb 2.5 from 2.4, since the version 2.4 is currently broken.

filbeofITK:
Odb 2.5 is in beta currently but, it is going live soon and it is said to be as stable as 2.4 was, before the gcc upgrades broke it. 
I included some needed headers to the ` plugins/cpp/model/include/model/cppentity.h` and the ` model/include/model/buildaction.h` which were missing, but previously the odb headers included them, so this error stayed silent. 
I wrote an install script too, and made the docker image to use it.


mcserep:
Please amend the CI script for Bionic accordingly which would verify that with these modifications CodeCompass can be built.

doc/deps.md
```diff
@@ -91,6 +93,19 @@ by other processes which could, in extreme cases, make the system very hard or
...
+Odb is an Object relational Mapping tool, that is required by CodeCompass.
+We are using odb 2.5 which is not part of the official ubuntu repositories, so
+it needs to be built manually. For this, we provide a convenience script, under
+the scripts directory.
+
+```bash
+cd CodeCompass/scripts
+chmod u+x install_odb_2.5.sh
+./install_odb_2.5.sh
+```
```
mcserep, line 106: I don't think that the install script should be "hidden" in the `install_odb_2.5.sh` file, as neither the Thrift, Google Test or previously LLVM installation was approached in such a way. (Long ago we had an *install dependencies* script file, it similar to that.)
Or to be more precise, it's confusing that there is an install script for ODB, but not for other dependencies (Thrift, Google Test). It should be unified.

The usage of the script is also undocumented, like configuring the install location.

filbeofITK: I'm working on the CI script but since I don't really know the ins and outs of the topic it may take some time.

And I will expand the documentation of the install_odb script, then too.
The reason why I wrote an install script rather than, a long and detailed description, on how to build odb 2.5 is that it has several steps, that can take a long time on a low end machine, and this would make then build process rather frustrating. And the odb build uses its own build system too `build2`, unlike thrift and googletest. Those can be installed via cmake, in a more conventional way.

scripts/install_odb_2.5.sh
```diff
@@ -0,0 +1,71 @@
...
+
+#Building the odb compiler
+mkdir /tmp/odb
+pushd "/tmp/odb"
+bpkg create --quiet --jobs $(nproc) cc \
+  config.cxx=g++                       \
+  config.cc.coptions=-O3               \
+  config.bin.rpath=/usr/local/lib      \
+  config.install.root=/usr/local       \
+  config.install.root=/usr/local
```
whisperity, line 62: This line seems to be duplicated..

scripts/install_odb_2.5.sh
```diff
@@ -0,0 +1,71 @@
...
+
+bpkg add https://pkg.cppget.org/1/beta --trust-yes
+bpkg fetch --trust-yes
+bpkg build odb --yes
+bpkg build libodb --yes
+bpkg build libodb-sqlite --yes
+bpkg build libodb-pgsql --yes
```
whisperity, line 69: Is the order of these commands okay? Usually ODB was built with `libodb` first, then `libodb-<database>` second (for which build2 told me to add an extra `config.import.libodb` line!), then the ODB binary itself.

I think this script needs to be tested in cleaned environments.

doc/deps.md
whisperity: Agree, magic scripts is somewhat of a bad idea. It's horrible already that you cannot trust the upstream packages to not contain conflicts on requirements...

scripts/install_odb_2.5.sh
whisperity: The commands I executed on my machine was:

`bpkg create [... your arguments in the script ...] config.import.libodb=libodb-2.5.0-b.17`
(the directory name mined from the downloaded thingy)
then

 * `bpkg build libodb`
 * `bpkg build libodb-sqlite` (this broke first and required me to add the import line)
 * and finally `bpkg build odb`

then the `install` one.

whisperity:
Also, we need to consider people who do not with to install *anything* non-package-manager driven to system-wide paths... the current script uses `/usr/local` which I'd rather not pollute with development stuff (it already contains plenty of vendor-specific stuff) that is not easily possible to uninstall.

whisperity:
> Also, we need to consider people who do not with to install _anything_ non-package-manager driven to system-wide paths... the current script uses `/usr/local` which I'd rather not pollute with development stuff (it already contains plenty of vendor-specific stuff) that is not easily possible to uninstall.

For this to fix, I needed to add `${ODB_INCLUDE_DIRS}` to the `target_include_directories` here:

https://github.com/Ericsson/CodeCompass/blob/22be6a609531ec40a92e6498f2843caa4b3d1c4c/Functions.cmake#L43-L53

scripts/install_odb_2.5.sh
filbeofITK: Well I believe the order is okay, but I'm not 100% sure. At the bpkg's web page the tutorial followed this order, that's why I did too. 
About the testing, I completely agree, the script is currently only tested on the ubuntu :bionic docker image.

filbeofITK:
I agree that the `/usr/local` isn't the best permanent solution, I'm going to add the destination as a mandatory option to the script.

.travis.yml
```diff
@@ -22,8 +22,9 @@ services:
     # Java
     - default-jdk
     # ODB
-    - odb
-    - libodb-dev
+    - gcc-7
+    - gcc-7-plugin-dev
```
mcserep, line 26: A useful note for Travis @filbeofITK , that the CI starts with a pre-built image, a similar concept like Docker images.
The selected image is defined by the `os` (in our case *linux*), the `distro` (in our case *xenial* or *bionic*) and the `language` configuration parameter (in our case *cpp* for the CI). 

These images have a couple of software preinstalled, as described here:  
https://docs.travis-ci.com/user/languages/cpp/

It says that *Xenial ships with GCC 5.4.0* and *Bionic ships with GCC 7.4.0*. These are typically installed in `/usr/local/`, but of course you can still install any APT package (to `usr/`), if necessary.

The `minimal` image is also useful to check, because the `cpp` image is a superset of that, so e.g. `curl` is part of all images, there is no need to install it explicitly:  
https://docs.travis-ci.com/user/languages/minimal-and-generic/

Minimizing additional dependency installation can boost up the CI job runtime.

filbeofITK: Thanks :)

scripts/install_odb_2.5.sh
bruntib: I have this error message on `fetch` command:
```
error: unable to fetch https://pkg.cppget.org/1/beta/repositories.manifest
```

csordasmarton: Is it possible that you have some network issues? `curl` can download this file successfully: `curl https://pkg.cppget.org/1/beta/repositories.manifest`.

docker/dev/Dockerfile
```diff
@@ -44,14 +45,21 @@ RUN cd /usr/src/googletest/ && \
...
 
-COPY codecompass-build.sh /usr/local/bin
+COPY scripts/install_odb_2.5.sh /tmp/install_odb.sh
+
+RUN chmod u+x /tmp/install_odb.sh && \
+    /tmp/install_odb.sh
+
+
+# Erase downloaded archive files, and unneded packages.
+RUN set +x && \
```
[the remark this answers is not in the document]
filbeofITK: Since docker 17 (along with the multistage builds), the creation of multiple layers, does not increase the size of the container as significantly as it did before.
Also it would be more profitable for us if our image consisted of more layers, if we were aiming to use dockerhub in the future, because each layer gets pulled simultaneously, speeding up the process to the maximum capacity of one's computer/internet.

I'm going to omit the package removal part, then :smile:

filbeofITK:
> > Also, we need to consider people who do not with to install _anything_ non-package-manager driven to system-wide paths... the current script uses `/usr/local` which I'd rather not pollute with development stuff (it already contains plenty of vendor-specific stuff) that is not easily possible to uninstall.
> 
> For this to fix, I needed to add `${ODB_INCLUDE_DIRS}` to the `target_include_directories` here:
> 
> https://github.com/Ericsson/CodeCompass/blob/22be6a609531ec40a92e6498f2843caa4b3d1c4c/Functions.cmake#L43-L53

Could you help me out a bit please?
What else did you do, in order to help cmake find odb? I'm trying to export the path to the odb's .pc files, so when cmake does `pkg_check_modules(PC_LIBODB QUIET "libodb")` it works but it doesn't.

whisperity:
@filbeofITK GitHub's reply feature is stupid, it doesn't tag the person, please tag me to make sure I get the message!

I didn't do anything else. I have build2 downloaded and checked out at ~/Work/CodeCompass/Dependencies-Source. Same with Thrift and gtest.
The libraries on my local machine get installed to ~/Work/CodeCompass/Dependencies-Install. Everything goes to that directory. It look like this: 

```
╰─ tree -L 2 $(pwd)
~/Work/CodeCompass/Dependencies-Install
├── bin
│  ├── odb
│  ├── odb.so
│  └── thrift
├── include
│  ├── cutl
│  ├── gmock
│  ├── gtest
│  ├── libstudxml
│  ├── odb
│  ├── sqlite3.h
│  ├── sqlite3ext.h
│  └── thrift
├── lib
│  ├── libcutl-1.11.0-b.7.so
│  ├── libcutl.so -> libcutl-1.11.0-b.7.so
│  ├── libgmock.a
│  ├── libgmock_main.a
│  ├── libgtest.a
│  ├── libgtest_main.a
│  ├── libodb-2.5.0-b.17.so
│  ├── libodb-sqlite-2.5.0-b.17.so
│  ├── libodb-sqlite.a
│  ├── libodb-sqlite.so -> libodb-sqlite-2.5.0-b.17.so
│  ├── libodb.a
│  ├── libodb.so -> libodb-2.5.0-b.17.so
│  ├── libsqlite3-1.so
│  ├── libsqlite3.a
│  ├── libsqlite3.so -> libsqlite3-1.so
│  ├── libstudxml-1.1.0-b.8.so
│  ├── libstudxml.so -> libstudxml-1.1.0-b.8.so
│  ├── libthrift-0.13.0.so
│  ├── libthrift.a
│  ├── libthrift.la
│  ├── libthrift.so -> libthrift-0.13.0.so
│  ├── libthrift_c_glib.a
│  ├── libthrift_c_glib.la
│  ├── libthrift_c_glib.so -> libthrift_c_glib.so.0.0.0
│  ├── libthrift_c_glib.so.0 -> libthrift_c_glib.so.0.0.0
│  ├── libthrift_c_glib.so.0.0.0
│  ├── libthriftz-0.13.0.so
│  ├── libthriftz.a
│  ├── libthriftz.la
│  ├── libthriftz.so -> libthriftz-0.13.0.so
│  └── pkgconfig
└── share
   └── doc
```

I built `libodb-sqlite` by explicitly specifying the import path for `libodb`:

(executed from ~/Work/CodeCompass/Dependencies-Source/build2, where I downloaded to and executed the `build2-install-0.12.0.sh` script)

```
bpkg create --quiet --jobs $(nproc) --wipe -d . cc \
  config.cxx=/usr/bin/g++                       \
  config.cc.coptions=-O3               \
  config.bin.rpath=$(readlink -f ../../Dependencies-Install/lib)      \
  config.install.root=$(readlink -f ../../Dependencies-Install) \
  config.import.libodb=libodb-2.5.0-b.17
```

Before creating the build directory *for CodeCompass*, I source in a script that sets the paths to the proper versions, as described in the documentation:


~~~{.sh}
#!/bin/sh
CODECOMPASSS_DEPROOT="$(readlink -f ~/Work/CodeCompass/Dependencies-Install)"
LLVM_ROOT="$(readlink -f ~/Work/LLVM/Build)"

if [ -z "$(command -v ep)" ]
then
  echo "Fuck: No envprobe!" >&2
else
  ep ^CMAKE_PREFIX_PATH
  ep +CMAKE_PREFIX_PATH ${CODECOMPASS_DEPROOT} ${LLVM_ROOT}

  ep +PATH ${CODECOMPASS_DEPROOT}/bin ${LLVM_ROOT}/bin

  ep +LD_LIBRARY_PATH ${CODECOMPASS_DEPROOT}/lib ${LLVM_ROOT}/lib
fi
~~~

`ep +PATH asdasd` is a shorthand for `export PATH="asdasd:${PATH}"` basically.
So I add the "Dependencies-Install" tree (with bin and lib where appropriate) and the LLVM tree to the necessary paths. To be fair, LLVM is redundant because the build uses stock G++ 7.5.0 for current 18.04, and the LLVM 7.0 from the package manager as the library.

Created the build as follows:
```
╰─ cmake -DCMAKE_INSTALL_PREFIX="$(readlink -f ../Install)" -DDATABASE=sqlite -DCODECOMPASS_LINKER=gold -DCMAKE_BUILD_TYPE=Release ../Source
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
[...]
-- Found Threads: TRUE
-- Found Boost: /usr/include (found version "1.65.1") found components:  filesystem log program_options regex system thread date_time log_setup chrono atomic
-- Found Java: /usr/bin/java (found version "11.0.7")
-- Found ODB: ~/Work/CodeCompass/Dependencies-Install/bin/odb  found components:  sqlite missing components:  pgsql
-- Found Thrift version 0.13.0: ~/Work/CodeCompass/Dependencies-Install/bin/thrift
-- Found GTest: ~/Work/CodeCompass/Dependencies-Install/lib/libgtest.a
-- Found the following CodeCompass plugins:
--   cpp
--   cpp_reparse
--   dummy
--   git
--   metrics
--   search
Loading plugin cpp
-- Found LLVM 7.0.0
-- Using LLVMConfig.cmake in: /usr/lib/llvm-7/cmake
-- Using ClangConfig.cmake in: /usr/lib/cmake/clang-7
-- Using RTTI for LLVM: ON
Skipping generation of test project cpptest.
Loading plugin cpp_reparse
Loading plugin dummy
Loading plugin git
-- Found libgit2: /usr/lib/x86_64-linux-gnu/libgit2.so
Loading plugin metrics
Loading plugin search
-- Configuring done
-- Generating done
-- Build files have been written to: ~/Work/CodeCompass/filbetest
```

whisperity:
All my recent patches were based upon this branch: https://github.com/whisperity/CodeCompass/commits/working-master. There is only one commit, e0190b6b74df8b92419970470a5da53a73e2f7d9, on top of you adding a few includes here and there in this patch, which allowed me to get master working again.

doc/deps.md
```diff
@@ -91,12 +93,39 @@ by other processes which could, in extreme cases, make the system very hard or
 impossible to recover. **Please do NOT add a `sudo` in front of any `make` or
 other commands below, unless *explicitly* specified!**
 
+### Odb
```
whisperity, line 96: ODB

doc/deps.md
```diff
@@ -91,12 +93,39 @@ by other processes which could, in extreme cases, make the system very hard or
...
+Odb is an Object relational Mapping tool, that is required by CodeCompass.
```
whisperity, line 97: Relational

doc/deps.md
```diff
@@ -91,12 +93,39 @@ by other processes which could, in extreme cases, make the system very hard or
...
+We are using odb 2.5 which is not part of the official ubuntu repositories, so
```
whisperity, line 98: ODB, Ubuntu

doc/deps.md
```diff
@@ -91,12 +93,39 @@ by other processes which could, in extreme cases, make the system very hard or
...
 Procedure Call (RPC) between the server and the client. Thrift is not part of
 the official Ubuntu 16.04 LTS and 18.04 LTS repositories, but you can download
 it and build from source:
 
+Thrift can generate stubs for many programming languages. The configure
+script looks at the development environment and if it finds the environment
+for a given language then it'll use it. For example in the previous step npm
+was installed which requires NodeJS. If NodeJS can be found on your machine
+then the corresponding stub will also compile. If you don't need it then you
+can turn it off: ./configure --without-nodejs.
```
whisperity, line 121: Put commands between `s

doc/deps.md
```diff
@@ -106,21 +135,13 @@ cd thrift-0.13.0
...
-# CodeCompass.
-
-./configure --prefix=<thrift_install_dir> --with-python=NO --with-php=NO
+./configure --prefix=<thrift_install_dir> --silent --without-python \
+  --enable-libtool-lock --enable-tutorial=no --enable-tests=no      \
+  --with-libevent --with-zlib --without-nodejs --without-lua        \
+  --without-ruby --without-csharp --without-erlang --without-perl   \
+  --without-php --without-php_extension --without-dart              \
+  --without-haskell --without-go --without-rs --without-haxe        \
+  --without-dotnetcore --without-d --without-qt4 --without-qt5;     \
```
whisperity, line 144: Is this command terminated properly? `;` and then a `\`

docker/dev/Dockerfile
```diff
@@ -19,7 +19,8 @@ RUN set -x && apt-get update -qq \
     libboost-filesystem-dev \
     libboost-log-dev \
     libboost-program-options-dev \
-    libboost-regex-dev
+    libboost-regex-dev \
+    curl
```
whisperity, line 23: Let's make this install line sorted alphabetically if we're already touching it.

docker/dev/Dockerfile
```diff
@@ -35,7 +36,7 @@ RUN tar -xf /opt/thrift-0.13.0.tar.gz -C /opt; \
     --without-dotnetcore --without-d --without-qt4 --without-qt5;    \
     cd /; \
     make -C /opt/thrift-0.13.0 -j $(nproc) --silent install; \
-    rm -rf /opt/thrift-0.13.0 /opt/thrift-0.13.0.tar.gz
+    rm -rf /opt/thrift-0.13.0
```
whisperity, line 39: Where did the targz go?

scripts/install_odb_2.5.sh
```diff
@@ -0,0 +1,84 @@
...
+            ;;
+        *)
+            usage >&2
+            exit 1
+            ;;
+    esac
+done
+
+
+#Build2 toolchain
```
whisperity, line 56: Please put a space between # and the comment body, at all places

scripts/install_odb_2.5.sh
```diff
@@ -0,0 +1,84 @@
...
+mkdir /tmp/build2src
+pushd "/tmp/build2src"
```
whisperity, line 58: To make sure we do not clash with anything on the system even if running outside of docker, use `mktemp -d` with an optional suffix or something.

scripts/install_odb_2.5.sh
```diff
@@ -0,0 +1,84 @@
...
+curl -sSfO https://download.build2.org/0.12.0/build2-install-0.12.0.sh
+sh build2-install-0.12.0.sh --yes --trust yes
+popd
+
+
+#Configuring the build
+mkdir /tmp/odb
+pushd "/tmp/odb"
```
whisperity, line 66: `mktemp -d` here too

docker/dev/Dockerfile
filbeofITK: It was pointed out by @csordasmarton that deleting a file in a separate layer does not decrease the container size, so it was unnecessary.

whisperity: Ah, right. Either way, to reduce the size of the container, you could add the curl command fetching the file to this `RUN`, or use the "multi-stage build" thing where you have multiple `FROM`s and in one container build Thrift and in the other one just copy the installation over.

doc/deps.md
filbeofITK: No, I just copied it from the Dockerfile, and forgot to remove the last \.

docker/dev/Dockerfile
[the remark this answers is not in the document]
filbeofITK: @csordasmarton 
I did some reading on the topic and docker has a feature, to squash all layers into one,  by only keeping the differences, like git does. If we wish to minimize the image size and also keep the advantages of multiple layers while developing, with `docker build --squash` we can.
Because that way we could delete the unneeded packages and data in one, last additional layer, making the final image size even smaller. (This could help to migrate a built image to Dockerhub maybe)
And docker still caches the intermediate layers this way, on the host machine so rebuilding the image does not loose it's incremental features.

doc/deps.md
```diff
@@ -11,16 +11,13 @@ be installed from the official repository of the given Linux distribution.
 
 - **`git`**: For fetching and managing CodeCompass source code.
 - **`cmake`** and **`make`**: For building CodeCompass.
-- **`g++`**: For compiling CodeCompass. A version which supports C++14 features
-  is required. (Alternatively, you can compile with Clang.)
+-**`gcc-X`**: Version 7 or higher.
```
mcserep, line 14: Do we really need version 7 or higher? On Xenial ODB is built with version 5 even in the CI.

filbeofITK: Yeah I wanted to write 5 there.

mcserep:
As requested by both @whisperity , @csordasmarton and me, we did not create separate install scripts for other dependencies to install, so neither should one exist for ODB. Therefore I have refactored your bash script into the markdown @filbeofITK , please review it.
Sorry if you have put extra work into that, but we shall keep the installation procedure consistent. It was still very useful and I had an easy job thanks to that.

filbeofITK:
> n

It seems great.
Now we should refactor the Dockerfile accordingly, because right now it uses the script.

mcserep:
> It seems great.
> Now we should refactor the Dockerfile accordingly, because right now it uses the script.
That's true.

Another question: I have started testing this PR. Building on Bionic with the new ODB now succeeds, the `CodeCompass_parser` binary also works fine. On the other hand, `CodeCompass_webserver` fails with a segfault immediately (with SQLite). Before I move on to debug this: did you face a similar issue or is it working for you?

filbeofITK:
> 
> 
> > It seems great.
> > Now we should refactor the Dockerfile accordingly, because right now it uses the script.
> 
> That's true.
> 
> Another question: I have started testing this PR. Building on Bionic with the new ODB now succeeds, the `CodeCompass_parser` binary also works fine. On the other hand, `CodeCompass_webserver` fails with a segfault immediately (with SQLite). Before I move on to debug this: did you face a similar issue or is it working for you?

I only tested whether it builds and logs so I dont know.  Sorry.

whisperity:
>`CodeCompass_webserver` fails with a segfault immediately (with SQLite). Before I move on to debug this: did you face a similar issue or is it working for you?

@mcserep I used @filbeofITK's patches on a locally merged branch during development of #386, #387, #388, and used SQLite backend (so bloat to install on the machine is less...), and saw no segfaults related to the database.

mcserep:
> > `CodeCompass_webserver` fails with a segfault immediately (with SQLite). Before I move on to debug this: did you face a similar issue or is it working for you?
> 
> @mcserep I used @filbeofITK's patches on a locally merged branch during development of #386, #387, #388, and used SQLite backend (so bloat to install on the machine is less...), and saw no segfaults related to the database.

Thanks for the info @whisperity , meanwhile I checked and with PostgreSQL, it also works for me. I will debug with SQLite tomorrow.

mcserep:
@whisperity It seems the segfault comes from the [sqlite3_create_function_v2()](https://github.com/Ericsson/CodeCompass/blob/ed26e0f2860ab821d153f2845129b7a136f06d53/util/src/dbutil.cpp#L341) function call in `dbutil.cpp`.

What is strange, that while the `CodeCompass_parser` binary is linked against *libodb-sqlite* (and for some reason also *libodb-pgsql*?):
```
ldd CodeCompass_parser 
	linux-vdso.so.1 (0x00007ffca6920000)
	libboost_filesystem.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.65.1 (0x00007f9053c57000)
	libboost_log.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_log.so.1.65.1 (0x00007f905397a000)
	libboost_program_options.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_program_options.so.1.65.1 (0x00007f90536f9000)
	libboost_system.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_system.so.1.65.1 (0x00007f90534f4000)
	libboost_thread.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1 (0x00007f90532cf000)
	libodb-2.5.0-b.17.so => /usr/local/lib/libodb-2.5.0-b.17.so (0x00007f90530a8000)
	libodb-pgsql-2.5.0-b.17.so => /usr/local/lib/libodb-pgsql-2.5.0-b.17.so (0x00007f9052e74000)
	libodb-sqlite-2.5.0-b.17.so => /usr/local/lib/libodb-sqlite-2.5.0-b.17.so (0x00007f9052c45000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9052a41000)
	libmagic.so.1 => /usr/lib/x86_64-linux-gnu/libmagic.so.1 (0x00007f905281f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9052600000)
	libboost_regex.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_regex.so.1.65.1 (0x00007f90522f8000)
	libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007f9051fef000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f9051c66000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9051a4e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f905165d000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9051455000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9054256000)
	libpq-5.9.so => /usr/local/lib/libpq-5.9.so (0x00007f9051225000)
	libsqlite3-1.so => /usr/local/lib/libsqlite3-1.so (0x00007f9050ec9000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f9050cac000)
	libicui18n.so.60 => /usr/lib/x86_64-linux-gnu/libicui18n.so.60 (0x00007f905080b000)
	libicuuc.so.60 => /usr/lib/x86_64-linux-gnu/libicuuc.so.60 (0x00007f9050453000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f90500b5000)
	libicudata.so.60 => /usr/lib/x86_64-linux-gnu/libicudata.so.60 (0x00007f904e50c000)
```

The `CodeCompass_webserver` binary is not:
```
ldd CodeCompass_webserver 
	linux-vdso.so.1 (0x00007fff73e59000)
	libboost_filesystem.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.65.1 (0x00007f4985182000)
	libboost_log.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_log.so.1.65.1 (0x00007f4984ea5000)
	libboost_program_options.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_program_options.so.1.65.1 (0x00007f4984c24000)
	libboost_system.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_system.so.1.65.1 (0x00007f4984a1f000)
	libboost_thread.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1 (0x00007f49847fa000)
	libodb-2.5.0-b.17.so => /usr/local/lib/libodb-2.5.0-b.17.so (0x00007f49845d3000)
	libodb-pgsql-2.5.0-b.17.so => /usr/local/lib/libodb-pgsql-2.5.0-b.17.so (0x00007f498439f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f4984180000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f4983f7c000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f4983bf3000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f49839db000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f49835ea000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f49833e2000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f498566c000)
	libpq-5.9.so => /usr/local/lib/libpq-5.9.so (0x00007f49831b2000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f4982e14000)

```

mcserep:
Another issue with the new ODB library is that the `FindOdb.cmake` script we use seems to be not fully compatible with it. The `${ODB_INCLUDR_DIRS}` variable is empty after the `find_package()` call, so in case ODB is not installed on a standard path, CodeCompass will not compile. (That's why the CI is failing currently.)

whisperity:
> Another issue with the new ODB library is that the `FindOdb.cmake` script we use seems to be not fully compatible with it. The `${ODB_INCLUDR_DIRS}` variable is empty after the `find_package()` call, so in case ODB is not installed on a standard path, CodeCompass will not compile. (That's why the CI is failing currently.)

It should be `ODB_INCLUDE_DIRS`, not `INCLUDR_DIRS` :smile: 

> What is strange, that while the CodeCompass_parser binary is linked against libodb-sqlite (and for some reason also libodb-pgsql?):

My mistake, I left out the crucial detail that I had never even compiled the Postgres libodb on my machine, **only** SQLite.

mcserep:
> It should be `ODB_INCLUDE_DIRS`, not `INCLUDR_DIRS` 

@whisperity I copied it from `FindOdb.cmake` together with the typo; but this typo is only present in the header description comment, not in any executed code.
So, according to the CI, the include path is still incorrect, unfortunately this issue was not that easy to solve :)

mcserep:
Your suggestion worked @whisperity , see a787af2 . Now the CI builds.

The webserver segfault issue (with sqlite) remains to solve.

whisperity:
@mcserep Strange, it didn't work for filbeof a month ago, see: https://github.com/Ericsson/CodeCompass/pull/385#issuecomment-618987193.

doc/deps.md
```diff
@@ -91,12 +92,63 @@ by other processes which could, in extreme cases, make the system very hard or
...
+bpkg build odb --yes
+bpkg build libodb --yes
+bpkg build libodb-sqlite --yes
+bpkg build libodb-pgsql --yes
+bpkg install --all --recursive # you may to run the instal with sudo, depending on the install dir
+```
+Please take into consideration that the odb installation can take up a long time (depending on the machine one is using),
+but you can increase the used threads with the `--jobs` option.
+
+*Note:* now you max delete the *build2* toolchain installed in the `<build2_install_dir>` folder, if you do not need any longer.
```
whisperity, line 131:
```suggestion
*Note:* now you may delete the *build2* toolchain installed in the `<build2_install_dir>` folder, if you do not need any longer.
```

doc/deps.md
```diff
@@ -91,12 +92,63 @@ by other processes which could, in extreme cases, make the system very hard or
...
+bpkg fetch --trust-yes
+
+# Building odb
+bpkg build odb --yes
+bpkg build libodb --yes
+bpkg build libodb-sqlite --yes
+bpkg build libodb-pgsql --yes
+bpkg install --all --recursive # you may to run the instal with sudo, depending on the install dir
+```
+Please take into consideration that the odb installation can take up a long time (depending on the machine one is using),
```
whisperity, line 128:
```suggestion
Please take into consideration that the ODB installation can take up a long time (depending on the machine one is using),
```

doc/deps.md
```diff
@@ -91,12 +92,63 @@ by other processes which could, in extreme cases, make the system very hard or
 impossible to recover. **Please do NOT add a `sudo` in front of any `make` or
 other commands below, unless *explicitly* specified!**
 
+### ODB
+ODB is an Object Relational Mapping tool, that is required by CodeCompass.
+We are using ODB 2.5 which is not part of the official Ubuntu repositories, so
+it needs to be built manually.
+
+The odb installation uses the build2 build system.
+(This is not needed for CodeCompass so you may delete it right after the installation of odb.)
```
whisperity, line 101:
```suggestion
The ODB installation uses the build2 build system.
(Build2 is not needed for CodeCompass so you may delete it right after the installation of ODB.)
```

whisperity:
I'll do a full-on check of the build guide, but unfortunately can only do tomorrow. :frowning_face:

mcserep:
I have debugged the source of the SQLite issue.

**Problem description:**
When CodeCompass (the version of this PR) is compiled and linked with the new ODB, and configured to use SQLite, the compilation is successful, the parsing is done, but the webserver cannot start, fails with a segfault.

**Reason of error:**
The `cpp_reparse` plugin fails upon connecting to the SQLite database. The segfault comes from the [sqlite3_create_function_v2()](https://github.com/Ericsson/CodeCompass/blob/ed26e0f2860ab821d153f2845129b7a136f06d53/util/src/dbutil.cpp#L341) function call in `dbutil.cpp`
The stack trace is:
```
util/src/dbutil.cpp:341
webserver/include/wevserver/pluginhelper.h:68
plugins/cpp_reparse/service/src/plugin.cpp:33
```

The issue is only present with the `cpp_reparse` plugin. By disabling this plugin (removing its .so file from the install directory), the webserver can start and operates seemingly correctly (as far as I have tested).

Do you have the time to take a more detailed look on this @whisperity , as the author of this plugin? Thanks!

whisperity:
@mcserep I can certainly try to look at it, but I can't promise anything, it's been years since that plugin was conceived. And what the f.... I don't remember the reparse plugin using anything directly from SQLite. It's just a pure plugin like every other, using the same database interface as every other plugin. What even is this `create_function_v2` line, involving regexes and whatnot... I don't recall using **any** regex *in the database* (???) in the reparse plugin...

Is this a configuration where you compiled both ODB PgSQL and ODB SQLite? What happens if only one of the libraries for ODB is compiled and linked against CodeCompass?

mcserep:
@whisperity Guess I have managed to solve it. Here is the case:
- Until now `FindOdb.cmake` was configured to try to find both *sqlite* and *pgsql* components; but somehow (not clear right now for me how) it only linked against the proper libary.
- Now with ODB 2.5.0 beta it links against **both** libraries which for some another reason mess up the `cpp_reparse` plugin (which I did not analyze how it works).

So in 57ba7bb I modified `FindOdb.cmake` to only find the ODB component which is set in the `DATABASE` variable. As a result only the needed component will be found and linked, and `cpp_reparse` does not segfault on webserver startup.

mcserep:
From me it's an approve on this PR, thanks for your work @filbeofITK . Please write if you have any major concerns @whisperity , @csordasmarton  and @bruntib  and let's get this PR merged ASAP, so the master branch could be compiled again on Bionic.
Since the issue is critical, I advise that minor concerns shall go to a separate PR.

doc/deps.md
```diff
@@ -106,21 +158,13 @@ cd thrift-0.13.0
 
 # Ant is required for having Java support in Thrift.
 sudo apt-get install ant
```
whisperity, line 160: Maybe this is for me, but simply running the build on 18.04 does not work in 2020. I'm testing the build on a brand new virtual machine installed this morning.
Even though [Gradle started not to support HTTP download](https://blog.gradle.org/decommissioning-http) in 2020, Thrift 0.13 -- released 2019 Oct -- is still using an HTTP URL. The following output happens during the build, towards the end:

~~~
make[3]: Entering directory '/home/codecompass/thrift-0.13.0/lib/java'
./gradlew  assemble \
        -Prelease=true \
        -Pthrift.version=0.13.0 \
        --console=plain
> Task :copyDependencies FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':copyDependencies'.
> Could not resolve all files for configuration ':testRuntime'.
   > Could not resolve org.slf4j:slf4j-api:1.7.25.
     Required by:
         project :
      > Could not resolve org.slf4j:slf4j-api:1.7.25.
         > Could not get resource 'http://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.25/slf4j-api-1.7.25.pom'.
            > Could not GET 'http://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.25/slf4j-api-1.7.25.pom'. Received status code 501 from server: HTTPS Required
   > Could not resolve org.apache.httpcomponents:httpclient:4.5.6.
     Required by:
         project :
      > Could not resolve org.apache.httpcomponents:httpclient:4.5.6.
         > Could not get resource 'http://repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.5.6/httpclient-4.5.6.pom'.
            > Could not GET 'http://repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.5.6/httpclient-4.5.6.pom'. Received status code 501 from server: HTTPS Required
   > Could not resolve org.apache.httpcomponents:httpcore:4.4.1.
~~~

The relevant breaking line is in `lib/java/gradle.properties`:
~~~{.gradle}
mvn.repo=http://repo1.maven.org/maven2
~~~

Swapping HTTP to HTTPS in the Gradle cfg fixes this:

```suggestion
sudo apt-get install ant
# Fix Thrift install script not using HTTPS for Java dependencies (http:// -> https://)
sed 's/http:\/\//https:\/\//g' -i ./lib/java/gradle.properties
```

doc/deps.md
```diff
@@ -13,6 +13,7 @@ be installed from the official repository of the given Linux distribution.
 - **`cmake`** and **`make`**: For building CodeCompass.
 - **`g++`**: For compiling CodeCompass. A version which supports C++14 features
   is required. (Alternatively, you can compile with Clang.)
+- **`gcc-X`, `gcc-X-plugin-dev`**: For building ODB.
 - **`libboost-all-dev`**: Boost can be used during the development.
 - **`llvm-7-dev`**, **`libclang-7-dev`**: C++ parser uses LLVM/Clang for
```
whisperity, line 18: `.travis.yml` installs `clang-7` too. Unfortunately, this is needed so CMake's `FindClang` works...

```suggestion
- **`llvm-7-dev`**, **`clang-7`**, **`libclang-7-dev`**: C++ parser uses LLVM/Clang for
```

whisperity:
> @whisperity Guess I have managed to solve it. Here is the case:
> 
>     * Until now `FindOdb.cmake` was configured to try to find both _sqlite_ and _pgsql_ components; but somehow (not clear right now for me how) it only linked against the proper library.
> 
>     * Now with ODB 2.5.0 beta it links against **both** libraries which for some another reason messes up the `cpp_reparse` plugin (which I did not analyze how it works).
>

@mcserep Right, thanks! I was thinking it's some *weird* link-level ODR yadda... Anyways, we shan't depend on the database that is *not* to be used by the build, so this clears it up.

whisperity:
:tada: And happy news: the built binaries also work. I have parsed the built CodeCompass, with all navigation, reparse, search, etc.

Screenshot_20200529_123632
Screenshot_20200529_123422

whisperity:
@mcserep Please check my suggestions above, especially about the system packages. There is a dissonance between the Travis- and the Dockerfile vs. going on the guide manually which is not user friendly. Otherwise, thank you for all the work coordinating this patch! :trophy:

whisperity:
@filbeofITK I'm not sure. I'd vote for not resorting to self-compile on a system where it works. Ubuntu 16.04 will have it's support terminated in less than a year... Until we do not depend on anything from ODB 2.5.0 that is not available in 2.4.0, let's not require users to do extra steps.

Please can you squash the commits and tidy the patch out for merging? It has become 25 commits by now, most of which are few-liners.

mcserep:
> Please can you squash the commits and tidy the patch out for merging? It has become 25 commits by now, most of which are few-liners.

I think we can depend on the "Squash and merge" feature of GitHub. The one who merges shall select that button and everything will be squashed properly.

doc/deps.md
```diff
@@ -61,7 +62,7 @@ sudo apt-get install git cmake make g++ libboost-all-dev \
 #### Ubuntu 18.04 LTS
 
 ```bash
-sudo apt-get install git cmake make g++ libboost-all-dev \
+sudo apt-get install git cmake make g++ gcc-7-plugin-dev libboost-all-dev \
```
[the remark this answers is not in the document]
mcserep: I guess this code suggestion is a typo, but instead `libssl-dev` shall be replaced with `libssl1.0-dev` in the Ubuntu 18.04 install guide?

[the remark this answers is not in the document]
mcserep: On the other hand, wouldn't it be a better solution to install a newer version of nodejs by the users from the official PPA? See very simple instructions here:
https://github.com/nodesource/distributions

E.g. I have nodejs-11 installed from the official PPA and `libssl-dev` from the standard package repository. That way I did not have to degrade to `libssl1.0-dev`. Since other packages I use depend on `libssl-dev`, degrading was not an option for me.

doc/deps.md
```diff
@@ -27,7 +28,7 @@ be installed from the official repository of the given Linux distribution.
 - **`default-jdk`**: For search parsing CodeCompass uses an indexer written in
   Java.
 - **`libssl-dev`**: OpenSSL libs are required by Thrift.
```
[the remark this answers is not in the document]
mcserep: See my previous comment for possible other workaround: ugrade nodejs instead of degrading libssl.

doc/deps.md
```diff
@@ -61,7 +62,7 @@ sudo apt-get install git cmake make g++ libboost-all-dev \
 #### Ubuntu 18.04 LTS
 
 ```bash
-sudo apt-get install git cmake make g++ libboost-all-dev \
+sudo apt-get install git cmake make g++ gcc-7-plugin-dev libboost-all-dev \
   llvm-7-dev libclang-7-dev odb libodb-dev \
   default-jdk libssl-dev libgraphviz-dev libmagic-dev libgit2-dev ctags \
```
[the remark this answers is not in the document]
mcserep: See my previous comment for possible other workaround: ugrade nodejs instead of degrading libssl.

doc/deps.md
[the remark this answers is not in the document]
whisperity: We can resolve this later. I merely wanted to make sure the human-readable guide and the script "guide" are in sync.

If you are a power user who understands conflicts of packages, you'll most likely find a way to fix it. I tended in the way of the guard that if someone doesn't *really* know how this works in the big picture could also get it going.

Also, using custom PPAs... Well... I think I will be a minority, unfortunately, there is an upcoming corporate regulation that we cannot add custom PPAs to our machine. :facepalm:

doc/deps.md
```diff
@@ -91,12 +92,63 @@ by other processes which could, in extreme cases, make the system very hard or
...
+it needs to be built manually.
+
+The ODB installation uses the build2 build system.
+(Build2 is not needed for CodeCompass so you may delete it right after the installation of ODB.)
+```bash
+wget https://download.build2.org/0.12.0/build2-install-0.12.0.sh
+sh build2-install-0.12.0.sh --yes --trust yes "<build2_install_dir>"
+```
+
+Now, utilizing the *build2* toolchain, we can build the *odb* library. In the script below, we assume that ODB is built in the `<odb_build_dir>` directory and installed in the `<odb_install_dir>` directory. A good default install directory could be `/usr/local/`.
```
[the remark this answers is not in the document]
mcserep: Actually it's very easy to clean up what ODB-2.5 installs even manually. In single-user developer environment self-compiled tools can go to `/usr/local`, because that is the folder for it by definition. A production system is different.
But we can remove that line.

doc/deps.md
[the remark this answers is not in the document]
mcserep: Valid point, let's update the docs this way for now, I have applied your suggestions.

doc/deps.md
mcserep: I will check this in a Docker container on a fresh system soon.

Also, we are at this point of the guide, we don't need Ant to be installed. I don't have ant installed. The Docker environment does not install ant.

mcserep:
> @mcserep Please check my suggestions above, especially about the system packages. There is a dissonance between the Travis- and the Dockerfile vs. going on the guide manually which is not user friendly. Otherwise, thank you for all the work coordinating this patch!

Thanks for the through review of the PR and the install guide @whisperity !

I think the ODB version question on Ubuntu 16.04 is though, as both options has advantages and backwards. While it is true that uing ODB 2.5-beta eliminates any version difference issue between the 2 supported Ubuntu version, it also adds the additional complexity of compiling ODB, when it is not really necessary. If the developers behind ODB are nice guys and take semantic versioning seriously, there should be no API change between ODB 2.4 and 2.5.

Since package-repository installed dependencies may already diverge between Xenial and Bionic, I guess it is not so bad if the ODB version differs too. So I agree with @whisperity in this topic.

doc/deps.md
```diff
@@ -52,7 +53,7 @@ repositories:
 sudo deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-7 main
 sudo deb-src http://apt.llvm.org/xenial/ llvm-toolchain-xenial-7 main
 
-sudo apt-get install git cmake make g++ libboost-all-dev \
+sudo apt-get install git cmake make g++ gcc-5-plugin-dev libboost-all-dev \
   llvm-7-dev libclang-7-dev odb libodb-dev \
```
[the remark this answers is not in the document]
mcserep: Fixed in e15620c566a67b584acdcdf298eeda3f7925f692

.travis.yml
```diff
@@ -22,13 +22,10 @@ services:
...
     - default-jdk
     # ODB
-    - odb
-    - libodb-dev
+    - gcc-5-plugin-dev
     # SQLite
-    - libodb-sqlite-dev
     - libsqlite3-dev
     # PostgreSQL
-    - libodb-pgsql-dev
```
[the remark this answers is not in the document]
mcserep: Fixed in e15620c566a67b584acdcdf298eeda3f7925f692

doc/deps.md
mcserep: My conclusion is that for CodeCompass we only need the **Java generator** and not the **Java library**, which we have also discussed through emails last November.

Without Ant the Java library won't be built and there is no *HTTP->HTTPS* issue with Gradle. (We could even add the `--without-java` flag to Thrift, actually the CI was building CodeCompass that way for quite a long time previously.)

whisperity: @mcserep Did you check this with running a server and trying out the search parser and search service? Those are our only upstream stuff that "depend on" Java, but I am not sure about their inside architecture.

PR #388, opened 2020-05-01
Update Mongoose to newest release version (6.18)

whisperity:
The Mongoose library major version we were using (5.4) is from 2014 and is no longer actively developed. There is a newer version of the 6.y line, however, moving to this version required a **significant** rewrite of the threaded response handling.

> Originally my first hunch of implementing SSL for #387 was to update Mongoose and try, I only afterwards found out that 5.4 also supports SSL.

The previous behaviour was that we were running on each thread a Mongoose server that was listening on the same (manually injected) socket. This behaviour is not supported any more, the new Mongoose version threw away "real" multithreading support, and the API no longer allows injecting the socket.
(In addition, I was told that the same process on `n` threads listening on the same socket resulted in the kernel waking up *every* thread whenever a request was received.)

To align the API with the new version, a different approach is used: incoming HTTP requests, if are of trivial nature (static file access, which will also help for this patch to be joined together with #386!) are served as if Mongoose was serving a static directory - there were also **massive** API changes here.
Otherwise, the request will be saved into a thread pool's queue, similar to the one used by the parser engine, and handled by dispatching to the Thrift endpoints.

The indicators `MG_FALSE` and `MG_TRUE` are also gone, the new API requires for the user of the library to manually handle doing A or B when an URL is accessed. Previously Mongoose got to know that the callback didn't do anything with the request and fell back by itself to serving the static resource.

The upside of this is that there is a dedicated thread for listening on the socket, and as such, heavy-handed operations will not lock static page accesses from being served. Of course, the number of threads for the service workers is an upper limit to how much Thrift requests can be served at a time.

Most of the changes in this patch are of a highly technical nature:

 * Need to bind the listener with new API.
 * The request handler callback (from Mongoose's point of view) does not directly handle the request, and as such, returning from it would destroy the request buffer, so we have to side-step this issue by saving the important data out.
 * Performing the same decoupling on the handlers' API.


plugins/cpp/test/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ include_directories(
   ${PROJECT_BINARY_DIR}/service/language/gen-cpp
   ${PROJECT_BINARY_DIR}/service/project/gen-cpp
   ${PROJECT_SOURCE_DIR}/model/include
+  ${PROJECT_SOURCE_DIR}/webserver/include
```
whisperity, line 7: Changes in other headers broke the test's file.
The test's header includes the service header, and as such, this subproject also needs to know how to resolve the `#include <webserver/servercontext.h>` in `cppservice.h`.

util/include/util/threadpool.h
```diff
@@ -4,6 +4,7 @@
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
+#include <functional>
```
whisperity, line 7: The default for a template parameter is `std::function`. This include was needed ever since #120, but the files this header was included to already made `<functional>` available.

webserver/src/webserver.cpp
```diff
@@ -1,3 +1,7 @@
+// Ensure that Mongoose API is defined the same way in all translation units!
+#include <webserver/mongoose.h> // IWYU pragma: keep
```
whisperity, line 2: This was weird to debug. Basically, if C++ headers are included before mongoose, there'll be some weird issues and the `static_assert` in `mongooseserver.h` will fire... Which shouldn't because the typedef and all that remains the exact same thing. Digging through a Clang AST I was able to find out that somewhere deep `__falign_t` becomes defined in a different way.

Without the static assert, the code compiles, and I did not find any issues run-time, but it's better to be ODR-safe.

webserver/src/webserver.cpp
```diff
@@ -8,15 +12,21 @@
...
 #include <util/filesystem.h>
 #include <util/logutil.h>
+#include <util/threadpool.h>
 #include <util/webserverutil.h>
 
-#include "threadedmongoose.h"
+#include <webserver/httprequest.h>
+
 #include "mainrequesthandler.h"
+#include "mongooseserver.h"
```
whisperity, line 21: If these include were in a different order (keeping the alphabetic style), `webserver.cpp` would not compile.

webserver/src/mongooseserver.h
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Callback function for the Mongoose server that handles the parsing of
+   * incoming requests.
+   */
+  static void
+  nativeRequestHandler(struct mg_connection* conn_, int ev_, void* data_);
+  static_assert(
+    std::is_same<decltype(&nativeRequestHandler), ::mg_event_handler_t>::value,
+    "Mongoose-interfacing handler function should match signature!");
```
whisperity, line 116: This is the static assert I (will be|was) talking about. It basically just says: make sure the interface of the function pointer you are passing actually matches the receiver's side.

webserver/src/webserver.cpp
```diff
@@ -92,35 +102,68 @@ int main(int argc, char* argv[])
...
+        // implementation.
+        requestHandler(req_);
+        server.wakeUp();
+      },
+      true);
+
+  // The callback function of the server itself is to push a received and
+  // understood request to the worker threads' queue.
+  server.setHTTPServiceRequestHandler(
+    std::bind( // NOLINT: Need to pass function obj!
```
whisperity, line 156: Clang-Tidy 10.0 would whine that I should pass a lambda instead of a bind... but unfortunately, the other side is a `std::function` to which a lambda with a capture group would not compile.

whisperity: Actually, Clang-Tidy was right for the warning, just the automatic fix-it didn't compile by itself. Starting from C++14, you can actually give a lambda with a capture to an `std::function`, just need to do copy initialisation instead of truly capturing. Which is okay, given we uniquely own the thread-pool.

whisperity:
Rebased over #386.

mcserep:
> With this patch the page load and every network request take a few seconds. Everything became slow, so this needs some debugging. I'll try to investigate too. Apart from this, the PR seems to be fine.

I can verify @bruntib observation, with this PR, a HTTP request towards e.g. `ProjectService` takes more than 1 sec. Previously it was around 4-7 ms, so it got slower about 150-250 times, which cannot be a normal functionality. Tested on localhost, in Release build.

whisperity:
Huh, weird, I haven't measured a difference when I made this. Well, one reason could be, is that the handling of the request is a blocking operation because Mongoose no longer allows on the API to share the listening socket between individual threads...

I am, unfortunately, out of ideas, then. Apart from sticking to the old Mongoose ad infinitum.

webserver/include/webserver/httprequest.h
```diff
@@ -0,0 +1,54 @@
...
+ * Contains the parsed values of an incoming HTTP request as understood by the
+ * server. Unlike Mongoose's buffers, this record owns its resources directly.
+ */
+struct HTTPRequest
+{
+  struct mg_connection* connection;
+
+#if MG_ENABLE_IPV6 == 1
+  // '[' + sizeof(ipv6-address) + ']' + strlen(":12345")
+  char address[1 + INET6_ADDRSTRLEN + 1 + 6];
```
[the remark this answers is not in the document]
whisperity: Both `INET_ADDRSTRLEN` and `INET6_ADDRSTRLEN` already calculate the null terminator into their length. For example, the IPv4 version is defined to be **16**, which means `255.255.255.255` (4 * 3 + 3 = 15 real characters) fits perfectly into a buffer of `char[INET_ADDRSTRLEN]`.

mcserep:
@whisperity I don't think it is a threading (blocking) issue, since I have measured it on a single HTTP request, making sure that there were no other requests. (Also if Moongose's performance would this poor under normal usage, it would render it unusable, so my guess is that we are doing something not correctly.)

It would be nice to update Moongose, maybe we can execute a performance test on `CodeCompass_webserver` (e.g. with CLion and perf) and check where does it spends so much time when handling a request.

whisperity:
I'm leaving the patch up, but I do not think I'll have the mental capacity and acuity to reverse engineer and dissect this patch in the foreseeable future.

mcserep:
> I'm leaving the patch up, but I do not think I'll have the mental capacity and acuity to reverse engineer and dissect this patch in the foreseeable future.

Okay, I might add some information meanwhile if I get to debug this.
Right now the addition I can make is that static files are served normally, only our services are served slowly.

webserver/CMakeLists.txt
```diff
@@ -1,22 +1,59 @@
...
+  MG_ENABLE_CALLBACK_USERDATA=0
+  MG_ENABLE_DNS=0
+  MG_ENABLE_FILESYSTEM=1
+  MG_ENABLE_HTTP_WEBSOCKET=0
+  MG_ENABLE_HTTP_URL_REWRITES=0
+  MG_ENABLE_IPV6=0
+  MG_ENABLE_MQTT=0
+
+  # Only exposes a pthread_create() API, changes no behaviour w.r.t CodeCompass.
+  MG_ENABLE_THREADS=0
```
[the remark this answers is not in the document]
whisperity: I went purely by reading `mongoose.h` and `mongoose.c`. As the code is meant for embedded system, the header (and to some extent, the implementation code), is relatively straightforward to read.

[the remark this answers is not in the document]
whisperity: `MG_ENABLE_DEBUG` simply results in a very obscure dump of steps taken and enum values for each request. It's *somewhat* useful for debugging if you manually trace the requests and such, but not useful for production system. (It's set to 0 because is a **build-time** flag, due to how Mongoose's code is structured, you can't make this dependent on `--loglevel debug` at runtime.)

Mongoose is not just a server, it can also be used as a **client** library. `MG_ENABLE_DNS` (defaulting to `1`) toggles whether, if used *as a client*, the code has the ability to do DNS resolution. Because we are not using Mongoose as a client, we do not need this functionality.

(`_IPV6` and `_MQTT` similarly disable networking stacks that we are not using.)

`MG_ENABLE_ASYNC_RESOLVER` is related to this, too. (If enabled, *clientside* DNS resolution is done asynchronously, by default via `MG_DEFAULT_NAMESERVER "8.8.8.8"`.

`MG_ENABLE_FILESYSTEM` is to be used if we are serving a conventional HTTP root (i.e. like Apache `Options Index`) and static files. This can't be used, as we have special endpoints (for Thrift) which would conflict, if Mongoose was to serve the filesystem natively.

`MG_ENABLE_HTTP_URL_REWRITES` is explained pretty well in the header:
Screenshot_20200728_154040

`MG_ENABLE_CALLBACK_USERDATA` is a flag that changes the signatures of Mongoose handlers to take an additional userdata argument, like this:
Screenshot_20200728_154303
In many functions, if this is enabled, an additional `void *` argument is present in the functions' signatures, which allows registering/passing an arbitrary object through Mongoose.

mcserep:
What is the status of this PR, @whisperity ? Do you think you will have the time to get back on this issue anytime in the future? It would be really nice to upgrade the obsolete Mongoose version we are using. As I have checked, version **7.x** came out December 2020, and we still use version 5.4.

mcserep:
I have updated this development to be sync with the current master:
https://github.com/mcserep/CodeCompass/tree/update-mongoose

If you enable in the PR settings to _"Allow edits by maintainers"_, then I can update it on your fork, @whisperity.

More importantly, I have made progress towards the reason of the slow page load time of the Mongoose 6 refactored webserver. The key is the timeout parameter for `mg_mgr_poll()`.

```cpp
void MongooseServer::loop()
{
  // Trap termination so the server stops.
  _detail::SignalHandler trapSigint(SIGINT, MongooseServer::signalHandler);
  _detail::SignalHandler trapSigterm(SIGTERM, MongooseServer::signalHandler);

  _loopThreadId = std::this_thread::get_id();

  while (!exitCode)
    mg_mgr_poll(_mongoose.get(), 1000);
}
```

With the value 1000, all (non-statically served) requests take at least 1000 milliseconds. If I set the parameter to 20, then requests are served in 24-26 ms. If I set the parameter to 10, then requests are served in 14-16 ms. However, if I set the parameter to 0, then some requests are failed to be served.

This timeout parameter of Mongoose seems to work strange.

mcserep:
Apart from this, the webserver does not seem to serve requests in parallel - at least for Thrift requests, where it was easily observable with the high page load time.

Is there any reason we are not using the built-in multithreaded support of Mongoose? (`MG_ENABLE_THREADS` and `mg_start_thread`)

mcserep:
I close this PR, as there have been no real development on this in the past 3+ years.

This could be reopened if the discussed issues are resolved; or since Mongoose 7.x is the current major release, maybe a new PR could be made in the future, updating the webserver to that version.

PR #393, opened 2020-05-29
Guide and update for Ubuntu 20.04

### LINKED ISSUE #392: [docs] Ubuntu 20.04 LTS build guide
It seems that with #385 being fleshed out more or less, installing CodeCompass on Ubuntu 20.04 isn't that hard. I'll put the relevant differences as a patch once #385 is merged.

whisperity:
> **Note:** Depends on #385.

Closes #392.

 - Tidied up the Travis file, use APT addon, use as many cores as allowed in the VM. (Currently two.)
 - Worked around some backwards-incompatbile minor API breaks of Libgit2 and Boost.
 - Tidied up the user guide.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,19 @@
 os: linux
-sudo: require
-
+dist: bionic
 language: generic
-
 services:
   - postgresql
 
-.apt_xenial: &apt_xenial
+# ---------------------------- APT configuration -----------------------------
+
+.apt_xenial_config: &apt_xenial_config
   update: true
   sources:
     - ubuntu-toolchain-r-test
     - llvm-toolchain-xenial-7
+
+.apt_xenial_sqlite: &apt_xenial_sqlite
+  <<: *apt_xenial_config
   packages:
     # Boost
     - libboost-all-dev
@@ -27,6 +30,37 @@ services:
     # SQLite
     - libodb-sqlite-dev
     - libsqlite3-dev
+    # Parser
+    - libmagic-dev
+    - libgit2-dev
+    - ctags
+    # Service
+    - libgraphviz-dev
+    - nodejs-legacy
+    - npm
+    # GTest
+    - libgtest-dev
+    # Thrift
+    - libssl-dev
+    # Thrift (compile)
+    - byacc
+    - flex
+
+.apt_xenial_pgsql: &apt_xenial_pgsql
+  <<: *apt_xenial_config
+  packages:
+    # Boost
+    - libboost-all-dev
+    # LLVM/clang
+    - llvm-7
+    - llvm-7-dev
+    - clang-7
+    - libclang-7-dev
+    # Java
+    - default-jdk
+    # ODB
+    - odb
+    - libodb-dev
     # PostgreSQL
     - libodb-pgsql-dev
     - postgresql-server-dev-9.5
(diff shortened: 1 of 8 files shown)


whisperity:
TO-DO for now:

 * `libssl-dev` instead of `libssl`**~~`1.0`~~**`-dev`, the dependency issue with Node `npm` is fixed. It was _also_ not apparent in 16.04, so seemingly only 18.04 was "broken" in this regard
 * `gcc-`**`9`**`-plugin-dev`: 20.04 ship with GCC 9 by default
 * Need to explicitly specify we want LLVM **7** for now...
   * `-DLLVM_DIR="/usr/lib/llvm-7/cmake" -DClang_DIR="/usr/lib/cmake/clang-7"` to the command-line options
 * Gitservice's Thrift enum and handling has to be changed because there was an API break in `libgit2-28`: `GIT_OBJ_xxx` is now `GIT_OBJECT_xxx`, **but** in 18.04 (libgit2-26) this is not the case

whisperity:
TO-DO list:
 * `libssl-dev` instead of `libssl`**~~`1.0`~~**`-dev`, the dependency issue with Node `npm` is fixed. It was _also_ not apparent in 16.04, so seemingly only 18.04 was "broken" in this regard
 * `gcc-`**`9`**`-plugin-dev`: 20.04 ship with GCC 9 by default

These will be added once this patch can be truly rebased over #385.

.travis.yml
```diff
@@ -46,20 +47,97 @@ services:
...
+    - clang-7
+    - libclang-7-dev
+    # Java
+    - default-jdk
+    # ODB
+    - gcc-7-plugin-dev
+    # SQLite
+    - libsqlite3-dev
+    # PostgreSQL
+    - libodb-pgsql-dev
```
mcserep, line 67: libodb-pgsql is not needed for Bionic build (manual ODB compilation)

whisperity:
Something's not quite right with this job, Travis seemingly can't spin up a Postgres instance?

~~~
$ travis_setup_postgresql 12

Starting PostgreSQL v12
Assertion failed on job for postgresql@12-main.service.
sudo systemctl start postgresql@12-main
~~~

mcserep:
> Something's not quite right with this job, Travis seemingly can't spin up a Postgres instance?

The Bionic image of Travis comes with postgres 9-11.  
https://docs.travis-ci.com/user/reference/bionic/

For some reason there is no help page for Focal (Travis moves slowly as always), but maybe Postgres 12 is not added, it can be even a bug. Try to add the `postgresql-12` and `postgresql-client-12` to the APT addon, to install them anyway. Maybe it will solve the issue.

whisperity:
> For some reason there is no help page for Focal (Travis moves slowly as always), but maybe Postgres 12 is not added, it can be even a bug. Try to add the `postgresql-12` and `postgresql-client-12` to the APT addon, to install them anyway. Maybe it will solve the issue.

Yeah, found similar suggestions here: https://travis-ci.community/t/services-for-postgresql-11-and-12-fail-to-start-assertion-failed-on-job-for-postgresql-11-main-service/7069/7

Accordingly, even for Postgres 11 the install script is bogus if you simply just call `postgres: "11"`, only up to version 10 works out of the box. I've added version 12 to the 20.04 scripts directly, let's see...

whisperity:
Okay, I managed to get it running. The cornerstone of the issue was that Travis does not spin the Postgres up automatically, and that the authentication defaults for Postgres has changed by the looks of it. The first one was "easily" fixed by installing it from APT, while the second needed a bit of manual tinkering in the configuration file, but now it reliably starts a database you can connect to with an empty password. 😋

plugins/git/service/src/gitservice.cpp
```diff
@@ -60,7 +60,53 @@ int gitDiffToStringCompactCallback(
   return 0;
 }
 
+#define GIT_MACRO_TO_THRIFT_TYPE(GIT_TYPE, THRIFT_TYPE) \
+  case GIT_TYPE: \
+    return cc::service::git::GitObjectType::THRIFT_TYPE;
+
+#if LIBGIT2_VER_MAJOR == 1 || LIBGIT2_VER_MINOR >= 28
```
bruntib, line 67: Wouldn't we like to get prepared for the future by saying `LIBGIT2_VER_MAJOR >= 1`?

Functions.cmake
```diff
@@ -106,7 +106,10 @@ function(find_boost_libraries)
     foreach(_path ${Boost_LIBRARIES})
       if(_path MATCHES ".*boost_${_lib}\.so$")
         list(APPEND LIBS ${_path})
-      endif(_path MATCHES ".*boost_${_lib}\.so$")
+      elseif(_path MATCHES "Boost::${_lib}$")
+        # There is no path for the lib, included as a module.
+        list(APPEND LIBS ${_path})
+      endif()
```
bruntib, line 112: What is the purpose of this `find_boost_libraries()` function? I think we don't need this. We should simply use `Boost_LIBRARIES` instead of `Boost_LINK_LIBRARIES` everywhere.

whisperity: I have no idea. It was like this for some reason on the original code, I just made sure it works for newer Boost and Cmake logic too.

.travis.yml
```diff
@@ -1,16 +1,19 @@
 os: linux
-sudo: require
-
+dist: bionic
```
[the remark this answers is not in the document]
whisperity: All of those we plan to support and target. Which, as it stands now, 16, 18, and 20 LTSes, all of them?

16.04 will be deprecated by Canonical in around half a year from now, then I think we can also reasonably drop it.

18.04 is still officially supported by the vendor for like 4 more years.

We could make an executive decision on saying if you're using CodeCompass then use the newest Ubuntu, but you have to remember there are forks of Ubuntu used internally en masse, which lag behind the official release schedule by years...

Functions.cmake
bruntib: I think if something has no purpose, it shouldn't be in the code. I think we can get rid of this CMake function.

.travis.yml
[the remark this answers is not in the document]
mcserep: While it does not affect me, I think we shall support Ubuntu 16.04, until the ELX release is not upgraded to 18.04.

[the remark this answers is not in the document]
whisperity: > While it does not affect me, I think we shall support Ubuntu 16.04, until the ELX release is not upgraded to 18.04.

ELX is 18.04. But 20.04-based one won't come for another year, if the previous schedule can be believed. So we need to keep targetting 18.04 at least.

doc/deps.md
```diff
@@ -81,6 +95,20 @@ sudo apt-get install libodb-sqlite-dev libsqlite3-dev
 sudo apt-get install libodb-pgsql-dev postgresql-server-dev-<version>
 ```
 
+##### Ubuntu 18.04 ("Bionic Beaver") and 20.04 ("Focal Fossa") LTS
+
+The database connector library must be compiled manually on these systems,
+however, the database programs themselves should be installed from the
+package manager.
```
mcserep, line 102: @bruntib  confirmed that ODB 2.4 works under Ubuntu 20.04, so it's fine to manually compile ODB only on 18.04.

.travis.yml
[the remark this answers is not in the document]
mcserep: @whisperity Ok, that is new info for me. In that case I agree we can drop support for 16.04 soon.
@intjftw This might be relevant for you.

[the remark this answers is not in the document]
whisperity: Actually, I'm just seeing which line this comment refers to. This line is irrelevant, and I only put it back so Travis doesn't show a "build config warning" on this line missing.

The actual distribution versions being used in the jobs are configured for each entry in the `jobs` list.

(`sudo: require` has been deprecated for a while according to the docs, at least for Linux machines.)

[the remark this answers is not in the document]
whisperity: See: Screenshot_20200622_104339

plugins/git/service/git.thrift
```diff
@@ -2,14 +2,14 @@ namespace cpp cc.service.git
...
-  GIT_OBJ_OFS_DELTA = 6,  /**< A delta, base is given by an offset. */
-  GIT_OBJ_REF_DELTA = 7,  /**< A delta, base is given by object id. */
+  Reserved1 = 0,  /**< Reserved for future use in Libgit2. */
+  Commit = 1,     /**< A commit object. */
+  Tree = 2,       /**< A tree (directory listing) object. */
+  Blob = 3,       /**< A file revision object. */
+  Tag = 4,        /**< An annotated tag object. */
+  Reserved2 = 5,  /**< Reserved for future use in Libgit2. */
+  OffetDelta = 6, /**< A delta, base is given by an offset. */
+  RefDelta = 7,   /**< A delta, base is given by object id. */
```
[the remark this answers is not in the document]
whisperity: [Thrift `enum`s do not support negative values.](https://thrift.apache.org/docs/idl#enum)
In addition, these two are not needed. I've delved into libgit2's source:

* `_ANY` is a generic mask, for when you are searching only by the object's ID and you do not know what type it has. I think an application like CodeCompass can pass semantic information that makes this not needed. We should not blindly trust the user or client to start looking stuff up by SHA randomly. In addition, most of this value's usage is in code that handles *creating* Git objects, which CodeCompass does not do.
* `_INVALID` is an error value. Thrift has semantically more advanced (and performance-wise doesn't really matter considering we're JavaScript frontend...) exceptions for this purpose.

Functions.cmake
whisperity: This was introduced in #260 by @mcserep, for 16.04.
I don't have a reasonable multi-boost environment to test this out nicely. If you agree this can be removed, I'll be happy to remove it... considering that linking against `Boost::foobar` most likely breaks the stuff that #260 fixed, either way...

We could introduce in the CMakeList that we need at least Boost *whichever is the stock version for 18.04 at that point* once 16.04 is unsupported? But this CMake module-target (which has nothing to do with C++20 _Modules_ ...) is only apparent with the Boost version supplied for 20.04. 18.04 still uses the "old" code.

whisperity: Another thing this function is used for is to set that only the necessary Boost libs for each end binary is linked, not *everything* from Boost that the *whole* CodeCompass needs. I.e. the `util` library does not need `boost_thread` or `boost_program_options`. If we remove this function and tell every boost-needing end binary to "link against Boost" then the project-wide `Boost COMPONENTS` list will end up being used...

mcserep: The issue was that I have multiple Boost versions installed on my development computer and had the compilation problem described in #260. I will check again, but I think I was already using Ubuntu 18.04 at that time.

mcserep:
Are there any more unresolved issues with this PR? @bruntib ?
If not, we shall get it merged!

Functions.cmake
mcserep: I did not manage to recreate this issue in the end (guess something changed in the last 2 years affecting the source of the problem), so we can remove this `find_boost_libraries()` function, as already proposed by #398 .

PR #395, opened 2020-05-29
Use the APT addon for Bionic in Travis

mcserep:
At the time Bionic (18.04) build was added to the CI, Travis did not support the APT addon for the Bionic image, so manual installation of packages were required (see d3ee99663879056a2a82dd1f58558bc5458dc4e7).

That issue was solved since then, so we shall use the APT addon of Travis, in a similar way like with the Xenial (16.04) build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,20 +46,38 @@ services:
     - byacc
     - flex
 
-.install_bionic: &install_bionic
-  >
-  sudo apt-get install -y
-  libboost-all-dev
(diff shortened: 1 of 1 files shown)


whisperity:
I've already taken care of moving to the "apt addon" in `travis.yml` in #393 😉

mcserep:
> I've already taken care of moving to the "apt addon" in `travis.yml` in #393

Good, I haven't noticed. I will close this PR then.

PR #198, opened 2017-07-06
Symbol clustering

whisperity:
> **NOTICE!** Makes use of changes introduced in #120 and thus **CAN NOT** be merged until #120 is accepted!

Introduce the symbol clustering algorithm and improvements to CodeCompass. This change makes CodeCompass use the linkage information available in the `build.json` (if present), to store linkage data to certain translation units in a quick and optimalised manner.

This data is used when the user is searching between symbols (such as _Jump to Definition_ or _List Callees_ queries) to generate the list of symbols that are statically linked to the "search origin" (e.g. the `f()` call site). These results are highlighted (while those that are not statically link are highlighted in another way).

With this change, if a project contains ambiguous mangled names, the user can easier select a definition that is _supposedly_ (!) more relevant to the point where he/she is searching from, than other definitions matching based on the function signature.

This change produces minimal overhead in the database and has a minimal impact on parse and navigation time. The algorithm and the results has been discussed in detail on the **33rd Conference of Scientific Students' Associations** (XXXIII. OTDK), and on the **17th IEEE International Working Conference on Source Code Analysis and Manipulation** (SCAM-2017).

PR #398, opened 2020-06-20
Docker image upgrade to Ubuntu 20.04

bruntib:
The development Docker image has been upgraded to use Ubuntu 20.04.
This Ubuntu version contains ODB and Thrift in its package manager, so there is no need to build these tools from source code.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,74 +1,47 @@
-# CodeCompass development dependencies
-FROM ubuntu:bionic
+FROM ubuntu:20.04
 
+# tzdata package is installed implicitly in the following command. This package
+# sets timezone interactively during the installation process. This environment
+# variable prevents this interaction.
+ARG DEBIAN_FRONTEND=noninteractive
+
+# CodeCompass development dependencies.
 RUN set -x && apt-get update -qq \
-    && apt-get -y install \
-    cmake make \
-    curl ctags \
-    default-jdk \
-    gcc-7 gcc-7-plugin-dev \
-    libboost-filesystem-dev \
-    libboost-log-dev \
-    libboost-program-options-dev \
-    libboost-regex-dev \
-    libgit2-dev \
-    libgraphviz-dev \
-    libgtest-dev \
-    libmagic-dev \
-    libsqlite3-dev \
-    libssl1.0-dev \
-    llvm-7 clang-7 llvm-7-dev libclang-7-dev \
-    nodejs-dev node-gyp npm
+  && apt-get -y install --no-install-recommends \
+  cmake make \
+  default-jdk \
+  ctags \
+  gcc-9 gcc-9-plugin-dev g++-9 \
+  libboost-filesystem-dev \
+  libboost-log-dev \
+  libboost-program-options-dev \
+  libboost-regex-dev \
+  libgit2-dev \
+  libgraphviz-dev \
+  libgtest-dev \
+  libmagic-dev \
+  libsqlite3-dev \
+  libssl-dev \
+  llvm-7 clang-7 llvm-7-dev libclang-7-dev \
+  npm \
+  thrift-compiler libthrift-dev \
+  odb libodb-sqlite-dev && \
+  ln -s /usr/bin/gcc-9 /usr/bin/gcc && \
+  ln -s /usr/bin/g++-9 /usr/bin/g++
 
-# Download the Thrift source code.
-ADD http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.13.0/thrift-0.13.0.tar.gz /opt/thrift-0.13.0.tar.gz
-# Build Thrift
-RUN tar -xf /opt/thrift-0.13.0.tar.gz -C /opt; \
-    cd /opt/thrift-0.13.0 && \
-    ./configure --prefix=/opt/thrift_install \
-    --silent --without-python                                        \
-    --enable-libtool-lock --enable-tutorial=no --enable-tests=no     \
-    --with-libevent --with-zlib --without-nodejs --without-lua       \
-    --without-ruby --without-csharp --without-erlang --without-perl  \
-    --without-php --without-php_extension --without-dart             \
-    --without-haskell --without-go --without-rs --without-haxe       \
-    --without-dotnetcore --without-d --without-qt4 --without-qt5;    \
-    cd /; \
-    make -C /opt/thrift-0.13.0 -j $(nproc) --silent install; \
-    rm -rf /opt/thrift-0.13.0
 # Build GTest.
-RUN cd /usr/src/googletest/ && \
-     mkdir build && \
-     cd build && \
-     cmake .. && \
-     make install && \
-     cd / && \
-     rm -rf /usr/src/googletest/build
+RUN cd /usr/src/googletest && \
+  mkdir build && \
+  cd build && \
+  cmake .. && \
+  make install && \
+  cd / && \
+  rm -rf /usr/src/googletest/build
 
-# Download build2 toolchain install script.
-ADD https://download.build2.org/0.12.0/build2-install-0.12.0.sh /opt/build2-install-0.12.0.sh
-# Build build2, then ODB.
-RUN cd /opt/ && \
-    sh build2-install-0.12.0.sh --yes --trust yes; \
-    mkdir /opt/odb_build && \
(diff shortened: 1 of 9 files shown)


bruntib:
I've just seen that there are some pull requests which target Ubuntu upgrade. I'll check those and merge the necessary things accordingly.

whisperity:
> This Ubuntu version contains ODB and Thrift in its package manager, so
> there is no need to build these tools from source code.


ODB in 20.04 is still the old 2.4.0 version: https://packages.ubuntu.com/focal/devel/odb

I didn't check if using this version _works_ (i.e. doesn't produce the GCC invocation issues with `-mtune`), but this would create a roller-coaster where 18.04 needs to use newer ODB, whereas 16.04 **and 20.04** can use the old one?

bruntib:
I did check it, at least in Docker. This Docker image which I introduced here works properly. I think ODB version doesn't really matter for us as long as it works.

util/include/util/hash.h
```diff
@@ -5,7 +5,7 @@
 #include <string>
 #include <sstream>
 
-#include <boost/uuid/sha1.hpp>
+#include <boost/uuid/detail/sha1.hpp>
```
mcserep, line 8: CodeCompass does not compile with this under Xenial and Ubuntu, according to the CI.

bruntib: Yes, I created this pull request before I noticed that #393 also contains similar improvements. Maybe we could consider that PR as a precondition of this one. I'll rebase it after that one is merged.

whisperity:
> I did check it, at least in Docker. This Docker image which I introduced here works properly. I think ODB version doesn't really matter for us as long as it works.

@bruntib True only until #198 gets merged, but considering it's been 3 years, I've given up hope it ever will.

filbeofITK:
> 
> 
> > This Ubuntu version contains ODB and Thrift in its package manager, so
> > there is no need to build these tools from source code.
> 
> ODB in 20.04 is still the old 2.4.0 version: https://packages.ubuntu.com/focal/devel/odb
> 
> I didn't check if using this version _works_ (i.e. doesn't produce the GCC invocation issues with `-mtune`), but this would create a roller-coaster where 18.04 needs to use newer ODB, whereas 16.04 **and 20.04** can use the old one?

The 2.4 version works on Focal Fossa because it was `gcc-7` that broke it, and 20.04 ships with `gcc-9`. Thats why it worked on Xenial too, because that was using `gcc-5` by default.

util/include/util/hash.h
mcserep: Isn't this issue related to the Boost version instead? The path `boost/uuid/detail/sha1.hpp` available since version 1.68 as I have checked. Under Ubuntu 18.04 the version 1.65 is installed from the APT repository.

plugins/metrics/parser/CMakeLists.txt
```diff
@@ -8,13 +8,10 @@ include_directories(
...
   src/metricsparser.cpp)
 
-find_boost_libraries(
-  filesystem
-  log)
 target_link_libraries(metricsparser
   metricsmodel
   util
-  ${Boost_LINK_LIBRARIES})
+  ${Boost_LIBRARIES})
```
whisperity, line 14: Are we sure it's okay to link *every* CodeCompass lib that uses *some* boost libraries against **every Boost lib ever** used by *any* CodeCompass target?

Because right now, this will make `libmetricsparser.so` depend on every boost lib we use:
https://github.com/Ericsson/CodeCompass/blob/master/CMakeLists.txt#L13

As opposed to just depending on `libboostlog` and `libboostfilesystem`.

bruntib: I played around with this Boost library issue. I noticed that even in the current master version, too many Boost libraries are imported, because those are dependencies of other Boost libraries. For example if you check what libs are linked agains metrics parser, you can see that it gets "atomic", "thread", "date_time" and many other Boost libs despite its CMakeLists.txt contains only "filesystem" and "log".
So as far as I can see, this new solution is not worse than the previous one.

bruntib:
@mcserep Could you please check if you still get a compilation error?

mcserep:
> @mcserep Could you please check if you still get a compilation error?

I have already evaluated this when reviewing #393 , and I did not manage to recreate this issue in the end (guess something changed in the last 2 years affecting the source of the problem).

docker/dev/Dockerfile
```diff
@@ -77,6 +47,5 @@ ENV DATABASE=sqlite \
     TEST_WORKSPACE=/CodeCompass/test_workspace \
     TEST_DB="sqlite:database=$TEST_WORKSPACE/cc_test.sqlite"
 
-ENV PATH="$INSTALL_DIR/bin:/opt/thrift_install/bin:/opt/odb_install/bin:$PATH"
-ENV CMAKE_PREFIX_PATH="/opt/thrift_install:/opt/odb_install:$CMAKE_PREFIX_PATH"
+ENV PATH="$INSTALL_DIR/bin:$PATH"
 ENV LD_LIBRARY_PATH="$INSTALL_DIR/lib:$LD_LIBRARY_PATH"
```
whisperity, line 51: I think this line is not needed, because "installdir/lib" is automatically set as search root by the server, and plugins are built with RPATH.

https://github.com/Ericsson/CodeCompass/blob/3f413ed284ced641a00de64a8b9ef41232b8515f/webserver/src/webserver.cpp#L51

https://github.com/Ericsson/CodeCompass/blob/3f413ed284ced641a00de64a8b9ef41232b8515f/Config.cmake#L48-L61

docker/dev/Dockerfile
```diff
@@ -1,74 +1,44 @@
...
+  libgit2-dev \
+  libgraphviz-dev \
+  libgtest-dev \
+  libmagic-dev \
+  libsqlite3-dev \
+  libssl-dev \
+  llvm-7 clang-7 llvm-7-dev libclang-7-dev \
+  npm \
+  thrift-compiler libthrift-dev \
+  odb
```
whisperity, line 27: One minor nit w.r.t. image size: simply installing `odb` will install **all** odb connectors, even mysql and postgres and such, something we're not using inside the image.

plugins/metrics/parser/CMakeLists.txt
whisperity: Right, on the new system and new CMake, imported targets do this. Anyways... we can come back to this later when it comes to one plugin depending alone on a "new" boost lib in particular.

util/include/util/hash.h
whisperity: Fixed by rebase over #393

whisperity:
@mcserep Any idea on how we could CI these `Dockerfile`s? At least to the point that they *build*?

mcserep:
> @mcserep Any idea on how we could CI these `Dockerfile`s? At least to the point that they _build_?

@whisperity We simply need to add the docker service to the YAML configuration file:
```
services:
  - docker
```
Then we can use the `docker` CLI. Further info: https://docs.travis-ci.com/user/docker/

docker/web/Dockerfile
```diff
@@ -27,26 +27,33 @@ RUN mkdir /CodeCompass-build && \
     -DDATABASE=$CC_DATABASE \
     -DCMAKE_INSTALL_PREFIX=/CodeCompass-install \
     -DCMAKE_BUILD_TYPE=$CC_BUILD_TYPE && \
-  make && \
+  make -j8 && \
```
mcserep, line 30:
```suggestion
  make -j $(nproc) && \
```

docker/dev/Dockerfile
whisperity: This comment is still valid.

mcserep:
> I think the `web` image should not spend time and disk space creating (and downloading the dependencies of) the `parser` binary and tools, but this is an issue we can resolve later.

@whisperity  I think the `web` image should be renamed to e.g. `runtime`. In a real-life situation I think one of the following two scenarios can be expected:
1. One parses and then serves projects on the host machine, where CodeCompass can be built.
2. One parses and then serves projects in Docker containers, so CodeCompass is not required to be built on the host machine.

It is highly unlikely that someone would like to use the CodeCompass *parser* on the host machine, then the *webserver* in a docker container. Therefore the docker image shall contain the runtime for both in my opinion. (Or we can create separate images for them.)

But this shall be a different PR.

Issue #124, opened 2017-03-08
Plugins should be able to supply their own CSS


whisperity:
Currently it looks like stylesheet for the plugins are stored in the common `codecompass.css`.

whisperity:
Duplicate of #22.

PR #402, opened 2020-07-17
Use the Dojo Build System

mcserep:
Currently loading the front page of the CodeCompass web interface requires 450-500 HTTP requests, most of them JavaScript resources, as the Dojo library dynamically loads the modules.

This PR utilizes the [Dojo Build System](https://dojotoolkit.org/documentation/tutorials/1.10/build/) to create a custom build for the WebGUI. The build system collects all the required Dojo modules (also traversing their dependencies) and bundle them into a single JavaScript file, so the number of HTTP requests can be drastically reduced. These bundles are written into the `share/codecompass/webgui/scripts/release/` directory of the installation path.
The Dojo Build System for CodeCompass was configured to have 3 bundles:
1. Bundle for Core dojo modules.
2. Bundle of custom Dojo-compliant JavaScript files and their dependencies of Dojo modules.
3. Bundle of custom Dojo-compliant JavaScript files for the login page and their dependencies of Dojo modules.

When CodeCompass is compiled in **Release** mode, the [*Google Closure Compiler*](https://developers.google.com/closure/compiler) is also utilized for minification through comments and whitespace removal, and also dead-code removal.
The Dojo Build System is utilized in the installation phase of CodeCompass and have the following time footprint:
- in *Debug* mode an extra 4-5 sec;
- in *Release mode an extra 10-15 sec (Google Closure Compiler also has to be run).

As a result only ca. 100 HTTP requests are done on the load of the front page. About 50 of them are JavaScript files, but this can hardly be further reduced, as other 3rd party libraries like CodeMirror or the Thrift generated JS files are not loaded through Dojo.

**The page load time for me was reduced from 2.7-3.2 sec to 0.7-1.2 sec, which is around 60% of page load reduction.** Therefore I hope this PR will radically reduce the page load time on weaker machines where it was around 20-30 sec previously.

### DIFF
--- a/service/plugin/include/pluginservice/pluginservice.h
+++ b/service/plugin/include/pluginservice/pluginservice.h
@@ -22,6 +22,8 @@ public:
 
   void getPlugins(std::vector<std::string>& return_) override;
 
+  void getThriftPlugins(std::vector<std::string> & _return) override;
+
   void getWebPlugins(std::vector<std::string> & _return) override;
 
   void getWebStylePlugins(std::vector<std::string> & _return) override;
--- a/service/plugin/src/pluginservice.cpp
+++ b/service/plugin/src/pluginservice.cpp
@@ -1,4 +1,5 @@
 #include <pluginservice/pluginservice.h>
+#include <boost/regex.hpp>
 
 namespace
 {
@@ -37,20 +38,36 @@ void PluginServiceHandler::getPlugins(std::vector<std::string>& return_)
     return_.push_back(it.first);
 }
 
-void PluginServiceHandler::getWebPlugins(std::vector<std::string>& return_)
+void PluginServiceHandler::getThriftPlugins(std::vector<std::string>& return_)
 {
   std::string webrootDir = _configuration["webguiDir"].as<std::string>();
 
   std::vector<std::string> generatedFiles = getFiles(
     webrootDir + "/scripts/codecompass/generated");
 
-  std::vector<std::string> jsModules = getFiles(
-    webrootDir + "/scripts/codecompass/view");
-
   for (const std::string& file_ : generatedFiles)
     return_.push_back(file_.substr(webrootDir.size() + 1)); // +1 to remove starting slash
+}
+
+void PluginServiceHandler::getWebPlugins(std::vector<std::string>& return_)
+{
+  std::string webrootDir = _configuration["webguiDir"].as<std::string>();
+
+  std::vector<std::string> jsModules = getFiles(
+    webrootDir + "/scripts/release/codecompass/view");
+
+  /*
+   * Pattern to match only *.js files which does not contain the '.js.' substring, to exclude:
+   * *.js.uncompressed.js
+   * *.js.consoleStripped.js
+   * *.js.map
+   */
+  static const boost::regex pattern("^((?!\\.js\\.).)*\\.js$");
   for (const std::string& file_ : jsModules)
-    return_.push_back(file_.substr(webrootDir.size() + 1)); // +1 to remove starting slash
+  {
+    if (boost::regex_match(file_, pattern))
+      return_.push_back(file_.substr(webrootDir.size() + 1)); // +1 to remove starting slash
+  }
 }
 
 void PluginServiceHandler::getWebStylePlugins(std::vector<std::string>& return_)
--- a/webgui/InstallGUI.cmake
+++ b/webgui/InstallGUI.cmake
@@ -2,7 +2,41 @@ message("Install npm packages...")
 
 execute_process(
   COMMAND cp ${CC_PACKAGE} ${INSTALL_SCRIPTS_DIR}/package.json
+  WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
+
+execute_process(
   COMMAND npm install
   WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
 
+execute_process(
+  COMMAND ${CMAKE_COMMAND} -E create_symlink dojo-util node_modules/util
+  WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
+
 message("Installation of npm packages are finished.")
+
+message("Initiating the Dojo Build System...")
+
+# Collect CodeCompass view JavaScript files
+set(DOJO_VIEWLIST "")
+file(GLOB _jsfiles RELATIVE "${INSTALL_SCRIPTS_DIR}" "${INSTALL_SCRIPTS_DIR}/codecompass/view/[^A-Z]*.js")
+foreach(_jsfile ${_jsfiles})
+  string(REGEX REPLACE "\\.js*$" "" _jsfile ${_jsfile})
+  list(APPEND DOJO_VIEWLIST "'${_jsfile}'")
+endforeach()
+list(JOIN DOJO_VIEWLIST  ", " DOJO_VIEWLIST)
+
+set(DOJO_OPTIMIZE "")       # optimization for non-layer modules
+set(DOJO_LAYEROPTIMIZE "")  # optimization for layer modules
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
+  set(DOJO_LAYEROPTIMIZE "closure")
+endif()
+configure_file(${CC_PROFILE} ${INSTALL_SCRIPTS_DIR}/codecompass.profile.js)
+
+execute_process(
+  COMMAND node_modules/util/buildscripts/build.sh --profile codecompass.profile.js
+  WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
+
+# The build-report may leak sensitive information
+file(REMOVE ${INSTALL_SCRIPTS_DIR}/release/build-report.txt)
+
+message("The Dojo Build System finished.")
--- a/webgui/index.html
+++ b/webgui/index.html
@@ -12,12 +12,12 @@
     <link rel="stylesheet" href="scripts/node_modules/codemirror/lib/codemirror.css" />
     <link rel="stylesheet" href="scripts/node_modules/codemirror/addon/dialog/dialog.css" />
     <link rel="stylesheet" href="scripts/node_modules/codemirror/addon/fold/foldgutter.css" />
-    <link rel="stylesheet" href="scripts/node_modules/dojo/resources/dojo.css" />
-    <link rel="stylesheet" href="scripts/node_modules/dijit/themes/claro/claro.css"/>
-    <link rel="stylesheet" href="scripts/node_modules/dojox/layout/resources/ResizeHandle.css" />
-    <link rel="stylesheet" href="scripts/node_modules/dojox/layout/resources/FloatingPane.css" />
-    <link rel="stylesheet" href="scripts/node_modules/dojox/image/resources/Lightbox.css" />
-    <link rel="stylesheet" href="scripts/node_modules/dojox/grid/resources/claroGrid.css" />
+    <link rel="stylesheet" href="scripts/release/dojo/resources/dojo.css" />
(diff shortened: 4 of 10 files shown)


webgui/index.html
```diff
@@ -44,24 +44,29 @@
...
+      }
+
       var dojoConfig = {
-        baseUrl: 'scripts/',
+        baseUrl: 'scripts/release/',
         tlmSiblingOfDojo: false,
         defaultDuration: 1,
         async: true,
-        cacheBust: true,
+        cacheBust: Cookies.get('dojo-cache'),
```
mcserep, line 59: I have also addressed the caching issue.

service/plugin/include/pluginservice/pluginservice.h
```diff
@@ -22,6 +22,8 @@ class PluginServiceHandler : virtual public PluginServiceIf
 
   void getPlugins(std::vector<std::string>& return_) override;
 
+  void getThriftPlugins(std::vector<std::string> & _return) override;
+
```
mcserep, line 26: This separation was required, as the thrift generated JS resources cannot be loaded with the Dojo bundle, while the custom JavaScript files are Dojo-compliant. Therefore it was very inconvenient that the `getWebPlugins()` method returned both collection of files.

whisperity:
I haven't tried this out yet. I do like the general idea. One question: what happens with the `Release` build on a system that either has no internet access, or accessing *Google* is forbidden? Will the build still work, just by skipping the "closure compiler"? Or will it break?

mcserep:
> I haven't tried this out yet. I do like the general idea. One question: what happens with the `Release` build on a system that either has no internet access, or accessing _Google_ is forbidden? Will the build still work, just by skipping the "closure compiler"? Or will it break?

It works offline.

whisperity:
Roughly 3 sec load time in release mode, and 6 secs in debug mode. :heart_eyes: For such a "simple" change, this is amazing for the cold boot.

service/plugin/src/pluginservice.cpp
```diff
@@ -37,20 +38,30 @@ void PluginServiceHandler::getPlugins(std::vector<std::string>& return_)
...
+}
+
+void PluginServiceHandler::getWebPlugins(std::vector<std::string>& return_)
+{
+  std::string webrootDir = _configuration["webguiDir"].as<std::string>();
+
+  std::vector<std::string> jsModules = getFiles(
+    webrootDir + "/scripts/release/codecompass/view");
+
+  const boost::regex pattern("^((?!\\.js\\.).)*\\.js$"); // only *.js files which does not contain the '.js.' substring
```
whisperity, line 59: If Boost.Regex supports `constexpr`, could you try:

```cpp
static constexpr boost::regex patttern(...)
```

If not, even simply doing `static const` is enough. This constructor should not be executed every time this function runs.

I would even move this to a file-scope static.

webgui/app.profile.js
```diff
@@ -0,0 +1,111 @@
...
+                'codecompass/view/cppReparseFileAST',
+                'codecompass/view/cppReparseMenu',
+                'codecompass/view/gitBlame',
+                'codecompass/view/gitCommitView',
+                'codecompass/view/gitMenu',
+                'codecompass/view/gitNavigator',
+                'codecompass/view/gitUtil',
+                'codecompass/view/metrics',
+                'codecompass/view/searchFields',
+                'codecompass/view/searchResult'
```
whisperity, line 101: I am conflicted about hardcoding the list of files of **plugins** in something that is **outside** the `plugin` directory. This change is reversing the efforts we made to make the architecture pluginable.

Plugins should work in a way that if I have, let's say, a Lua plugin, I just throw it into `plugins/` and build CodeCompass with it present, and it works...
This is an additional configuration step that complicates matters...

I know about #124 still not done, but we shouldn't *introduce* a similar problem.

CMake could be able to generate this particular file at build time from a template, putting the concrete list of files in from glob expressions.

webgui/InstallGUI.cmake
```diff
@@ -2,7 +2,31 @@ message("Install npm packages...")
...
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "closure")
+else()
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "")
+endif()
+configure_file(${CC_PROFILE} ${INSTALL_SCRIPTS_DIR}/app.profile.js)
+
+execute_process(
+  COMMAND node_modules/util/buildscripts/build.sh --profile app.profile.js
```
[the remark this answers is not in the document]
mcserep: It is installed with the *dojo-util* npm package.

webgui/InstallGUI.cmake
```diff
@@ -2,7 +2,31 @@ message("Install npm packages...")
...
+message("Initiating the Dojo Build System...")
+
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "closure")
+else()
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "")
+endif()
+configure_file(${CC_PROFILE} ${INSTALL_SCRIPTS_DIR}/app.profile.js)
```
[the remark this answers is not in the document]
mcserep: The `app.profile.js` is a widely used name (I have checked multiple projects for reference), but we can rename it. Until we have multiple profiles (if we will ever need), it does not really matter.

webgui/InstallGUI.cmake
```diff
@@ -2,7 +2,31 @@ message("Install npm packages...")
...
+
+message("Initiating the Dojo Build System...")
+
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "closure")
+else()
+  set(DOJO_OPTIMIZE "")
+  set(DOJO_LAYEROPTIMIZE "")
+endif()
```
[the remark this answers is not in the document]
mcserep: That is the optimization parameter for the non-layered files. Setting it to `closure` will add an extra ~60 sec to installation time, while the runtime performance will not be improved noticeably, since the JS resources are well bundled (layered) in our case. So I left it disabled also in the release mode.

service/plugin/src/pluginservice.cpp
mcserep: @bruntib The release directory will contain other JS outputs, e.g `*.js.uncompressed.js` for the non-minified versions, which can be useful for debugging. However we **must** not include those.

webgui/InstallGUI.cmake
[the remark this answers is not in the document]
mcserep: We can completly remove the `DOJO_OPTIMIZE` variable, if it is confusing.

webgui/app.profile.js
mcserep: Yes, I agree. I will try to introduce some CMake magic here :)

webgui/InstallGUI.cmake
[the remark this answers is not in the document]
mcserep: For the recoed: the `util` directory is a symlink created during the installation phase of CodeCompass, because the Dojo Build System expects it to be installed there (it is hard-coded). However, *npm* does not support for a package to be in a directory other than its name. Hence the symlink. (Dojo therefore recommends *bower* to be used, where it is supported, but I did not want to change that much.)

webgui/InstallGUI.cmake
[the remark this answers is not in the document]
whisperity: CMake `set()` calls can have a `DESCRIPTION`... maybe it's worth to leave this variable in (but define *outside* the `if` to empty-string) and put a description to both `_OPTIMIZE` and `_LAYEROPTIMIZE`?

service/plugin/src/pluginservice.cpp
bruntib: In this case we should rather match `.uncompressed.js` postfix. Regex is not even needed.

mcserep: There are other possible extensions, e.g. `.js.strippedConsole.js`. While we could list and match all of them (there are like 4-5 as I remember), I think this the future-proof solution.

webgui/InstallGUI.cmake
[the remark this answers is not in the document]
mcserep: Renamed.

webgui/InstallGUI.cmake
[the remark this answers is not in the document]
mcserep: Modified accordingly. (I have not found a `DESCRIPTION` option without making the variable cached, but have added a comment to the code.)

webgui/app.profile.js
mcserep: Done it based on the `install_webplugin` function.

webgui/InstallGUI.cmake
```diff
@@ -2,7 +2,38 @@ message("Install npm packages...")
...
+set(DOJO_VIEWLIST "")
+file(GLOB _jsfiles RELATIVE "${INSTALL_SCRIPTS_DIR}" "${INSTALL_SCRIPTS_DIR}/codecompass/view/[^A-Z]*.js")
+foreach(_jsfile ${_jsfiles})
+  string(REGEX REPLACE "\\.js*$" "" _jsfile ${_jsfile})
+  list(APPEND DOJO_VIEWLIST "'${_jsfile}'")
+endforeach()
+list(JOIN DOJO_VIEWLIST  ", " DOJO_VIEWLIST)
+
+set(DOJO_OPTIMIZE "")       # optimization for non-layer modules
+set(DOJO_LAYEROPTIMIZE "")  # optimization for layer modules
```
whisperity, line 29: I meant the comment as in not a code comment but a descriptive comment for the variable itself, saved by CMake.

```suggestion
set(DOJO_OPTIMIZE "" CACHE STRING "Optimisation method for non-layer Dojo modules")
set(DOJO_LAYEROPTIMIZE "" CACHE STRING "Optimisation method for layer-built Dojo modules")
```

mcserep: While providing the description in a fancy way, this will make these variables cached by CMake and configurable by the user e.g. through `cmake-gui`. These variables are not for the user to configure, they are for internal usage only.

whisperity: I missed that, sorry.

mcserep:
Any further reviews or can this be merged?

service/plugin/src/pluginservice.cpp
```diff
@@ -37,20 +38,30 @@ void PluginServiceHandler::getPlugins(std::vector<std::string>& return_)
...
+}
+
+void PluginServiceHandler::getWebPlugins(std::vector<std::string>& return_)
+{
+  std::string webrootDir = _configuration["webguiDir"].as<std::string>();
+
+  std::vector<std::string> jsModules = getFiles(
+    webrootDir + "/scripts/release/codecompass/view");
+
+  static const boost::regex pattern("^((?!\\.js\\.).)*\\.js$"); // only *.js files which does not contain the '.js.' substring
```
mcserep, line 59: Add list of excluded extensions.

whisperity:
There is a file generated under `INSTALL_DIR/share/codecompass/webgui/scripts/release` named `build-report.txt` which contains information about the paths' used during the build. This may leak username and additional details.

I think this file should be removed by CMake when doing a Release build...

mcserep:
> There is a file generated under `INSTALL_DIR/share/codecompass/webgui/scripts/release` named `build-report.txt` which contains information about the paths' used during the build. This may leak username and additional details.
> 
> I think this file should be removed by CMake when doing a Release build...

@whisperity Thanks, fixed.

bruntib:
Thanks for this patch, great work!

PR #403, opened 2020-07-19
Adding C++ compiler diagnostic messages to the database

bruntib:
`clang::DiagnosticConsumer` is handling the diagnostic messages (compilation errors, warnings, notes, etc.) during C/C++ compilation. These diagnostic messages are collected and stored in the database.

### DIFF
--- a/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
+++ b/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
@@ -18,7 +18,6 @@ void DiagnosticMessageHandler::HandleDiagnostic(
   clang::DiagnosticsEngine::Level diagLevel_,
   const clang::Diagnostic& info_)
 {
-  // TODO: Is this needed?
   clang::DiagnosticConsumer::HandleDiagnostic(diagLevel_, info_);
 
   model::BuildLog buildLog;
--- a/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
+++ b/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
@@ -1,0 +1,76 @@
+#include <llvm/ADT/SmallString.h>
+#include <cppparser/filelocutil.h>
+#include "diagnosticmessagehandler.h"
+
+namespace cc
+{
+namespace parser
+{
+
+DiagnosticMessageHandler::DiagnosticMessageHandler(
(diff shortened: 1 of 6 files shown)


plugins/cpp/parser/src/diagnosticmessagehandler.cpp
```diff
@@ -0,0 +1,76 @@
...
+    case clang::DiagnosticsEngine::Fatal:
+      buildLog.log.type = model::BuildLogMessage::FatalError;
+      break;
+  }
+
+  //--- Message content ---//
+
+  llvm::SmallString<512> content;
+  info_.FormatDiagnostic(content);
+  buildLog.log.message = content.c_str();
```
whisperity, line 48: Isn't this a lifetime violation? `content` dies but `buildLog` is saved in a data structure...

bruntib: `buildLog.log.message` is an `std::string` so a copy is made here.

whisperity: Okay... or rather, not great but not terrible...

plugins/cpp/parser/src/diagnosticmessagehandler.cpp
```diff
@@ -0,0 +1,76 @@
...
+    : _srcMgr(srcMgr_), _db(db_)
+{
+}
+
+void DiagnosticMessageHandler::HandleDiagnostic(
+  clang::DiagnosticsEngine::Level diagLevel_,
+  const clang::Diagnostic& info_)
+{
+  // TODO: Is this needed?
+  clang::DiagnosticConsumer::HandleDiagnostic(diagLevel_, info_);
```
mcserep, line 22: This only increments the `numWarnings` or `numErrors` counters. We do not use them now, so technically it is not needed, but later it could be useful, so in my opinion we shall keep it and remove the TODO.

whisperity:
>
> Am I right, that these logged messages are not available yet from the web interface? There is a `getBuildLog` endpoint in `project.thrift`, but I did found any call to it.

@mcserep Yes, the similar feature was never wired in. The JS code that builds the log table UI elements is there, but that function is also nowhere called. That function would be called with the records populated by the API call.

PR #406, opened 2020-07-21
I read throught the docs, and fixed some typos.

filbeofITK:
Also I modified some sentences, because I found the original ones
hard to understand for someone that reads the instructions for the first time.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -32,7 +32,7 @@ psql
 In the PostgreSQL command-line shell, type:
 
 ```sql
-CREATE USER compass WITH SUPERUSER LOGIN PASSWORD <mypassword>;
+CREATE USER compass WITH SUPERUSER LOGIN PASSWORD '<mypassword>';
 ```
 
 You can exit this shell by typing `\q` and pressing the `ENTER` key. A user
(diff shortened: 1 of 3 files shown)


doc/usage.md
```diff
@@ -32,7 +32,7 @@ psql
 In the PostgreSQL command-line shell, type:
 
 ```sql
-CREATE USER compass WITH SUPERUSER LOGIN PASSWORD 'mypassword';
+CREATE USER compass WITH SUPERUSER LOGIN PASSWORD <mypassword>;
```
whisperity, line 35: Are you sure this will work? The documentation specifies that passwords are strings, and as such, shall be in `'`s: https://www.postgresql.org/docs/current/sql-createrole.html

doc/usage.md
```diff
@@ -195,7 +194,7 @@ CodeCompass_webserver -w <workdir> -p <port>
   connection string the user can give the location of a running database
   system. In the parsing session the database name could have been provided.
   This database name is written in a config file in the workspace directory, so
-  it is useless to provide it at server start.
```
whisperity, line 198: Remove the entire "Database" element from this list instead, it's no longer present or used.

filbeofITK: That is good with me I just didn't want to remove someones work without a word.

doc/usage.md
filbeofITK: Well I changed that one for consistency, because everywhere else we use <string>, but I suppose we could use '<string>' here?

doc/usage.md
whisperity: I believe it's actually an *error* to provide `-d` for the server binary, as that thing is no longer in the `boost::po` as a key.

doc/usage.md
whisperity: Everywhere else, we specify *variables*, and those variables are paths which rarely (on UNIX) contain spaces and unambiguous with regards to capitalisation.
However, I am not sure if postgres even **supports** specifying this thing *wihout* the quotes, i.e. will its parser know we are giving a string option, and not something else.

PostgreSQL's string vs. entityname handling is really whack, *especially* in the `psql` shell.

PR #407, opened 2020-07-22
Link authentication guide into main usage guide

mcserep:
There is a separate authentication guide in the `doc` folder, but it is not linked in the main usage guide (or basically anywhere), therefore it is hard to find and is easily looked over.


whisperity:
How did I miss this... Thanks!

PR #408, opened 2020-07-23
Modified the file manager menu to contain a menu item for git blame.

### LINKED ISSUE #404: Git "blame" action should be available from the file menu
Calling blame is a file-level and not a symbol-level action, and as such, "Git -> Blame" should be available in the right click menu for files in the file browser, not just in the right click menu for source code.

intjftw:
Added a new menu item to `gitMenu.js` that is identical to the existing one, except the new one appears in the File Manager menu when right clicking a file. Directories do not have this menu item.
Closes #404

### DIFF
--- a/plugins/git/webgui/js/gitMenu.js
+++ b/plugins/git/webgui/js/gitMenu.js
@@ -9,9 +9,39 @@ function (topic, Menu, MenuItem, PopupMenuItem, model, viewHandler) {
 
   model.addService('gitservice', 'GitService', GitServiceClient);
 
-  var team = {
-    id : 'git-text-team',
+  var nodeMenu = {
+    id : 'git-text-team-node',
     render : function (nodeInfo, fileInfo) {
+      var submenu = new Menu();
+
+      submenu.addChild(new MenuItem({
+        label : 'Blame',
+        onClick : function () {
+          topic.publish('codecompass/gitblame', {
+            fileInfo : fileInfo,
+            nodeInfo : nodeInfo
+          });
+        }
+      }));
+
+      return new PopupMenuItem({
+        label : 'Team',
+        popup : submenu
+      });
+    }
+  };
+
+  viewHandler.registerModule(nodeMenu, {
+    type     : viewHandler.moduleType.TextContextMenu,
+    priority : 70
+  });
+
+  var fileMenu = {
+    id : 'git-text-team-file',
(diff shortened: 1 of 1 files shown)


plugins/git/webgui/js/gitMenu.js
```diff
@@ -30,8 +31,37 @@ function (topic, Menu, MenuItem, PopupMenuItem, model, viewHandler) {
...
+  var fileMenu = {
+    id : 'git-text-team-file',
+    render : function (fileInfo) {
+      if (fileInfo.type === "Dir")
+        return;
+
+      var submenu = new Menu();
+
+      submenu.addChild(new MenuItem({
+        label : 'Blame',
```
whisperity, line 48: I know this might be a bit too much political correctness but this is the part we could rename "Blame" into something less negative. GitHub calls it "History" or "recent changes". :slightly_smiling_face:

plugins/git/webgui/js/gitMenu.js
```diff
@@ -30,8 +31,37 @@ function (topic, Menu, MenuItem, PopupMenuItem, model, viewHandler) {
...
+        onClick : function () {
+          topic.publish('codecompass/gitblame', {
+            fileInfo : fileInfo
+          });
+        }
+      }));
+
+      return new PopupMenuItem({
+        label : 'Team',
```
whisperity, line 57: Also, @bruntib, why do we call this "Team"?

plugins/git/webgui/js/gitMenu.js
bruntib: I think we should use the Git conventional names. If it's "blame" there then we should follow it. If they rename it, we can rename it too.

plugins/git/webgui/js/gitMenu.js
bruntib: Yes, if we want then we can rename it to something else which describes better that this is a Git feature.

plugins/git/webgui/js/gitMenu.js
intjftw: I agree with @bruntib .

plugins/git/webgui/js/gitMenu.js
intjftw: I think Team is fine, but if we look for another name, simply "Git" could be appropriate.

PR #409, opened 2020-07-25
Docker runtime image

mcserep:
Currently there are 2 docker images:
1. The `codecompass:dev` image holds the development environment for CodeCompass.
2. The `codecompass:web` contains a compiled CodeCompass with its runtime dependencies and is configured to easily launch the `CodeCompass_webserver` binary.

Unfortunately none of them suffices to use the `CodeCompass_parser` binary in a dockerized production environment.
* The `codecompass:dev` image is not self-contained, only the dependencies are installed inside the container, the source code and the binaries of CodeCompass are outside it to support development. Starting with this image CodeCompass has to be downloaded and built first, which is unnecessary extra time spent.
* The  `codecompass:web` image is a set up to create so called *executing containers*. Upon the start of a container the `CodeCompass_webserver` is started as a non-root `codecompass` user inside the container. It requires significant extra work to circumvent this and use this image as the root user without starting the webserver, so some parsing can be done. (Root privilege is required to install the dependencies of the project to be parsed.)

This pull request proposes to a third, `codecompass:runtime` image which contains the built CodeCompass binaries and their runtime dependencies (but not those required only for compilation time). Then the `codecompass:web` image can be based upon this `codecompass:runtime` and really only contain the dependencies of the webserver. (Smaller size.)

I have also done some completion on the documentation.

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -27,6 +27,7 @@ RUN set -x && apt-get update -qq \
     libgvc6 \
     libthrift-dev \
     libodb-sqlite-dev \
+    libodb-pgsql-dev \
     # To switch user and exec command.
     gosu \
   && apt-get clean \
--- a/docker/README.md
+++ b/docker/README.md
@@ -127,7 +127,7 @@ variables:
 
 |       Variable       |                  Meaning                 |
 | -------------------- | ---------------------------------------- |
-| `CC_VERSION` |  The branch of the CodeCompass repository to use. |
+| `CC_VERSION` |  The branch, version hash or tag of the CodeCompass repository to use. |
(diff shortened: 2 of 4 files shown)


docker/web/Dockerfile
```diff
@@ -51,10 +25,9 @@ RUN set -x && apt-get update -qq \
     libgit2-dev \
     libssl1.1 \
     libgvc6 \
-    libmagic-dev \
     libthrift-dev \
     libodb-sqlite-dev \
-    ctags \
+    libodb-pgsql-dev \
```
mcserep, line 30: `libodb-pgsql-dev` shall be a dependency, otherwise the image would be unusable with PgSQL.

PR #410, opened 2020-07-25
Use PostgreSQL in Docker containers

mcserep:
The `libodb-pgsql-dev` dependency is missing from the new Ubuntu 20.04 docker images.
PostgreSQL supersedes SQLite in performance, therefore we shall support it also in the docker images for production systems.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -25,7 +25,7 @@ RUN set -x && apt-get update -qq \
   llvm-7 clang-7 llvm-7-dev libclang-7-dev \
   npm \
   thrift-compiler libthrift-dev \
-  odb libodb-sqlite-dev && \
+  odb libodb-sqlite-dev libodb-pgsql-dev && \
   ln -s /usr/bin/gcc-9 /usr/bin/gcc && \
   ln -s /usr/bin/g++-9 /usr/bin/g++
 
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -54,6 +54,7 @@ RUN set -x && apt-get update -qq \
     libmagic-dev \
     libthrift-dev \
     libodb-sqlite-dev \
+    libodb-pgsql-dev \
     ctags \
     # To switch user and exec command.
     gosu \


whisperity:
> The `libodb-pgsql-dev` dependency is missing from the new Ubuntu 20.04 docker images.

Ah, at first I thought you mean this in a way that they are gone from the upstream...

whisperity:
Is installing `libodb-pgsql-dev` enough to have the ability to connect to a remote postgresql host? The changes in the patch don't bring in the ability to run PgSQL to the image.

mcserep:
> Is installing `libodb-pgsql-dev` enough to have the ability to connect to a remote postgresql host? The changes in the patch don't bring in the ability to run PgSQL to the image.

The `web` image contains `postgresql-server-dev-12`, but the `dev` image does not contain it, because it neither did before.
Would it be required, shall we add it? I would assume that `libodb-pgsql-dev` contains everything required to use ODB with PostgreSQL.

whisperity:
`libodb-pgsql-dev` depends on `libodb-pgsql-2.4` which in turn depends on `libpq5`, so we're good.

bruntib:
Lacking PostgreSQL from the development image was half-intentional. For CodeCompass development SQLite could be enough. Anyway if we want to support our Docker image to work with PosgreSQL, then shouldn't it connect to a database running on the host machine or in a PostgreSQL Docker container? Maybe the extension of the documentation would also be useful on how to do it.

mcserep:
> Lacking PostgreSQL from the development image was half-intentional. For CodeCompass development SQLite could be enough. Anyway if we want to support our Docker image to work with PosgreSQL, then shouldn't it connect to a database running on the host machine or in a PostgreSQL Docker container? Maybe the extension of the documentation would also be useful on how to do it.

I agree and have very similar configuration myself in production: CodeCompass running in docker, PostgreSQL server running on the host machine.

But `libodb-pgsql-2.4` is required to compile CodeCompass in the first place. The `dev` image is used to build CodeCompass for the `web` image. If the `dev` image does not contain `libodb-pgsql-2.4`, the `web` image will currently fail to build when `CC_DATABASE` is set `pgsql`. That is a current configuration now and therefore this is a bug currently.

PR #414, opened 2020-07-28
Enable Google Analytics monitoring

mcserep:
With this new feature Google Analytics can be enabled on the webserver, monitoring webpage traffic, which can be useful in production environments.

Using Google Analytics is completely optional and by default disabled.
It can be enabled by creating a `ga.txt` file in the workspace directory, storing the tracking ID.

When the `CodeCompass_webserver` is started like it will check for the existence of this `ga.txt` file in the workspace directory. When found, the webserver will reroute all HTTP requests for `/ga.txt` to this file. CodeCompass web frontend will look for this file (through async Javascript) and: 
 * when found, Google Analytics will be loaded and configured in the browser;
 * when not found, Google Analytics is **NOT** loaded.


whisperity:
I think embedding Google Analytics (i.e. third-party, but also potentially invoking privacy issues!) stuff into the code needs to be looked over by the legal managerial team for approval.

The CodeCompass binary requiring *write* access to the install directory is a too harsh requirement (for corporate deployments, anyways). In most cases, these tools are available to users from a central, network-mounted, from the user's POV read-only storage. In addition, this will result in two servers from the same install directory (or many servers, in case of network mounted storage...) not having the possibility to use different analytics cookies.

It does not have to be a network-mounted storage. Think about having CodeCompass installed into standard `/usr/bin` and `/usr/lib` (or even `/usr/local/bin` and `/usr/local/share`, etc.).

We should instead re-route the serving of the HTTP request of `GET /scripts/ga.txt HTTP/1.1 [...]` *inside* the running server to echo back the file contents from memory.

webgui/scripts/ga.js
```diff
@@ -0,0 +1,20 @@
...
+            gtag('js', new Date());
+            gtag('config', gaId);
+        },
+        statusCode: {
+            404: function() {
+                console.log('Google Analytics disabled.');
+            }
+        }
+    });
+});
```
whisperity, line 20: Please add the newline.

mcserep: > I think embedding Google Analytics (i.e. third-party, but also potentially invoking privacy issues!) stuff into the code needs to be looked over by the legal managerial team for approval.

CodeCompass is an open-source software, and this is a purely optional (by default disabled) feature, so I don't think any approval from Ericsson's legal managerial team is required. Legal approval *could be* required for CodeCompass servers running inside Ericsson with this feature enabled, but that is another topic.
I would like to emphasize that no Google Analytics code is executed when the feature is disabled, therefore there is no possible privacy issue in that case.

> The CodeCompass binary requiring _write_ access to the install directory is a too harsh requirement (for corporate deployments, anyways). [...]

Your technical suggestions are more concerning. To be honest, I wanted this be this a minimal effort feature, since it is not that important, but it would be nice to see some web traffic info e.g. on codecompass.net in the future.
But your concerns are valid, I will update this feature some time later and serve `/scripts/ga.txt` dynamically instead from the install folder.

whisperity:
I just wanted to make sure we can go forward in way that's both legally okay (thanks for the clarification that we're not embedding any code!), and that won't jeopardize efforts of ever supporting, for example, a Debian package sort of use case.

(Note that with Docker this "overlaying" issue is trivially fixed.)

I'm also curious about the traffic stats of the webpage!

IIRC, the code which routes a request to either Thrift or to the filesystem is rather small and self-contained (both in the old and the new Mongoose). Simply creating a temporary file and serving that when `/ga.txt` is requested should be okay, I do not think we need to go out of our way to construct the appropriate response headers and etc. when we can just serve a "static file" as is.

Another, and come to think of it, maybe **better** option would be to just simply ditch the whole command-line option for the analysis tag, and similar to how `authentication.json` and `server.key` are *expected* to be in the "workspace" for the running server if you want those features, simply expect a `ga.txt` in the workspace - if it is there, go and enable the feature.
This still has the issue of routing the service of the file to a different location, but we no longer have the issue of having to manage temporary files.
In addition, setting something in the configuration file is considered a tiny bit safer than specifying something on the command-line, as by default on a shared Linux system, everyone can read command lines of every process. Not that this cookie tag would be that much of a *secret* in the first place.

But both of these methods do away with the requirement to *write* the install location.
I think it is a reasonable expectation that the user running the webserver has authority over the `workspace` directory...
It is also reasonable to design in a way that users won't start multiple servers with conflicting configuration but using the same workspace.

webgui/scripts/ga.js
```diff
@@ -0,0 +1,20 @@
+$(document).ready(function() {
+    $.ajax({
+        url: 'ga.txt',
+        dataType: 'text',
+        success: function (gaId) {
+            console.log('Google Analytics enabled: ' + gaId);
+            $.getScript('https://www.googletagmanager.com/gtag/js?id=' + gaId);
+
+            window.dataLayer = window.dataLayer || [];
+            function gtag(){dataLayer.push(arguments);}
```
whisperity, line 10: This function is weirdly formatted.

mcserep: Because I have copied the template code:
https://developers.google.com/analytics/devguides/collection/gtagjs

Of course we format it according to our coding conventions 😄

mcserep: Fixed.

mcserep:
> Another, and come to think of it, maybe **better** option would be to just simply ditch the whole command-line option for the analysis tag, and similar to how `authentication.json` and `server.key` are _expected_ to be in the "workspace" for the running server if you want those features, simply expect a `ga.txt` in the workspace - if it is there, go and enable the feature.
> This still has the issue of routing the service of the file to a different location, but we no longer have the issue of having to manage temporary files.
> In addition, setting something in the configuration file is considered a tiny bit safer than specifying something on the command-line, as by default on a shared Linux system, everyone can read command lines of every process. Not that this cookie tag would be that much of a _secret_ in the first place.

Implemented in such a way @whisperity , check it when you have time.

webgui/index.html
```diff
@@ -108,6 +108,9 @@
         });
       });
     </script>
+
+    <!-- Google Analytics -->
```
whisperity, line 112: As discussed on the meeting, for privacy's sake, let's add a comment here:

```suggestion
    <!-- Google Analytics (optionally loaded!) -->
```

webgui/login.html
```diff
@@ -52,6 +52,9 @@
             })
         });
     </script>
+
+    <!-- Google Analytics -->
```
whisperity, line 56:
```suggestion
    <!-- Google Analytics (optionally loaded) -->
```

webgui/scripts/ga.js
whisperity: Not fixed. It could be that Git add needs to be forced, or hooks disabled at commit, so this goes through. (My local Git config is also iffy about newlines on brand new files...)

webserver/src/webserver.cpp
```diff
@@ -38,7 +38,9 @@ po::options_description commandLineArguments()
          "Logging level of the parser. Possible values are: debug, info, warning, "
          "error, critical")
         ("jobs,j", po::value<int>()->default_value(4),
-         "Number of worker threads.");
+         "Number of worker threads.")
+        ("analytics", po::value<std::string>(),
+          "Google Analytics Tracking ID.");
```
whisperity, line 43: This is unused now.

```suggestion
```

webgui/scripts/ga.js
mcserep: Added it.

mcserep:
@zporky Is it fine to merge this PR or are there any legal (or other) concerns we shall still discuss?

zporky:
can we keep this unmerged until Wednesday's meeting?

Máté Cserép <notifications@github.com> ezt írta (időpont: 2020. aug. 4., K,
11:24):

webserver/src/webserver.cpp
```diff
@@ -117,6 +117,20 @@ int main(int argc, char* argv[])
...
+    auto gaFilePath = boost::filesystem::path(
+      vm["workspace"].as<std::string>()).append("ga.txt");
+    if (boost::filesystem::is_regular_file(gaFilePath))
+    {
+      requestHandler.gaTrackingIdPath = gaFilePath.string();
+      LOG(info) << "Google Analytics monitoring enabled.";
+    }
+    else
+    {
+      LOG(debug) << "Google Analytics monitoring disabled.";
```
bruntib, line 131: This message printed as a debug log, but on the other branch this is info log.

whisperity: That's not a problem. The default behaviour is to not enable it. SSL/authentication also uses different channels.

Issue #418, opened 2020-08-02
Reduce database size


bruntib:
According to `sqlite3_analyzer` tool the 3 largest tables are `CppDocComment` (50.3%), `CppAstNode` (33.1%) and `FileContent` (5.0%). CppDocComment is this huge, because we're storing the comments in HTML format with a lot of styling. We should remove HTML formatting.

Related issues: #417, #21.

mcserep:
I am moving the remaining tasks in this issue to the next release.

PR #419, opened 2020-08-02
Cpp documentation comments in markdown format

bruntib:
The `CppDocComment` table now contains documentation comments in markdown
format instead of HTML. With this change the SQLite database size of a Xerces
parsing is reduced to 219M from 302M.

Fixes a part of #418

### DIFF
--- a/plugins/cpp/model/include/model/cppdoccomment.h
+++ b/plugins/cpp/model/include/model/cppdoccomment.h
@@ -26,7 +26,7 @@ struct CppDocComment
   unsigned long long contentHash;
 
   #pragma db not_null
-  std::string contentHTML;
+  std::string content;
 
   #pragma db not_null index
   unsigned long long mangledNameHash;
--- a/plugins/cpp/parser/src/doccommentformatter.cpp
+++ b/plugins/cpp/parser/src/doccommentformatter.cpp
@@ -146,11 +146,11 @@ FullCommentParts::FullCommentParts(
 }
 
 // https://llvm.org/svn/llvm-project/cfe/trunk/lib/Index/CommentToXML.cpp
-class CommentToHTMLConverter
-  : public ConstCommentVisitor<CommentToHTMLConverter>
+class CommentToMarkdownConverter
+  : public ConstCommentVisitor<CommentToMarkdownConverter>
 {
 public:
-  CommentToHTMLConverter(
+  CommentToMarkdownConverter(
     const FullComment* fullComment_,
     const CommandTraits& traits_,
     std::stringstream& res_)
@@ -225,12 +225,6 @@ public:
    */
   void visitNonStandaloneParagraphComment(const ParagraphComment* c_);
 
-  /**
-   * This function escapes html elements from parameter and appends the escaped
-   * string to the result.
-   */
-  void appendToResultWithHTMLEscaping(const clang::StringRef& str_);
-
 private:
   const FullComment* _fullComment;
   const CommandTraits& _traits;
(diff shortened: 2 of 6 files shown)


plugins/cpp/parser/src/doccommentformatter.cpp
```diff
@@ -239,30 +233,9 @@ class CommentToHTMLConverter
...
-      case '>':  _res << "&gt;";   break;
-      case '"':  _res << "&quot;"; break;
-      case '\'': _res << "&#39;";  break;
-      case '/':  _res << "&#47;";  break;
-      default:   _res << *it;      break;
-    }
-  }
-}
-
 void CommentToHTMLConverter::visitTextComment(const TextComment* c_)
```
whisperity, line 236: This isn't really a `ToHTML` converter anymore.

plugins/cpp/parser/src/doccommentformatter.cpp
```diff
@@ -518,7 +451,7 @@ namespace parser {
 using namespace clang::comments;
 
 std::string DocCommentFormatter::format(
-  clang::comments::FullComment *fc_,
+  clang::comments::FullComment* fc_,
```
whisperity, line 454: Unrelated change, which also breaks formatting.

plugins/cpp/parser/src/doccommentformatter.cpp
```diff
@@ -280,166 +253,135 @@ void CommentToHTMLConverter::visitInlineCommandComment(
...
 {
-  _res << "<" << c_->getTagName().str();
+  _res << '<' << c_->getTagName().str();
 
   if (c_->getNumAttrs())
     for (unsigned i = 0, e = c_->getNumAttrs(); i != e; i++)
     {
-      _res << " ";
+      _res << ' ';
       const HTMLStartTagComment::Attribute &Attr = c_->getAttr(i);
```
whisperity, line 285: There's still a reference to some "HTML" stuff here.

plugins/cpp/model/include/model/cppdoccomment.h
```diff
@@ -26,7 +26,7 @@ struct CppDocComment
   unsigned long long contentHash;
 
   #pragma db not_null
-  std::string contentHTML;
+  std::string content;
```
whisperity, line 29: I know in our current development branch the database model is moldable as it is, but why aren't we calling this `contentMD`?

plugins/cpp/parser/src/doccommentformatter.cpp
bruntib: I don't know why GitHub shows you " instead of ', but for me it is '.
`HTMLStartTagComment` is in `clang::comments`, it is not ours. This describes HTML fragments embedded in doc comments.

plugins/cpp/model/include/model/cppdoccomment.h
bruntib: I don't know if it is worth including the format name in the attribute. We can do that if you wish. By the way a next improvement would be storing only the location of doc comments instead of its content.

plugins/cpp/parser/src/doccommentformatter.cpp
mcserep: Why is there a `visitHTMLStartTagComment` and a `visitHTMLEndTagComment` method? Are they required?

bruntib: Parts of doc comments are visited by a clang visitor. Clang visits the `\brief` part, the documentation of parameters, HTML tags, etc.
```.cpp
/**
 * \brief This is a brief description.
 *
 * This is a <b>long description</b> of this smart function containing <i>HTML</i> tags.
 * \param x This is the first parameter.
 * \param y This is the second parameter.
 * \return The function returns an integer.
 */
int func(int x, int y);
```
So in this example the brief, the long description, parameter and return value documentation are visited separately by clang. Also `<b>` and `<i>` tags are visited by `visitHTMLStartTagComment` visitor function. To be honest, I haven't tried if these tags are present in their original form as the content of main description. It this case we really shouldn't visit them separately. I'll check it later.

whisperity: > I don't know why GitHub shows you " instead of ', but for me it is '.
@bruntib My comment is tagging line 321/285, which is one line below your changes. I can see the changed code appending a single char as a char and not as a string literal. :slightly_smiling_face: I put "HTML" into quotes to indicate that there are words in the program that still refer to HTML stuff.

bruntib: Oh, I got it. I misunderstood this. So the HTML stuff here is conscious and it really refers to HTML things, but I'll check if we really need it.

whisperity: Well, *Doxygen* in the standard format generates HTML and thus I guess it is allowed (in the source code) so that you, as the library developer, can embed HTML formatting to your generated documentation... I'm interesting what Clang does with this in the first place...

bruntib: Yes, that's the embedded HTML that this visitor function processes. I've checked it and the result markdown wouldn't contain these HTML tags if these visitor functions are missing. So I consider this as the final solution.

PR #420, opened 2020-08-02
Remove mangledName column from CppAstNode table

### LINKED ISSUE #417: Remove mangledName column from CppAstNode table
We're not using `mangledName` for anything. It is `mangledNameHash` which connects `CppAstNode` records with `CppEntity`s. I also think that printing mangled name doesn't help in code comprehension. So I suggest removing this column in order to reduce database size.

bruntib:
Column mangledName is not used anywhere and it consumes too much space
unnecessarily. So this column has been removed.

Fixes #417

### DIFF
--- a/plugins/cpp/model/include/model/cppastnode.h
+++ b/plugins/cpp/model/include/model/cppastnode.h
@@ -70,9 +70,6 @@ struct CppAstNode
   #pragma db null
   FileLoc location;
 
-  #pragma db null
-  std::string mangledName;
-
   std::uint64_t mangledNameHash;
 
   SymbolType symbolType = SymbolType::Other;
@@ -145,7 +142,6 @@ inline std::string CppAstNode::toString() const
   return std::string("CppAstNode")
     .append("\nid = ").append(std::to_string(id))
     .append("\nastValue = ").append(astValue)
-    .append("\nmangledName = ").append(mangledName)
     .append("\nlocation = ").append(location.file->path).append(" (")
     .append(std::to_string(
       static_cast<signed>(location.range.start.line))).append(":")
@@ -187,7 +183,6 @@ inline std::uint64_t createIdentifier(const CppAstNode& astNode_)
 
   res
     .append(astNode_.astValue).append(":")
-    .append(astNode_.mangledName).append(":")
     .append(std::to_string(
       static_cast<SymbolTypeInt>(astNode_.symbolType))).append(":")
     .append(std::to_string(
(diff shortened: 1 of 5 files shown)


mcserep:
`CppParserTest.Variable` fails with these modifications, see the CI pipeline for details.

bruntib:
Yes, I've seen it. It's interesting, because it passes on my machine. In this change we're querying nodes by uint64 hashes instead of strings, so I'm a bit worried about handling comparing these in database.

whisperity:
@bruntib did you recreate the database?

zporky:
Is this a non-reverse compatible change on the database structure? How can
we upgrade previous databases?

gsd

Whisperity <notifications@github.com> ezt írta (időpont: 2020. aug. 3., H,
13:59):

mcserep:
> How can we upgrade previous databases?

There is no database migration strategy supported in CodeCompass for now, so  the only option is to drop and recreate the database.

ODB has some kind of support for database migrations (see [Schema migration](https://www.codesynthesis.com/products/odb/doc/manual.xhtml#13.2)), but even in that case a new feature may require to reparse the project, because e.g. it stores new data. We could handle that by introducing a semantic versioning for CodeCompass, but I am getting far from this PR here 😄

bruntib:
By the way there was another schema modifier patch in #403: `action` column which joins to `BuildAction` table was removed. There are also further plans on schema change: file contents and documentation comments should be stored as binary blob instead of plain text.

bruntib:
@whisperity Thanks, it was a good idea. Now the tests pass. The problem was that anonymous namespace has a different "mangled name" on Travis, so we really can't rely on mangled name hash.

whisperity:
\off At this point in the project I don't think we should give too much thought into database upgrades.

whisperity:
Why was the mangled name stored in the database? I am sure it was used for *something* in the previous release... are we sure that we aren't pulling the rug out from under some feature that should be migrated over to the current master?

bruntib:
As far as I remember, mangled name has always been used for identifying a symbol in a cross translation unit manner. This way we could determine which symbol the given AST nodes are occurrences of. If you check the patch, you can see that this mangled name has never been read, but written only. In practice we always used its hashed value. The string representation is useless, not to mention that we assembled it manually, so we could have figured out any other identifier. In a next step it would be nice changing `mangledNameHash` column to `usr` after (Unified Symbol Reference) that Clang serves us in order to identify nodes in a CTU manner.

\off I agree, I wanted to accomplish these schema changes before somebody finds out that we must stay reverse compatible. Don't tell anyone, but I'm also planning to eliminate `CppEntity::astNodeId` column too, but I'm not 100% sure yet if it's really unused :)

Issue #405, opened 2020-07-21
SEGFAULT at the exit of parser binary on Ubuntu 20.04


whisperity:
> :warning: **NOTE:** This is after applying #398.

When testing the aforementioned patch, the `SEGFAULT` that seemed to appear in Travis since #393 have found their way into the docker-

I have started the container as described [in the docs](https://github.com/Ericsson/CodeCompass/tree/master/docker#how-to-parse-a-project), but with changed mode:

    cd ~/Work/CodeCompass
    docker run --rm -ti \
      --env DATABASE=sqlite --env BUILD_TYPE=RelWithDebInfo \
      --volume $(readlink -f .):/CodeCompass \
      --volume $(readlink -f CodeCompass):/projects/myproject \
      -p 8001:8080 \
      codecompass:dev \
      /bin/bash

The `SEGFAULT` does **NOT** happen on `Debug` builds.

It happens on `Release` (and `RelWithDebInfo`), irrespective of the `-j` argument to the parser. So it's not a multithreading yadda.

Trying to parse CodeCompass itself (on the branch `bruntib:docker_ubuntu20`, **inside** the container):

    CodeCompass_parser -d "sqlite:database=/CodeCompass/workspace/myproject/data.sqlite" -w /CodeCompass/workspace -n CodeCompass -i /CodeCompass/CodeCompass -i /CodeCompass/build/compile_commands.json

Thanks to `RelWithDebInfo` (and manually installing `gdb` into the image), I've
got a trace:

~~~~
[INFO] Creating indexes from file /CodeCompass/install/share/codecompass/sql/statistics-odb.sql
[INFO] Creating indexes from file /CodeCompass/install/share/codecompass/sql/cpptypedef-odb.sql

Program received signal SIGSEGV, Segmentation fault.
0x000055df7d13c42e in boost::any::~any (this=0x55df7e067d90, __in_chrg=<optimized out>) at /usr/include/boost/any.hpp:79
79                  delete content;

(gdb) bt
#0  0x000055df7d13c42e in boost::any::~any (this=0x55df7e067d90, __in_chrg=<optimized out>) at /usr/include/boost/any.hpp:79
#1  boost::program_options::variable_value::~variable_value (this=0x55df7e067d90, __in_chrg=<optimized out>) at /usr/include/boost/program_options/variables_map.hpp:58
[...]

PR #425, opened 2020-08-07
Better AST value for nodes.

### LINKED ISSUE #358: What should astValue be?
I think we should discuss what `astValue` in `CppAstNode` table ought to be? We have always used this field as a human-readable description about the symbol which can be displayed in the Info Tree for example. However, in some situations this field is used as a meaningful part of the logic. For example syntax highlight uses it to find the borders of a symbol's name.

In case of functions currently `astValue` contains their signature. This is good for displaying in Info Tree, but not suitable for syntax highlight. By the way now we can't highlight functions until it's fixed. The name and the signature of a function are stored in `CppNamedEntity`, but I don't think it would be efficient to select all symbols in this table for syntax highlight.

So what is `astValue` of a function? Maybe this could contain only the function name and the signature could go to some `CppFunction` table. This way the signature still remains available for Info Tree where it is fetched more rarely.

bruntib:
The AST value of a symbol should be a human-readable source code fragment which
contains the given AST node. In this patch we try to determine the best context
in which a given AST node resides and choose this as the astValue.

Fixes #358

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -92,10 +92,15 @@ public:
     {
       model::CppAstNodePtr typeLocAstNode = p.second;
 
-      auto it = _locToAstType.find(p.first);
+      auto itAstType = _locToAstType.find(p.first);
 
-      if (it != _locToAstType.end())
-        typeLocAstNode->astType = it->second;
+      if (itAstType != _locToAstType.end())
+        typeLocAstNode->astType = itAstType->second;
+
+      auto itAstValue = _locToAstValue.find(p.first);
+
+      if (itAstValue != _locToAstValue.end())
+        typeLocAstNode->astValue = itAstValue->second;
 
       typeLocAstNode->id = createIdentifier(*typeLocAstNode);
 
@@ -129,7 +134,7 @@ public:
     if (decl_)
       _isImplicit = decl_->isImplicit() || _isImplicit;
 
-    bool b = clang::RecursiveASTVisitor<ClangASTVisitor>::TraverseDecl(decl_);
+    bool b = Base::TraverseDecl(decl_);
 
     _isImplicit = prevIsImplicit;
 
@@ -140,8 +145,7 @@ public:
   {
     _functionStack.push(std::make_shared<model::CppFunction>());
 
-    bool b = clang::RecursiveASTVisitor<
-      ClangASTVisitor>::TraverseFunctionDecl(fd_);
+    bool b = Base::TraverseFunctionDecl(fd_);
 
     if (_functionStack.top()->astNodeId)
       _functions.push_back(_functionStack.top());
@@ -154,8 +158,7 @@ public:
   {
     _functionStack.push(std::make_shared<model::CppFunction>());
 
-    bool b = clang::RecursiveASTVisitor<
-      ClangASTVisitor>::TraverseCXXMethodDecl(fd_);
+    bool b = Base::TraverseCXXMethodDecl(fd_);
 
     if (_functionStack.top()->astNodeId)
       _functions.push_back(_functionStack.top());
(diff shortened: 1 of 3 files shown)


bruntib:
This patch most probably messes up server-side syntax highlighting, because that feature also relies on `astValue` attribute.

zporky:
The ideal solution should be both flexible but also not wasting space. If AST value consists of multiple characters (much longer than today) it could be multiple the source fragments we store.

Tibor mentioned a possibility that the new solution would be not a repetition of some source fragment, but a reference to  the source content, like: (fileid,beg,end) triplet. That could work, but in a normal InfoTree we have to gather these information from different files - although in lazy way. This should be implement and measure. The reference can also store the information about formatting the text in InfoTree.

To keep flexibility we might still also add custom text to AST Value, like disambiguation (e.g. "template def" or "instantiation" or something similar).

mcserep:
When I applied this branch to the current master (fcf966621fdcd245d0ad91f0b7cc1d7ab609a87f), parsing segfaults for me:
```
[INFO] [cppparser] parse started!
[INFO] (1/2) Parsing /home/mate/cc/sources/tinyxml2/tinyxml2.cpp
[INFO] (2/2) Parsing /home/mate/cc/sources/tinyxml2/xmltest.cpp
Segmentation fault (core dumped)
```
I haven't done any further examination of the issue yet, anyone has similar issues?

whisperity:
> When I applied this branch to the current master ([fcf9666](https://github.com/Ericsson/CodeCompass/commit/fcf966621fdcd245d0ad91f0b7cc1d7ab609a87f)), parsing segfaults for me at the end of the C++ parser:
> 
> ```
> [INFO] [cppparser] parse started!
> [INFO] (1/2) Parsing /home/mate/cc/sources/tinyxml2/tinyxml2.cpp
> [INFO] (2/2) Parsing /home/mate/cc/sources/tinyxml2/xmltest.cpp
> Segmentation fault (core dumped)
> ```
> 
> I haven't done any further examination of the issue yet, anyone has similar issues?

@mcserep 18.04 or 20.04? There is some weirdness with a destructor in Boost on 20.04, see #405. Should not affect the created database.

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -277,6 +282,60 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
     return b;
   }
 
+  bool TraverseCallExpr(clang::CallExpr* ce_)
+  {
+    _contextStatementStack.push(ce_);
+    bool b
+      = clang::RecursiveASTVisitor<ClangASTVisitor>::TraverseCallExpr(ce_);
```
whisperity, line 289: Could you please create (if it doesn't exist yet), a private `using Base = clang::RecursiveASTVisitor<ClangASTVisitor>;` in the class (this is usually the convention with Clang RASTVs), and use `Base::TraverseCallExpr(ce_);` instead? Same with every other similar construct in the code.

mcserep:
> > When I applied this branch to the current master ([fcf9666](https://github.com/Ericsson/CodeCompass/commit/fcf966621fdcd245d0ad91f0b7cc1d7ab609a87f)), parsing segfaults for me at the end of the C++ parser:
> > ```
> > [INFO] [cppparser] parse started!
> > [INFO] (1/2) Parsing /home/mate/cc/sources/tinyxml2/tinyxml2.cpp
> > [INFO] (2/2) Parsing /home/mate/cc/sources/tinyxml2/xmltest.cpp
> > Segmentation fault (core dumped)
> > ```
> > 
> > 
> > I haven't done any further examination of the issue yet, anyone has similar issues?
> 
> @mcserep 18.04 or 20.04? There is some weirdness with a destructor in Boost on 20.04, see #405. Should not affect the created database.

Ubuntu 18.04. There is no such issue in the CI, but maybe we could update the PR branch with `master` to see if the error is reproduced in the CI in that case.

bruntib:
In case this is a release build and if this segmentation fault is at the very end of the parsing process (i.e. after adding database indexes) then this is a known issue as @whisperity mentioned. For some yet unknown reason a destructor from Boost argument parser lib gives it.

mcserep:
> In case this is a release build and if this segmentation fault is at the very end of the parsing process (i.e. after adding database indexes) then this is a known issue as @whisperity mentioned. For some yet unknown reason a destructor from Boost argument parser lib gives it.

That issue is only affecting Ubuntu 20.04, so it is not the reason.

In this case the segfault is before adding the database indexes. It also prevents other parsers to run.

bruntib:
@whisperity Thanks for the suggestion of the more concise base class referencing.
@mcserep There really was a bug which caused this segmentation fault, it is supposed to be fixed now.

Issue #415, opened 2020-07-29
C++ parser segfaults on PCL


mcserep:
I have tried to parse the Point Cloud Library (PCL) with CodeCompass.
[PCL](https://github.com/PointCloudLibrary/pcl) is a popular library written in C++ with a CMake build system and shall be "compatible" with CodeCompass.

I have built PCL (with GCC though), but then the C++ parser fails on the first TU with a segmentation fault (running on 1 cores):
```
[INFO] [cppparser] parse started!
[INFO] (1/389) Parsing /projects/pcl/common/src/point_types.cpp
warning: unknown warning option '-Wabi=11' [-Wunknown-warning-option]
Segmentation fault (core dumped)
```

Running with Valgrind:
```
[INFO] [cppparser] parse started!
[INFO] (1/389) Parsing /projects/pcl/common/src/point_types.cpp
warning: unknown warning option '-Wabi=11' [-Wunknown-warning-option]
==12408== Invalid read of size 1
==12408==    at 0xA77F860: cc::parser::getMangledName[abi:cxx11](clang::MangleContext*, clang::NamedDecl const*, cc::model::FileLoc const&) (in /codecompass/lib/parserplugin/libcppparser.so)
==12408==    by 0xA77FC9B: cc::parser::getMangledName[abi:cxx11](clang::MangleContext*, clang::NamedDecl const*, cc::model::FileLoc const&) (in /codecompass/lib/parserplugin/libcppparser.so)
==12408==    by 0xA738FC2: cc::parser::ClangASTVisitor::VisitVarDecl(clang::VarDecl*) (in /codecompass/lib/parserplugin/libcppparser.so)
==12408==    by 0xA73B73C: clang::RecursiveASTVisitor<cc::parser::ClangASTVisitor>::TraverseDecl(clang::Decl*) (in /codecompass/lib/parserplugin/libcppparser.so)
==12408==    by 0xA73C471: clang::RecursiveASTVisitor<cc::parser::ClangASTVisitor>::TraverseDeclStmt(clang::DeclStmt*, llvm::SmallVectorImpl<llvm::PointerIntPair<clang::Stmt*, 1u, bool, llvm::PointerLikeTypeTraits<clang::Stmt*>, llvm::PointerIntPairInfo<clang::Stmt*, 1u, llvm::PointerLikeTypeTraits<clang::Stmt*> > > >*) (in /codecompass/lib/parserplugin/libcppparser.so)
[...]

PR #423, opened 2020-08-06
Fix C++ parser segfaults when the parent node of a variable

zporky:
declaration is a CapturedDecl.

Fixes #415

PR #428, opened 2020-08-09
Fix parser segfault

zporky:
Workaround ISSUE #415 Parser segfaults.
There are contexts which getParentFunctionOrMethod() is not null but the result is not FunctionDecl (like CapturedDecl)
In this case we go up on the getLookupParent() until we reach a Function or TU decl context.
Seems to work on multiple nested #pragma context.

### DIFF
--- a/plugins/cpp/parser/src/symbolhelper.cpp
+++ b/plugins/cpp/parser/src/symbolhelper.cpp
@@ -7,7 +7,12 @@
 #include <model/fileloc.h>
 #include <model/fileloc-odb.hxx>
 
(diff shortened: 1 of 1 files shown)


whisperity:
Please try updating the original patch #423 where the original discussion is present. You have pushed the feature branch to the **wrong remote**! Use `git push origin fix_parser_segfault` and ensure that your `origin` follows conventions and it is the `github.com/zporky/CodeCompass.git` repository. Do **NOT** push feature branches to the `upstream` repository.

Issue #427, opened 2020-08-09
How to identify an AST node across translation units?


bruntib:
Currently our hand-made mangled name and later USR will be able identifying AST nodes across translation units uniquely. However, what happens in those cases when a project contains two executables?
```.cpp
// Executable 1: first.cpp
class MyClass {
public:
  int i;
  virtual void f() = 0;
};
int main()
{
}

// Executable 2: second.cpp
class MyClass {
public:
  int i;
};
int main()
{
}

// Build
// g++ first.cpp
// g++ second.cpp
```
When opening InfoTree for classes, CodeCompass is able to show whether the given class is an abstract type. We can do that, since we store this boolean information in table `CppType`. Due to historical and some unsolved technical reasons column `mangledNameHash` in `CppEntity` (from which `CppType` inherits) is not unique. It should be unique, since the idea behind these tables is that it contains general information about a named entity. But even if `mangledNameHash` is not unique in this table, how could we distinguish which binary's symbol has been clicked in the GUI for InfoTree? Mangled name doesn't contain this piece of information.

bruntib:
I don't know if this observation is related, but let's consider the following example:
```.cpp
void func(int param);
void func(int param);
void func(int param)
{
}
```
When opening the InfoTree of a `func` declaration and clicking on its parameters in the InfoTree, it jumps to the parameters of the corresponding `func` declaration. This is a good behavior. We can do this, because currently in this code fragment the mangled names of the three `param` variables are different. This also means that we can't say that these three variables are semantically equivalent. Fortunately we don't have a feature in CodeCompass which requires the ability of connecting these variables.

When changing to USR, we should check if it generates different identifiers for these three variables. If they have the same USR then we can't eliminate `astNodeId` column from `CppEntity`, because this column binds the entity to a specific declaration. Finding this connection will also cost some extra operations.

whisperity:
First things first, neither the mangled name nor the USR will "really" contain this information, as while some aspects of ODR is applicable to a program level and not the translation unit level, "multiple programs" are not defined by the standard at all. However, #198 does exactly this, as long as we are using a `compile_commands.json` that also contains link commands. At encountering of an ambiguous symbol set, it uses the link graph (flattened at parse time so it's not heavy on the service time) to highlight symbols that are from TUs that are known to be linked together.

PR #429, opened 2020-08-12
Usr

bruntib:
This is a continuation of #420, so that pull request must be accepted first and this will be rebased on that.
TODO for this pull request:
- `mangledName` is still used at several places in the code, however, we have USR instead.
- There is a `generateUSRForType()` function in a later version of Clang. We need to upgrade Clang so we can use it instead of the current `getMangledName(QualType)` implementation.


whisperity:
What if the following happens:

```cpp
// void use0() { f(-1); } -- cannot be done as symbol is not declared.

int f(int x);

void use1() { f(0); }
// CallExpr is linking the forward declaration

int f(int x) { return x; }

void use2() { f(1); }
// CallExpr is linking the definition
```

The two `CallExpr`s link two separate `FunctionDecl`s that are only reachable from one another because of a ReDecl chain.

Are we sure that the two functions will have USRs that are able to be found together? If not, we will break "jump to definition" across translation units, as calls from a file where in that TU only the "declaration" exists (due to headers) will only be able to jump to that symbol...

This should be tested.

plugins/cpp/parser/src/symbolhelper.cpp
```diff
@@ -64,127 +54,20 @@ const clang::Type* getStrippedType(const clang::QualType& qt_)
...
-    // #else ... #endif section then every usage will also have two versions.
-    // So if a function has a template and a non-template version at the same
-    // time then querying the definition of the non-template usage the template
-    // version also returns.
-    return out.str();
-  }
-
-  return nd_->getQualifiedNameAsString();
+  llvm::SmallVector<char, 64> usr;
+  clang::index::generateUSRForDecl(nd_, usr);
```
whisperity, line 60: `generateUSRFor...` functions return a bool which are signalling whether "the result of the call should be ignored".

bruntib: Thanks, I haven't noticed the return type. I'll check what it means and in what cases it returns true or false.

bruntib:
I strongly hope that all occurrences of function `f()` have the same USR in your code example. That's the goal of USR. It's a more interesting questions, what's the USR of the different `x` variables? This is the same question as in #427. Next week I'll have more time to investigate.

plugins/cpp/parser/src/symbolhelper.cpp
bruntib: We investigated a little the return value of this function. I put a relatively long comment in the source code which describes why we don't use this value:
https://github.com/Ericsson/CodeCompass/pull/429/files#diff-b735f57c014d7a4293c17acf122b284dR60-R82
I don't know if this explanation is satisfactory.

bruntib:
I changed "mangledName" to "entityHash" everywhere in the source code, because it is not a mangled name anymore.
I think I can remove WIP label from this pull request soon, because it approaches the final solution. Can anyone see any regression compared to the previous solution? Thank you!

bruntib:
I'm brave enough now to remove WIP label from this patch. If you find some regression after using USR instead of our manually generated USR, feel free to comment it.

mcserep (approved):
While the CodeCompass self-parse test fails in the CI (on Ubuntu 18.04), it does not fail when tested on Ubuntu 16.04 and 18.04, on multiple computers.

I approve it, so it can fix #471 and we can get back to the CI issue later.

PR #432, opened 2020-08-14
Uplift used Clang to 10.0

bruntib:


whisperity:
Not sure if this is intended... when parsing CodeCompass itself, for the warnings of a parsed TU, only the warning count comes up, the warnings themselves are hidden: 

~~~~
[INFO] (62/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/session.cpp
[INFO] (63/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/sessionmanager.cpp
2 warnings generated.
[INFO] (64/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/threadedmongoose.cpp
[INFO] (65/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/mongoose.c
[INFO] (66/129) Parsing /home/w/Work/CodeCompass/Source/webserver/authenticators/plain/src/plugin.cpp
2 warnings generated.
~~~~

whisperity:
Alright, I can confirm that the parsing part works, except for that weirdness with the diagnostics output above. In addition, the same behaviour is present for *erroring* TUs:

~~~
2 errors generated.
Error while processing /home/w/Work/CodeCompass/Source/plugins/cpp/test/src/cppparsertest.cpp.
[WARNING] (100/129) Parsing /home/w/Work/CodeCompass/Source/plugins/cpp/test/src/cppparsertest.cpp has been failed.
~~~

bruntib:
I think warnings and errors are not emitted to the standard output because they are captured by the diagnostic message handler (see #403). I'll create a ticket about bringing them back.

bruntib:
It seems that only Xenial build is missing for this pull request. If we're extremely lucky then CodeCompass works with Clang-8 which is available under Xenial. Could somebody please help with some Travis magic so Xenial build uses Clang-8? This patch is a dependency of some other.

intjftw:
It seems to me that unfortunately the CI already uses Clang 8 for Xenial, too (at least that is what I saw in the build configuration, but correct me if I'm wrong).

bruntib:
Yes, I tried installing Clang 8 hoping that it will work automatically out of the box. However, clang-10 binary is still set explicitly somewhere. I haven't investigated the issue too much.

mcserep:
Why aren't we using Clang 10? It is available on Xenial:
https://apt.llvm.org/

mcserep:
> Why aren't we using Clang 10? It is available on Xenial:
> https://apt.llvm.org/

I see the problem, `llvm-toolchain-xenial-10` is not in the **APT source safelist** for Travis, as defined here:
https://github.com/travis-ci/apt-source-safelist/blob/master/ubuntu.json

But we can still add this PPA with the `sourceline` attribute, I will give it a go.

mcserep:
There, proper Travis magic applied, it builds on Xenial now in the CI pipeline 😄 

(I have made some commits, feel free to squash them of course.)

whisperity:
Oh nice, TIL if the pull request author allows you to push their PR, you can actually *force push* the branch too. @mcserep I've squashed your 3 commits and updated the message in the squash with the elaboration about this `sourceline` fiasco. Lovely, that there is an *official* APT (technically it is **not** a `PPA` at that point :wink:) repository for the newer versions.

I believe @bruntib is serving a well-deserved holiday this week, considering the elongated weekend for the state anniversary. So it's up to us to review whether everything is fine with the patch. :yum:

bruntib:
Thanks for the magic, I appreciate it:) By the way I was only checking packages.ubuntu.com and I didn't see if Clang 10 was available under xenial.

PR #434, opened 2020-08-14
Stop supporting Ubuntu 16.04 LTS (Xenial)

bruntib:


intjftw:
Is there any explicit reason for this? I think we should support Xenial at least until the end of its lifetime, April 2021.

bruntib:
According to my last experiences, I couldn't make our dependencies work even by compiling them from source code under 18.04. So technically we're not supporting 18.04 neither, not to mention 16.04.
If the situation has changed since my last trials, then I'd agree keeping the support of 16.04. I gave up the struggle last time, but I'll give it another try. That would also require compiling Clang from source, because after merging #432 CodeCompass will not compile with Clang-7 and as far as I've seen, package repository of Xenial doesn't contain newer Clang versions.

bruntib:
You must be right by the way, because Travis can also build CodeCompass somehow. I'll check how it's done there.

whisperity:
In addition, 16.04 was using a PPA in the guide to get Clang 7 because the official version was 3.8 or something by default. There _might_ be a chance there is a PPA for Clang 10, but I wouldn't bet on it. Are we sure downloading the binary release either from the website or the official LLVM repository doesn't work due to the whole RTTI stuff? Current build of LLVM in 20.04 repositories come with RTTI.

mcserep:
I don't think we shall stop supporting Ubuntu 16.04 before its official support ends, unless we face major issue with it in the future, which does not worth to solve for this short period of time.

> According to my last experiences, I couldn't make our dependencies work even by compiling them from source code under 18.04. So technically we're not supporting 18.04 neither, not to mention 16.04.

@bruntib I don't recall any problems regarding 16.04. There was an issue with ODB on 18.04 (see #385), but that was solved (by compiling ODB from source), and it did not affect 16.04 at all.

> There might be a chance there is a PPA for Clang 10, but I wouldn't bet on it.

@whisperity Clang 10 can be easily installed from the official PPA with RTTI (just like Clang 7 is installed now):
https://apt.llvm.org/

whisperity:
I believe thanks to @mcserep's update to #432 this can be postponed until Xenial goes EOL (expected April 2021)?

Ubuntu 18.04 definitely can build the dependencies: it works on Travis, and in addition, I've set up a proof-of-concept dependency release repository at [`whisperity:CodeCompass-Dependencies`](http://github.com/whisperity/CodeCompass-Dependencies).
In addition, thanks to neither ODB+GCC nor NPM+libSSL not being broken on 16.04, it's actually simpler to set up the build environment on 16.04 than it is on 18.04.

mcserep:
I will close this PR for now.

whisperity:
We can resurrect and rebase this in April! :hourglass_flowing_sand:

PR #438, opened 2020-08-17
Appending mangledNameHash to AstNode identifier.

### LINKED ISSUE #436: Proper template handling in C++
Currently templates are not handled properly:
```.cpp
template <typename>
struct S
{
  int member;
};

int main()
{
  S<int> s;
  int i = s.member;
}
```
Issues:
- [x] Querying info tree on `member` inside the definition of `S` shows that this variable has only 1 usage in this code. 
- [x] Clicking on `member` inside `main()` crashes the web server. (Fix: #438)

intjftw:
Fixes half of #436 
The problem was that in 0e85d45b3b94c8289ba4097213570be8d4f0fb0d appending `mangledName` was deleted from `createIdentifier`. This resulted in identifiers not being unique, thus the database overwrote certain records in the `CppAstNode` table. I completed the identifier generation with appending `mangledNameHash`.
The other original problem (showing only 1 usage for the templated struct's attribute) is still present.


intjftw:
@bruntib The CI fails on `CppParserTest.Record` which was modified in #420 :
```cpp
case -1: 
          EXPECT_TRUE(
            // Compiler generated assignment operator return value.
            n.astType == model::CppAstNode::AstType::ReturnTypeLoc ||
            // TODO: What's this? Maybe a compiler generated copy constructor
            // parameter's type? Then why isn't this a ParameterTypeLoc?
            n.astType == model::CppAstNode::AstType::TypeLocation);
          break;
```
That is kind of a weird TODO. The only change was switching `ParameterTypeLoc` to `ReturnTypeLoc` in the first comparison. If I write it back to `ParameterTypeLoc`, the test passes. Why was this change made? Can I write it back to the original?

bruntib:
Yes, please, write it back to the original. I also had some trouble with determining what exactly that node is. But even if we use `ParameterTypeLoc` then a `ReturnTypeLoc` will be missing for the return type of the compiler generated assignment operator. Please, leave some TODO comment, because this question needs some more investigation.

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -357,10 +357,10 @@ TEST_F(CppParserTest, Record)
       {
         case -1: 
           EXPECT_TRUE(
-            // Compiler generated assignment operator return value.
-            n.astType == model::CppAstNode::AstType::ReturnTypeLoc ||
-            // TODO: What's this? Maybe a compiler generated copy constructor
-            // parameter's type? Then why isn't this a ParameterTypeLoc?
+            // TODO: investigate the type of this. It is possibly a compiler
+            // generated copy constructor for which ParameterTypeLoc as
+            // a condition is correct. But maybe not.
```
bruntib, line 362:
```suggestion
            // TODO: investigate the type of this. It is possibly the parameter of a
            // compiler generated copy constructor or assignment operator.
            // The type should also be visible as a return type of the compiler generated
            // assignment operator.
```

bruntib (changes requested):

mcserep (approved):

PR #445, opened 2020-08-18
Fix plugin loading in frontend

### LINKED ISSUE #431: [Web] Plugins not loaded on frontend
**How to reproduce:** run the latest version of CodeCompass (`Ericsson/master`) on Ubuntu 20.04.
(The bug is not present on Ubuntu 18.04.)

**Malfunctions:**
- search toolbar is not shown on top;
- user guide tab is not shown on main area;
- right click functions basically do not work at all (menus are not shown).

There are no error messages on the JavaScript error log (in the browser), no failed HTTP request and no error on the server log (even with debug level logging).

**The bug can be tested here:**
http://codecompass.net:8020/

**Screenshots:**
image

image

mcserep:
Fixes #431.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,10 +31,9 @@ add_subdirectory(parser)
 add_subdirectory(scripts)
 add_subdirectory(service)
 add_subdirectory(util)
+add_subdirectory(plugins) # must precede webgui
 add_subdirectory(webgui)
 add_subdirectory(webserver)
-# Plugins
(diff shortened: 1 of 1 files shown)


mcserep:
> It is interesting to see that the original comment was `must precede webgui`, but if it works then it's fine for me.

I have added the `must precede webgui` comment now, in this PR. Maybe you switched up what was removed and added?

bruntib:
Yes, my bad :)

PR #448, opened 2020-08-18
Excluding node sort at function parameters.

### LINKED ISSUE #440: Function parameters appear in lexical order in info tree
When retrieving info tree for a function with multiple parameters, the `Parameters` menu shows the parameters in lexical order instead of the original order of the function signature.

image

intjftw:
Fixes #440 
The `cppservice` gets the parameters in a good order but sorts them at the end. Now the sorting is excluded when querying funtion parameters.

### DIFF
--- a/plugins/cpp/service/src/cppservice.cpp
+++ b/plugins/cpp/service/src/cppservice.cpp
@@ -958,7 +958,9 @@ void CppServiceHandler::getReferences(
         break;
     }
 
-    if (referenceId_ != PARAMETER)
+    if (referenceId_ != PARAMETER &&
+        referenceId_ != ENUM_CONSTANTS &&
+        referenceId_ != LOCAL_VAR)
       std::sort(nodes.begin(), nodes.end(), compareByValue);
 
     return_.reserve(nodes.size());
--- a/plugins/cpp/service/src/cppservice.cpp
+++ b/plugins/cpp/service/src/cppservice.cpp
@@ -959,8 +959,7 @@ void CppServiceHandler::getReferences(
     }
 
     if (referenceId_ != PARAMETER &&
-        referenceId_ != ENUM_CONSTANTS &&
-        referenceId_ != LOCAL_VAR)
+        referenceId_ != ENUM_CONSTANTS)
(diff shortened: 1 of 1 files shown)


bruntib:
The change is good.
In ODB it matters whether the list of things is stored in a set or in a vector inside an ODB class. If it's a vector then the order is also stored in an extra column. Currently we store two other things in a vector: the local variables of a function and the enum constants of an enum. We should consider not sorting these too, or it would be even better not paying for storing their orders by using sets. I think the order of local variables is not important. What do you think of enum constants? Should those be listed in alphabetical order or in their original order?

mcserep:
If a function has many local variables, wouldn't it be harder to locate a specific one, if their order is not sorted?
On the other hand, enum constants should not be sorted in lexical order in my opinion, because their defined order often has a particular meaning.

mcserep (approved):

bruntib:
Ok, I can agree. So the point is that in this case we shouldn't sort these neither.

mcserep:
> Ok, I can agree. So the point is that in this case we shouldn't sort these neither.

I am a bit confused. My point was that the local variables **shall be sorted**!

bruntib:
Sorry, it might have been confusing really :) I'm just saying that either we're sorting or we're using a vector. The two of them together doesn't make too much sense.

bruntib:
So I think this should be a `std::set`, so the original order is not stored unnecessarily:
https://github.com/Ericsson/CodeCompass/blob/6dfc28a890af17947b4e11f58479e637e56eb5ec/plugins/cpp/model/include/model/cppfunction.h#L20

intjftw:
I tried changing said `std::vector` to `std::set` but it didn't compile, saying there is no `operator<` which was needed somewhere. I think this modification should be another PR since it takes more time to make it work and that is unrelated to this PR.

zporky:
Yes, associative containers requires operator<(). However, here elements are odb::lazy_shared_ptr<CppVariable>>, so perhaps the operator should just compare the string elements pointed by the lazy_shared_ptr. I am not sure how this has an effect on lazyness however. 
Does not the current "unordered" sequence of locals are sorted by the order of declaration? If yes, I could skip this task.

PR #450, opened 2020-08-19
Print complete error and warning messages at parsing time

### LINKED ISSUE #437: [C++] Compiler diagnostics are silent in parser output
As discovered in #432 with #403 being the potential culprit, the Clang compiler diagnostics during parse time are not printed to the parser's standard streams anymore:

When analysing CodeCompass with CodeCompass itself:

~~~~
[INFO] (62/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/session.cpp
[INFO] (63/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/sessionmanager.cpp
2 warnings generated.
[INFO] (64/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/threadedmongoose.cpp
[INFO] (65/129) Parsing /home/w/Work/CodeCompass/Source/webserver/src/mongoose.c
[INFO] (66/129) Parsing /home/w/Work/CodeCompass/Source/webserver/authenticators/plain/src/plugin.cpp
2 warnings generated.

[...]

2 errors generated.
Error while processing /home/w/Work/CodeCompass/Source/plugins/cpp/test/src/cppparsertest.cpp.
[WARNING] (100/129) Parsing /home/w/Work/CodeCompass/Source/plugins/cpp/test/src/cppparsertest.cpp has been failed.
~~~~

Previously, the Clang diagnostics were formatted and output as if a true compiler was running. While it's less of a problem for warnings, for errors it's challenging to understand the output (especially without booting a server) - it says **2 errors generated** but never says what the errors are.

intjftw:
Fixes #437 
As @mcserep suggested in the ticket, the solution was to derive `DiagnosticMessageHandler` from `TextDiagnosticPrinter`. The `HandleDiagnostic` function of the latter class prints the warning and error messages that may occur while parsing a project.

### DIFF
--- a/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
+++ b/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
@@ -8,11 +8,10 @@ namespace parser
 {
 
 DiagnosticMessageHandler::DiagnosticMessageHandler(
-  llvm::raw_ostream& os,
   clang::DiagnosticOptions* diags,
   SourceManager& srcMgr_,
   std::shared_ptr<odb::database> db_)
-    : TextDiagnosticPrinter(os, diags), _srcMgr(srcMgr_), _db(db_)
+    : TextDiagnosticPrinter(llvm::errs(), diags), _srcMgr(srcMgr_), _db(db_)
 {
 }
 
--- a/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
+++ b/plugins/cpp/parser/src/diagnosticmessagehandler.cpp
@@ -8,9 +8,11 @@ namespace parser
 {
 
 DiagnosticMessageHandler::DiagnosticMessageHandler(
+  llvm::raw_ostream& os,
+  clang::DiagnosticOptions* diags,
   SourceManager& srcMgr_,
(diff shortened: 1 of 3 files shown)


plugins/cpp/parser/src/diagnosticmessagehandler.cpp
```diff
@@ -8,17 +8,19 @@ namespace parser
 {
 
 DiagnosticMessageHandler::DiagnosticMessageHandler(
+  llvm::raw_ostream& os,
+  clang::DiagnosticOptions* diags,
   SourceManager& srcMgr_,
   std::shared_ptr<odb::database> db_)
-    : _srcMgr(srcMgr_), _db(db_)
+    : TextDiagnosticPrinter(os, diags), _srcMgr(srcMgr_), _db(db_)
```
bruntib, line 15:
```suggestion
    : TextDiagnosticPrinter(llvm::errs(), diags), _srcMgr(srcMgr_), _db(db_)
```
I think the first parameter of this constructor can be removed.
If you remove this, then don't forget to remove `#include <llvm/Support/raw_ostream.h>` from `diagnosticmessagehandler.h`.

PR #451, opened 2020-08-20
Correct syntax highlight after click on word

### LINKED ISSUE #442: Class syntax highlight changes
At first, the classes appear in a turquoise (or whatever that light blue is called) syntax highlight, but when clicking any class name, the highlighting changes to darker blue (like the functions) or black.

Before clicking:
image
image

After clicking:
image
image

intjftw:
Fixes #442 
The problem was `getSyntaxHighlight` was not called when any word was clicked in a file. Calling this method puts the correct highlight on all words.

### DIFF
--- a/webgui/scripts/codecompass/view/component/Text.js
+++ b/webgui/scripts/codecompass/view/component/Text.js
@@ -387,6 +387,13 @@ function (declare, domClass, dom, style, query, topic, ContentPane, Dialog,
       this._marks = {};
     },
 
+    clearSelection : function () {
+      for (var markId in this._marks) {
+        if (this._marks[markId].className === 'cb-marked-select')
+          this.clearMark(markId);
+      }
+    },
+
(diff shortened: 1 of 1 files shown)


intjftw:
As a matter of fact, I think it would be a better approach to make syntax highlighting permanent on the JS side. However, this PR serves as a quick solution.

bruntib:
I don't really understand two things here:
- When loading the source file, `getSyntaxHighlight()` API function is called lazily. In the attached screenshot it seems that it doesn't work. Function names should be bold for example. This turquoise color is the default syntax highlight of CodeMirror.
- Why does a click event disable syntax highlight?

I have a feeling that `getSyntaxHighlight()` is a quite costly API function, because all AST nodes must be queried with their entity types so they can be highlighted properly. We could cache its result at the first time it's called when the source file is loaded (line 525).

bruntib (approved):

PR #457, opened 2020-08-29
Open info tree on left click event

### LINKED ISSUE #455: Show info tree on left click event
From user experience point of view it is important that frequent functionalities are available through as less operations as possible. Info tree is the most frequently used feature, so it should be visible immediately on left click event.

This implies that file tree is automatically hidden. Maybe this is not a too big problem. If file tree is required to be always seen then we may consider splitting accordion panel on the left and displaying file tree permanently besides other tabs.

bruntib:
Info tree is the most frequent functionality in CodeCompass, so this should be
available with the least operations. Now it opens on left click event.

Fixes #455


mcserep (approved):
Tested, works properly.

PR #458, opened 2020-08-29
Download npm packages only if necessary

### LINKED ISSUE #456: Run npm lazily
`make install` command runs `npm` automatically for downloading 3rd party JS tools even if they are downloaded already. This makes installation process unnecessarily slow. `npm` should be run only if `node_modules` directory is not found in `install` folder.

bruntib:
This makes CodeCompass installation process faster.

Fixes #456


mcserep (commented):
What will happen if the NPM dependencies change (e.g. we add something or upgrade versions)?
Then, without manually deleting the `node_modules` directory, an incremented install of CodeCompass will be corrupted.

whisperity:
I agree with @mcserep. We have loads of issues with the CMake script not working properly in incremental development of the web stuff (`GLOB` expressions is one famous example...), we should not add another.

bruntib:
You're right, good point. I changed the solution so `npm` is run only if `packages.json` has been modified.

webgui/InstallGUI.cmake
```diff
@@ -4,9 +4,11 @@ execute_process(
   COMMAND cp ${CC_PACKAGE} ${INSTALL_SCRIPTS_DIR}/package.json
   WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
 
-execute_process(
-  COMMAND npm install
-  WORKING_DIRECTORY ${INSTALL_SCRIPTS_DIR})
+if(${CC_PACKAGE} IS_NEWER_THAN ${INSTALL_SCRIPTS_DIR}/node_modules)
```
mcserep, line 7:
```suggestion
if(${CC_PACKAGE} IS_NEWER_THAN ${INSTALL_SCRIPTS_DIR}/package.json)
```

mcserep (changes requested):
Nice, I did not know about this `IS_NEWER_THAN` CMake comparator.

webgui/InstallGUI.cmake
whisperity: > Behavior is well-defined only for full paths. 

Are these?

bruntib: Yes, these are.

bruntib:
Me neither, I didn't know this operator. I was happy to find it after some false trials :)

mcserep (approved):

Issue #271, opened 2018-07-22
AST dump display should be lazier


bruntib:
The AST dump view in the GUI can consume too much time while loading, and also freezes the browser. The process should be lazy, and maybe not only on client side. We should consider fetching the result by ~500 lines and scrolling should trigger this event.

Although I don't know if it is useful displaying the AST of a whole translation unit. It is barely searchable and usable. I know that there will be a feature of displaying the AST of only a specific element. Until then the current solution is fine, but in long terms I wouldn't suggest to keep it.

mcserep:
#459 closed this issue.

PR #459, opened 2020-08-29
Remove "Show AST html" menu item from files.

### LINKED ISSUE #422: AST dump for files are too big
Right click on a C/C++ source file -> C++ -> Show AST HTML.
This menu item fetches the AST of the whole translation unit. This is almost always so big that it either doesn't arrive from the server or can't be rendered by the browser. In the latter case the browser may consume all memory (including swap) of the client machine, so it freezes the whole OS.
AST query for files should be either eliminated from the GUI, or we should make it lazy: all tree nodes are closed by default.

bruntib:
This menu item displays the AST dump of a translation unit. Even in the
smallest cases this operation is not loaded or unreasonably slow.

Fixes #422

### DIFF
--- a/plugins/cpp_reparse/webgui/js/cppReparseMenu.js
+++ b/plugins/cpp_reparse/webgui/js/cppReparseMenu.js
@@ -44,34 +44,4 @@ function (topic, Menu, MenuItem, PopupMenuItem, model, viewHandler) {
     type     : viewHandler.moduleType.TextContextMenu,
     priority : 10
   });
-
-  //--- Reparse menu for source files in the project manager ---//
-  var fileMenu = {
-    id : 'cppreparse-file',
-    render : function (fileInfo) {
-      if (fileInfo.type !== "CPP")
-        return;
-
-      var submenu = new Menu();
-
-      submenu.addChild(new MenuItem({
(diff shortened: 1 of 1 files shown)


mcserep (approved):
Usually it is slow, because if standard library headers are included, then the AST for the included code parts also have to be generated.

Since this is a typical case, I agree that generating the AST for complete files are not reasonable.

intjftw (approved):

whisperity (changes requested):
The related API endpoint and implementation should also be removed, if we do not need this feature. There is an overload because for files we don't store an ASTNode...

I don't think it was ever meant to be more than a proof-of-concept, anyways.

bruntib:
Maybe we don't need to give up the whole implementation. I'm thinking of a lazy solution: we could implement a dynamic, but still textual tree view. This means that AST dump looks exactly the same in GUI as now except that we add some collapsing triangles which open the next level of the tree. If we cache AST dump for a few minutes in memory on server-side when this feature is used then the TU doesn't need to be recompiled when opening the next sub-tree. The server just cuts the corresponding lines of AST dump and sends it to the client.
We could continue brainstorming in #271.

PR #461, opened 2020-08-30
Verbose shell commands

filbeofITK:
Changed the shell command modifiers to their extended version so the code is more readable.
And somehow I messed up and I can't fix it, but this is only one commit not 35 :(


mcserep:
Don't worry about the number of commits (I guess it is because of the merges), we can squash your changes into a single commit upon merging the PR anyway.

On the other hand, I am a bit unsure whether this level of verbosity is beneficial or not?
Some abbreviations (e.g. `rm -rf`, `apt-get update -y`) are very well known and for me reading e.g. `rm --recursive --force` is not helping, but hindering readability. Other opinions on this?

intjftw:
I agree with @mcserep . Using full words in command options are scarcely used when an abbreviation is available, and makes it harder to read. It is also expected from a dev to look up an option they don't know.

filbeofITK:
If you think it's a bad idea I won't push it. The reason behind it was to help newcommers. Also I thought it takes like half a second at max to read the longer versions, so it would not cause a problem.

On the other hand I can see why experienced programmers prefer shorthand version.

docker/web/Dockerfile
```diff
@@ -43,8 +43,8 @@ ENV CC_UID ${CC_UID}
 ENV TINI_VERSION v0.18.0
 
 # Create user and group for CodeCompass.
-RUN groupadd -r codecompass -g ${CC_GID} && \
-    useradd -r --no-log-init -M -u ${CC_UID} -g codecompass codecompass
+RUN groupadd --system codecompass --gid ${CC_GID} && \
+    useradd --system --no-log-init --no-create-home --uid ${CC_UID} --gid codecompass codecompass
```
whisperity, line 47: I support keeping this change only.

whisperity:
One of the changes is actually beneficial (especially because `adduser` vs. `useradd` is a really common and confusing situation).

The rest of the changes are absolutely bonkers. I don't think many people actually know the long-hand form for those abbreviations, as they appear literally *everywhere* in the short form, being so common.

whisperity:
> The reason behind it was to help newcommers.

It doesn't help. Even newcomers don't **learn** the long-hand forms. Also, it is not up to us to teach basic Linux knowledge to someone who is wanting to work on a Linux project. How come the behaviour of `cd` isn't explained in the guide, in that case? If `rm -rf` isn't trivial knowledge, `cd` shouldn't be either.

> Also I thought it takes like half a second at max to read the longer versions, so it would not cause a problem.

`mkdir --parents` is actually misleading, as giving `-p` also turns off erroring if an already existing directory is to be created again...

scripts/remover.sh
```diff
@@ -1,11 +1,11 @@
...
-FILES=$(ls *.js)
+FILES=$(ls ./*.js)
 
 for FILE in $FILES
 do
-	if [ -f $FILE ]
+	if [ -f "$FILE" ]
 	then
-		sed -i -e "/if (typeof Int64 === 'undefined' && typeof require === 'function') {/,+2 d" $FILE
+		sed -i -e "/if (typeof Int64 === 'undefined' && typeof require === 'function') {/,+2 d" "$FILE"
```
whisperity, line 9: This is a good change for safety if there might be a filename with a space in it, but then it's not related to this patch.

whisperity (changes requested):
-

PR #468, opened 2020-09-18
rename CppType to CppRecord

### LINKED ISSUE #454: Rename CppType to CppRecord in database
Table name `CppType` is deceiving: it stores only record types (class, struct, union) and not all types. For example enums go to a different table.

When generating info tree for the return type of a function we're only searching entries in `CppType`: https://github.com/Ericsson/CodeCompass/blob/6dfc28a890af17947b4e11f58479e637e56eb5ec/plugins/cpp/service/src/cppservice.cpp#L455-L456
Return type of functions returning an enum type is thus not shown in info tree (see #11).

rk717:
Fixes #454

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -26,8 +26,8 @@
 #include <model/cppnamespace-odb.hxx>
 #include <model/cpprelation.h>
 #include <model/cpprelation-odb.hxx>
-#include <model/cpptype.h>
-#include <model/cpptype-odb.hxx>
+#include <model/cpprecord.h>
+#include <model/cpprecord-odb.hxx>
 #include <model/cpptypedef.h>
 #include <model/cpptypedef-odb.hxx>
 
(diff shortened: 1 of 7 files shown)


plugins/cpp/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 add_subdirectory(model)
 add_subdirectory(parser)
 add_subdirectory(service)
-add_subdirectory(test)
+#add_subdirectory(test)
```
bruntib, line 4: Please, don't comment C++ tests out.

bruntib (changes requested):

mcserep (approved):
With @bruntib 's comment applied.

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -62,7 +62,7 @@ namespace parser
  * stack which stores the currently iterated function at the top. This way the
  * parameters and local variables of the function find their parent at the top
  * of the stack. If stack is used then they are pushed and popped in the
- * corresponding Traverse... function.
+ *e corresponding Traverse... function.
```
bruntib, line 65: I think this character addition was accidental.

bruntib (changes requested):

bruntib (approved):

Issue #471, opened 2020-09-24
Parser segfault when parsing CodeCompass


mcserep:
With the current state of the `master` branch I tried to parse CodeCompass itself (the same state of the master branch).

I get the following segfault:

> [INFO] (21/138) Parsing /home/mate/cc/sources/CodeCompass/parser/src/pluginhandler.cpp
> Segmentation fault (core dumped)

I have debugged that the segfault occurs in the `getMangledName()` function, here, when `fileLoc_.file.object_id()` gets evaluated:
```cpp
else if (llvm::isa<clang::NamespaceDecl>(nd_))
{
  const clang::NamespaceDecl* na_ = llvm::dyn_cast<clang::NamespaceDecl>(nd_);
  return na_->isAnonymousNamespace()
    ? "anonymous-ns:" + std::to_string(fileLoc_.file.object_id())
    : nd_->getQualifiedNameAsString();
}
```
https://codecompass.net/demo/#wsid=CodeCompass&fid=5704517689050331186&select=117|1|118|1&center=text

This can be followed to ODB's `lazy_ptr` implementation, where `i_.template object_id<O> ()` yields the segfault:
```cpp
template <class T>
template <class O>
inline typename object_traits<O>::id_type lazy_shared_ptr<T>::
object_id () const
{
  typedef typename object_traits<T>::object_type object_type;

  return p_
    ? object_traits<object_type>::id (*p_)
    : i_.template object_id<O> ();
}
```
https://codecompass.net/demo/#wsid=CodeCompass&fid=12316403250594715644&select=1351|1|1352|1&center=text

Which finally ends up here (with `r` being the culprit):
```cpp
template <typename T>
template <typename O>
inline typename object_traits<O>::id_type lazy_ptr_impl<T>::
object_id () const
{
  typedef typename object_traits<T>::id_type id_type;
  const id_type& id (*static_cast<const id_type*> (id_));

  // Make sure that O' and T's object id types are the same
  // (or implicit-convertible). If you get a compile error
  // pointing here, then you most likely used a wrong type
  // as a template argument in the object_id() call.
  //
  const typename object_traits<O>::id_type& r (id);
  return r;
}
```
https://codecompass.net/demo/#wsid=CodeCompass&fid=9722353701976068959&select=395|1|397|1&center=text

I tried and can parse some other projects, e.g. TinyXML and the first 23 TU's were parsed correctly even for CodeCompass.

intjftw:
[...]

PR #474, opened 2020-10-03
Add CI test for parsing CodeCompass itself

mcserep:
#425 seemingly introduced a bug, where CodeCompass compiles, the tests passes, but more complex projects (like CodeCompass) cannot be parsed, a segfault occurs (see #471).

To prevent such issues in the future, this PR adds a new job to the CI, where the built `CodeCompass_parser` has to parse the CodeCompass project itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -306,13 +306,6 @@ services:
     make install -j $(nproc)
   fi
 
-cache:
-  directories:
-    - $HOME/build2_install
-    - $HOME/gtest_install
-    - $HOME/odb_install
-    - $HOME/thrift_install
-
 # ---------------------------------- Setup -----------------------------------
 
 before_script:
@@ -352,6 +345,7 @@ before_script:
 
 .build_postgres: &build_postgres
   - cd $TRAVIS_BUILD_DIR
+  - rm -rf install_pgsql
   - mkdir build_pgsql && cd build_pgsql
   - >
     cmake ..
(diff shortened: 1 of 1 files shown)


mcserep:
Of course this new CI job will fail currently, as `master` is broken, until #471 gets fixed.

whisperity (commented):
Am I reading it right that the self-parse phase also compiles CodeCompass first? I had issue with synch of content through individual stages, so I want to make it clear that we won't run into such with this.

mcserep:
> Am I reading it right that the self-parse phase also compiles CodeCompass first? I had issue with synch of content through individual stages, so I want to make it clear that we won't run into such with this.

Initially YES, since [caching in Travis CI](https://docs.travis-ci.com/user/caching/#caches-and-build-matrices) does not support cache keys (like GitLab CI or GitHub Actions), so we cannot cache the CodeCompass binaries just for a single job.  
So if we cache the built CodeCompass binaries, an extra effort is required to clean that part of the cache in the compile jobs, otherwise not a clean install would be performed.

I will give a try, because of course it could speed up that CodeCompass parse job a little.

mcserep:
It works with caching the built CodeCompass in the previous stage. (Well, it fails as it should right now.)

whisperity (approved):
Alright, it is convincing enough to me. I do have a *slight* fear it won't be robust enough, but most likely that comes from the previous bad experience. Hopefully explicitly wiping the cached dirs when it is recreated is enough.

PR #477, opened 2020-10-13
Revert "Better AST value for nodes."

mcserep:
#425 resulted in critical error for the parser component, as CodeCompass cannot even parse itself (see #471).
Implementing USR (#429) instead of mangled names will probably fix this issue (as it eliminates the erroneous  `getMangledName()` method completely), but that PR also fails for segmentation fault currently.

As the *master* branch contains such a serious issue for several weeks now and the immediate solution cannot be seen yet, I suggest to revert #425, as it introduced the bug in the first place.


zporky:
I agree to revert this change but only after testing that the reverted code is working fine (and CodeCompass can compile themselves).
We still have to understand the actual chain of events leading to the crash.

mcserep:
> I agree to revert this change but only after testing that the reverted code is working fine (and CodeCompass can compile themselves).

For me CodeCompass can parse itself with this modification reverted.
In the CI this test fails unfortunately, because we run out of disk space. (All system dependencies + compiled CodeCompass + parse > 24GB as it seems like). I will try to find a solution for this, as it would be nice to see the CI pass before merging this PR.

whisperity:
> `> 24 GiB`

Wait.. what... HOW?!
Did we start parsing the LLVM dependencies, too?

mcserep:
> > `> 24 GiB`
> 
> Wait.. what... HOW?!
> Did we start parsing the LLVM dependencies, too?

On second check we do not have 24GB free space, because that only applies to the `minimal` environment and we use `generic`.
https://docs.travis-ci.com/user/common-build-problems/#im-running-out-of-disk-space-in-my-build

I will try out the `minimal` image maybe to gain more initial free space.

mcserep:
With the `minimal` image there is no issue, the CodeCompass self-parse test passes.
So in conclusion, the 24GB space is enough @whisperity. It turns out the `generic` image contained quite a lot of preinstalled software (see  #479), thus the available free space was more restricted.

mcserep:
Original issue #471 solved by #429 , so closing this. Improvements on CI have been moved to separate PR (#479).

PR #479, opened 2020-10-15
Travis CI: Switch to minimal image

mcserep:
Switch to the `minimal` image in Travis CI from `generic`.
See differences here: https://docs.travis-ci.com/user/languages/minimal-and-generic/

The `generic` image has many software preinstalled, e.g.:
 - various database services and tools (Apache Cassandra, CouchDB, ElasticSearch, MongoDB, MySQL, Neo4j, PostgreSQL, RabbitMQ, Redis, Riak, SQLite);
- go (multiple versions);
- jvm (multiple versions);
- nodejs (multiple versions);
- php (multiple versions);
- ruby (multiple versions).

We do not need them and they take up a lot of disk space. Therefore we ran out of disk space when performing a self-parse test on CodeCompass. On the `minimal` image only essential tools (e.g. build tools, version control tools) are included and we have approx. 24GB free disk space which shall be enough.

Additional modifications with this PR related to CI:
- Only install required Boost components (as in the Docker image, also saving disk space and making the CI jobs a little bit faster).
- Manual installation of PostgreSQL server as it is no longer bundled in the image.
- Build and install CodeCompass in $HOME instead of $TRAVIS_BUILD_DIR, so the build and install directories won't be parsed on the self-parse test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 os: linux
 dist: bionic
-language: generic
+language: minimal
 services:
   - postgresql
 
@@ -17,7 +17,10 @@ services:
   <<: *apt_xenial_config
   packages:
     # Boost
-    - libboost-all-dev
+    - libboost-filesystem-dev
+    - libboost-log-dev
+    - libboost-program-options-dev
+    - libboost-regex-dev
     # LLVM/clang
     - llvm-10
     - llvm-10-dev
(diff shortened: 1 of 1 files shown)


bruntib:
If you would rebase this patch to master (since #429 has been merged) then we can see if the segfaul has really been fixed. I'm excited to see the result :)

mcserep:
> If you would rebase this patch to master (since #429 has been merged) then we can see if the segfaul has really been fixed. I'm excited to see the result :)

I have applied these modifications to #477, the CI works there. I have rebased this PR now to the new master including USR, let's what happens :)

mcserep:
It works! 🏆 

I don't understand why it works now and not before. Maybe it was an out of disk space issue after all, but the error message was misleading? (Segmentation fault instead of out of disk space.)

I will get this merged.

PR #480, opened 2020-10-16
Test parsing of TinyXML2 and Xerces-C++ projects in the CI

mcserep:


mcserep:
Parsing Xerces-C++ fails, because we really have an issue with it currently (most likely since #429).
Probably we shall merge this PR after we managed to find out what is the issue with it, but feel free to review it meanwhile. (Not a big change.)

intjftw:
I like the idea of having multiple actual test projects in the CI, I think this is useful independent of the current random segfaults. I think we should try parsing on all currently supported Ubuntu versions if possible.

intjftw (approved):

mcserep:
Rebased upon new `master`, following the merge of #481 .

mcserep:
It was a long way to debug this, but for some reason we have to create a PostgreSQL user aside the default `postgres`, otherwise parsing Xerces-C++ yields an UTF8 decoding error. While it is true that `XMLFloat.cpp` contains a non-UTF8 char, the encoding of the database is `SQL_ASCII`, therefore it should not be a problem.

However it seems that it is a problem with the `postgres` user, but no with another user. Strange as it is, but since the `postgres` superuser shall not be used anyway in a production system, I will not examine this issue further.

PR #484, opened 2020-10-24
Set RelWithDebInfo as default build type

mcserep:
Support for **`RelWithDebInfo`** and **`MinSizeRel`** build types. Default to **`RelWithDebInfo`** if not specified otherwise.


intjftw (approved):
Works fine.

bruntib (approved):

Issue #476, opened 2020-10-13
SQL Indexe commands crash when running CodeCompass_parser 


shijunjing:
Hello,
I meet a SQL commands running crash when use CodeCompass_parser to setup the PostgreSQL database for my project. Below is the crash log:
```
~/CodeCompass/install/bin/CodeCompass_parser -d "pgsql:host=localhost;port=5433;user=compass;password=123456;database=mydatabase" -w /home/jshi19/CodeCompass/workspace -n edk2 -i /home/jshi19/wksp_efi/edk2-2/build_commands.json -i /home/jshi19/wksp_efi/edk2-2 -j 16 
... ...
[INFO] Recursive directory iteration: visited 45681 files in 3922 directories so far.
[INFO] Indexed 22,577 file(s)
[INFO] Modified 0 file(s)
[INFO] Indexer finished!
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/cppenum-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/cppfriendship-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/buildlog-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/metrics-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/cpprecord-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/filecontent-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/statistics-odb.sql
[INFO] Creating indexes from file /home/jshi19/CodeCompass/install/share/codecompass/sql/cpptype-odb.sql
[WARNING] Exception when running SQL command: 42P07: ERROR:  relation "CppMemberType_memberAstNode_i" already exists

Exception when running SQL command: 42P07: ERROR:  relation "CppMemberType_memberAstNode_i" already exists
```

After quickly search the "CppMemberType_memberAstNode_i", I found there are two installed sql files containing the same "CppMemberType_memberAstNode_i" index creation SQL commands as below.

- CodeCompass\install\share\codecompass\sql\cpprecord-odb.sql
- CodeCompass\install\share\codecompass\sql\cpptype-odb.sql

```
[...]

PR #486, opened 2020-10-25
Do not check for virtual tag on FunctionPtr definitions

mcserep:
As reported in #476, clicking a function pointer call results in a web server crash with a segmentation fault.
Example code (here `OpenProtocol` is a public data member of `gBS`, a function pointer):

The cause of the segmentation fault is that in `CppServiceHandler::getTags()` we assumed that if the observed AST node had the `symbolType` *Function*, then its definition node should also have the `symbolType` *Function*. However, the definition node can have the `symbolType` *FunctionPtr*; in such case the `funcNodes` variable will be empty, thus dereferencing its first element is UB.
https://github.com/Ericsson/CodeCompass/blob/173a3cf15e15be6de38b476fa7ff5eb258829e8c/plugins/cpp/service/src/cppservice.cpp#L1526-L1533

This PR adds a check on the definition node `symbolType`, so the segmentation fault is avoided.

### DIFF
--- a/plugins/cpp/service/src/cppservice.cpp
+++ b/plugins/cpp/service/src/cppservice.cpp
@@ -1523,14 +1523,17 @@ CppServiceHandler::getTags(const std::vector<model::CppAstNode>& nodes_)
             tags[node.id].push_back(visibility);
         }
 
-        //--- Virtual Tag ---//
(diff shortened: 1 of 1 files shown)


mcserep:
Can we merge this, did you have time to take a look at it?

intjftw (approved):

PR #489, opened 2020-11-10
change invalid connectionstring to valid

matebelenyesi:
Characters are replaced to underscore in connection string which are not in [a-zA-Z0-9_] set.

### DIFF
--- a/util/src/dbutil.cpp
+++ b/util/src/dbutil.cpp
@@ -172,7 +172,7 @@ bool checkPsqlDatbase(
 
     if (!exists && create_)
     {
-      std::string createCmd = "CREATE DATABASE " + dbName_
+      std::string createCmd = "CREATE DATABASE \"" + dbName_ + "\""
         + " ENCODING = 'SQL_ASCII'"
         + " LC_CTYPE='C'"
         + " LC_COLLATE='C'"


intjftw (changes requested):
The changes should be put into _util/src/dbutil.cpp_

mcserep (changes requested):
I have checked and ODB automatically convert identifiers to *quoted identifier* correctly. (@bruntib already made a remark on one of our previous meetings that there is no issue with our database table names containing uppercase letters. Only the database name raises an error.)

> There is a second kind of identifier: the *delimited identifier* or *quoted identifier*. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). A delimited identifier is always an identifier, never a key word. So "select" could be used to refer to a column or table named “select”, whereas an unquoted select would be taken as a key word and would therefore provoke a parse error when used where a table or column name is expected. The example can be written with quoted identifiers like this:
> 
> ```
> UPDATE "my_table" SET "a" = 5;
> ```

The issue is with the database creation, which we perform through a native SQL script. Here, the database name is not enclosed between double-quotes:
https://github.com/Ericsson/CodeCompass/blob/1097e5ce0519278dffe44c97e57ed4120d96e2e7/util/src/dbutil.cpp#L175-L179

Fixing it like this solves the original issue:
```cpp
std::string createCmd = "CREATE DATABASE \"" + dbName_ + "\""
  + " ENCODING = 'SQL_ASCII'"
  + " LC_CTYPE='C'"
  + " LC_COLLATE='C'"
  + " TEMPLATE template0;";
```

mcserep (approved):
Seems good. The commits shall be squashed before merge.

mcserep:
@matebelenyesi I advise do not create a pull request from the *master* branch of your fork, instead create a new branch.
Now there will be a conflict between Ericsson's *master* and your *master*, and basically you will have to force overwrite your own *master* branch to get them in sync again.

PR #492, opened 2020-11-20
Bugfix for forced full parse

intjftw:
In case the user wanted to incrementally parse a project but a full parse was forced due to the large number of changed files, the database was not dropped and recreated which caused database errors. This fix aims to recreate the database tables in case of a forced full parse.

### DIFF
--- a/parser/src/parser.cpp
+++ b/parser/src/parser.cpp
@@ -378,6 +378,7 @@ int main(int argc, char* argv[])
     cc::util::createTables(db, SQL_DIR);
 
     srcMgr.cacheFiles();
+    ctx.fileStatus.clear();
   }
 
   if (!vm.count("force"))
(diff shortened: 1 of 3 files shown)


parser/src/parser.cpp
```diff
@@ -373,6 +373,11 @@ int main(int argc, char* argv[])
     LOG(info) << "The number of changed files exceeds the given incremental "
                  "threshold ratio, full parse will be forced.";
     vm.insert(std::make_pair("force", po::variable_value()));
+
+    cc::util::removeTables(db, SQL_DIR);
+    cc::util::createTables(db, SQL_DIR);
+
+    srcMgr.cacheFiles();
```
mcserep, line 380: The `ParserContext ctx` also has a `fileStatus` cache, containing the incremental status of all files.
Probably it is not used further in case of a "clean" parse, but still should be cleared.

mcserep (changes requested):

mcserep (approved):

PR #493, opened 2020-11-20
Incremental parsing of projects with circular dependencies

intjftw:
At the cleanup phase of incremental parsing, files to be cleaned up are organized into levels (topological order). Trying to make the topological order in a project that contains circular dependencies in the to-be-cleaned up files caused an infinite loop. This PR fixes this issue by putting the files with the circular dependencies into an additional level in the topological order thus the loop is guaranteed to terminate.


mcserep (changes requested):
Since the cleanup of circular dependencies are not guaranteed to be fail-safe (or at least I am not sure), we shall add a log message (maybe just a debug level), so in case the cleanup fails, we would have this information.

mcserep (approved):

PR #497, opened 2020-12-06
added quotation marks to databasename (squashed)

matebelenyesi:
change convert function

move connection string modifying functions to dbutil

smaller typo and formatting fixes

change connenctionstring only with pgsql 

change back everything and adding quotation marks


mcserep:
How is it different from the already merged #489?

matebelenyesi:
Sorry, i didn't realize it's already merged. I created this request from a different branch than master. But then i undo it.

PR #498, opened 2020-12-09
Checking if the result of a query or query_one function is not empty or nullptr

matebelenyesi:


whisperity:
Please rebase and squash the commits there's a back-and-forth diff added.

mcserep:
Please add a short description what this PR is about. We know about it now, but documentation is required so we will know about it later.

matebelenyesi:
I close this pull request, and open an other with the same changes, because i accidently merged not related commits here.

mcserep:
You can also force-push the `querycheck` branch on your fork, in that case you can "reuse" this PR.

PR #508, opened 2021-01-18
Parsing test projects in CI

intjftw:
I continued Github Actions CI migration with parsing Xerces-C and TinyXML on both Ubuntu 18.04 and 20.04, with Postgresql and SQLite3 as well. Also complemented building CodeCompass with caching ODB and Thrift on 18.04. I reorganized CI commands to group by OS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: Build project
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 env:
   BUILD_TYPE: Release
@@ -64,8 +64,16 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-18.04' && matrix.db == 'sqlite3' }}
         run: sudo apt-get install libsqlite3-dev
 
-      - name: Install Thrift
+      - name: Cache Thrift Ubuntu 18
+        id: thrift-cache
         if: ${{ matrix.os == 'ubuntu-18.04' }}
+        uses: actions/cache@v2
+        with:
+          path: ~/thrift_install
+          key: ${{ matrix.db }}-thrift-v2
+
+      - name: Install Thrift
+        if: ${{ matrix.os == 'ubuntu-18.04' && steps.thrift-cache.outputs.cache-hit != 'true' }}
         run: |
           cd $HOME
           wget -O thrift-0.13.0.tar.gz "http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.13.0/thrift-0.13.0.tar.gz"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -95,9 +98,29 @@ jobs:
...
+        run: |
+          ls $HOME/odb_install/bin
+          ls $HOME/thrift_install/bin
+          echo "CMAKE_PREFIX_PATH=$HOME/thrift_install:$HOME/odb_install:$CMAKE_PREFIX_PATH" >> $GITHUB_ENV
+          echo "PATH=$HOME/thrift_install/bin:$HOME/odb_install/bin:$PATH" >> $GITHUB_ENV
+
+      # Ubuntu 20.04 commands
+      - name: Install required packages Ubuntu 20
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: sudo apt-get install -y git cmake make g++ libboost-all-dev llvm-10-dev clang-10 libclang-10-dev odb libodb-dev thrift-compiler libthrift-dev default-jdk libssl-dev libgraphviz-dev libmagic-dev libgit2-dev ctags libgtest-dev npm
```
mcserep, line 113: Probably split long lines into multiple lines for easier readability:
https://stackoverflow.com/questions/59954185/github-action-split-long-command-into-multiple-lines

.github/workflows/ci.yml
```diff
@@ -139,4 +162,142 @@ jobs:
...
+      - name: Export environment variables Ubuntu 18
+        if: ${{ matrix.os == 'ubuntu-18.04' }}
+        run: |
+          echo "LD_LIBRARY_PATH=$HOME/thrift_install/lib:$HOME/odb_install/lib:$CMAKE_PREFIX_PATH" >> $GITHUB_ENV
+          echo "PATH=$HOME/thrift_install/bin:$HOME/odb_install/bin:$PATH" >> $GITHUB_ENV
+
+      # Ubuntu 20.04 commands
+      - name: Install required packages Ubuntu 20
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: sudo apt-get install -y git cmake make g++ libboost-all-dev llvm-10-dev clang-10 libclang-10-dev odb libodb-dev thrift-compiler libthrift-dev default-jdk libssl-dev libgraphviz-dev libmagic-dev libgit2-dev ctags libgtest-dev npm
```
mcserep, line 240: Do we need all things here? The runtime dependencies of CodeCompass are not as extensive as the build type dependencies:
https://github.com/Ericsson/CodeCompass/blob/master/docker/runtime/Dockerfile#L44-L58

mcserep (changes requested):
Was the CodeCompass parse check omitted on purpose from Travis CI?

.github/workflows/ci.yml
intjftw: I tried this solution from time to time, it never worked. I tried it yesterday as well, still no good results. Instead I chose to use the basic yaml syntax to split lines. (>)

.github/workflows/ci.yml
intjftw: I erased some unnecessary dependencies.

intjftw:
I included parsing of CodeCompass which was intentionally left out in the first place. Since building CodeCompass and parsing the test projects are two separate jobs, the entire CodeCompass project would have to be either built again in the parsing job which makes having a separate job just for building redundant, or archive the project with its build directory as an artifact. I tried this latter solution, however, this extended the jobs' runtime significantly. This is why I decided to parse CodeCompass in the first job so I don't have to archive it and download from the second one.

.github/workflows/ci.yml
mcserep: I still see a bunch of unnecessary development packages, which are not required at runtime, based on the corresponding Docker environment:
https://github.com/Ericsson/CodeCompass/blob/master/docker/runtime/Dockerfile#L57

I suggest the following:
- Replace `libboost-all-dev`
With `libboost-filesystem-dev libboost-log-dev libboost-program-options-dev `

- Replace `llvm-10-dev clang-10 libclang-10-dev`
With `llvm-10`

- Replace `default-jdk`
With `default-jre`

- Replace `libssl-dev`
With `libssl1.1`

- Add `libgvc6` and `libmagic-dev` for GraphViz and magic.

mcserep (changes requested):
Please move CodeCompass parsing to second stage.

mcserep (approved):
Good work, looks great! :rocket:

PR #510, opened 2021-02-05
Local testing containers

filbeofITK:
I made this, because I felt like we lack the ability to test locally properly, since every build has to be tested for multiple Ubuntu versions and our current `codecompass:dev` container only work for 20.04. This image can be built with multiple Ubuntu  versions, allowing us to test our build in both environments locally. Also it opens up the possibility to develop CodeCompass to any linux operating system later.

### DIFF
--- a/docker/test/Dockerfile
+++ b/docker/test/Dockerfile
@@ -1,0 +1,45 @@
+ARG UBUNTU_VERSION=20.04
+
+FROM ubuntu:${UBUNTU_VERSION}
+
+ARG UBUNTU_VERSION
+
+# tzdata package is installed implicitly in the following command. This package
+# sets timezone interactively during the installation process. This environment
+# variable prevents this interaction.
+ARG DEBIAN_FRONTEND=noninteractive
+
+
+#Setting user and group properties
+ARG CC_GID=960
+ARG CC_UID=960
+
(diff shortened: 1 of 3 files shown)


mcserep (commented):
I only remember a single case in the past few years of CodeCompass development, where a bug could be only reproduced in one of the supported OSes. It is extremely rare, probably because we support only the Ubuntu distro officially.

Therefore, in my point of view, the Docker images should serve the purpose of creating a unified development environment for us; and a production-level runtime environment for end-users. Because of this, Docker images should be built on a single selected OS version. (Speaking about our official images, of course anyone can create a custom image.)

Continuous testing of all supported OS versions cannot be expected from the developers, as it would severely hinder the development process. It is the job of the CI to perform the testing, completely automatized, the best way it can, on all supported OSes. Of course, thorough manual testing should be done before major releases and regularly by a testing team (if we would have one :smile: ), but that is another topic.

In my opinion, maintaining yet another CodeCompass build script (beside the documentation, the GitHub Actions and the GitLab CI for tarball creation) has more disadvantages than benefits, as these testing images would be very rarely used.

mcserep:
Summoning a few other team members to discuss this.

mcserep:
Closing without merging, as there were no other comments on this PR in 3 months.

PR #512, opened 2021-02-05
Displaying build log messages when clicking on the parse status label.

intjftw:
We talked some time ago about not being able to see build logs from the webservice. In this PR, I connected the endpoint that queries build logs of a file with the onclick event of the parse status label in the header of an open file.


mcserep (approved):
Done some modifications in 0858b069a925b505d39c1d4ebd56c6932f61fcf9 to fix JS errors.

PR #515, opened 2021-02-18
Skip circular symlinks

intjftw:
Circular symlinks cause every parser to throw an exception and fail. I complemented `parserutil.cpp` with local exception handling to skip the problematic file and avoid failure of the entire parser.

### DIFF
--- a/util/src/parserutil.cpp
+++ b/util/src/parserutil.cpp
@@ -50,17 +50,17 @@ bool iterateDirectoryRecursive(
 
   fs::path p(path_);
 
-  try
+  boost::system::error_code ec;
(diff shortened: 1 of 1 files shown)


mcserep (approved):

intjftw:
I asked about it on Stack Overflow, and got an easy solution for the problem. 
[solution](https://stackoverflow.com/questions/66275694/find-circular-symlinks-with-boost-filesystem/66279856#66279856)
Its focus isn't symlinks specifically but other such failures as well. I rewrote my previous exception handling solution to this.

mcserep (approved):

PR #516, opened 2021-02-20
Fix C++ parser relative path

bruntib:
C++ parser is working based on a compilation_database.json file.
This file may contain relative file paths in its "file" attributes.
This file is given to C++ parser to build the AST of this file.
However, this works only if the filename is correctly relative to
the current working directory.
In this patch the directory and file components are joined so the
full path of a source file is passed to the parser.


intjftw (approved):
Looks good to me!

mcserep (approved):

PR #520, opened 2021-03-09
Track file browsing with Google Analytics

mcserep:
Currently only a single page view event is sent to *Google Analytics* (**if enabled**) when opening a new project in the web browser. Since browsing in the File Manager and opening files are done asynchronously in the background, they do not trigger a page reload. Only selecting a new project in the web browser triggers a page reload.

The aim of this PR is to send a `page_view` event to Google Analytics when the content of a new file is loaded.
That way we could get much more precise diagnostic about which project and files are visited most on a CodeCompass instance (e.g. http://codecompass.net/demo/)

### DIFF
--- a/webgui/scripts/codecompass/view/component/Text.js
+++ b/webgui/scripts/codecompass/view/component/Text.js
@@ -342,6 +342,14 @@ function (declare, domClass, dom, style, query, topic, ContentPane, Dialog,
       this.set('header', this._fileInfo);
 
       this._getSyntaxHighlight(this._fileInfo);
+
+      if (gtag) {
+        gtag('event', 'page_view', {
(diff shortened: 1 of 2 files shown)


intjftw (approved):
Great job! It would be really nice to track actual activities in the future as well, if there is a possibility for that.

mcserep:
> Great job! It would be really nice to track actual activities in the future as well, if there is a possibility for that.

Yes, it is possible to track custom defined events with GA. Some documentation on it:
 - [Measure Google Analytics Events](https://developers.google.com/analytics/devguides/collection/gtagjs/events)
 - Maybe also: [App / Screen Measurement](https://developers.google.com/analytics/devguides/collection/gtagjs/screens)

PR #523, opened 2021-03-10
Fixes for CppParserTest

### LINKED ISSUE #439: Unit test failure in debug mode
The `CppParserTest` fails when run in debug mode with the following error message:
```cpp
2: [==========] Running 10 tests from 1 test case.
2: [----------] Global test environment set-up.
2: [----------] 10 tests from CppParserTest
2: [ RUN      ] CppParserTest.FilesAreInDatabase
2: [       OK ] CppParserTest.FilesAreInDatabase (41 ms)
2: [ RUN      ] CppParserTest.SimpleFunction
2: [       OK ] CppParserTest.SimpleFunction (6 ms)
2: [ RUN      ] CppParserTest.FunctionDeclarationOnly
2: [       OK ] CppParserTest.FunctionDeclarationOnly (2 ms)
2: [ RUN      ] CppParserTest.FunctionWithMultipleDeclarations
2: cppparsertest: /usr/local/include/odb/result.txx:50: void odb::result<T>::value(T&) [with T = cc::model::CppFunction]: Assertion `++i == end ()' failed.
2/2 Test #2: cppparser ........................***Exception: Child aborted  3.56 sec
```
The error is not present when the tests are run in release mode.

mcserep:
Closes #439.

This PR fixes the strange issue that the unit tests failed when CodeCompass was compiled in *Debug* mode.
It turned out that these tests were really failing (for a long time actually), but when compiled *Release* mode, some ODB functions just fail silently and assertions are only done in *Debug* mode.

In most cases, the root of these issues was the `query_value()` function, which returns the selected record from the database *as a value*. The number of resulted records *must be precisely one*, otherwise assertion failure is done in debug mode. In release mode no error is raised at all: when multiple results are available, the first one is returned and in case of zero results, a default constructed object is returned (or even memory garbage, I haven't checked it).

As the [ODB Manual](https://www.codesynthesis.com/products/odb/doc/manual.xhtml#4.3) states:
> If the query executed using `query_one()` or `query_value()` returns more than one element, then these functions fail with an assertion. Additionally, `query_value()` also fails with an assertion if the query returned no elements.

These misbehaviours went unnoticed in release mode, because when multiple results were available, simply the first was returned; and when no results were available, all `EQ_*` assertions were inside a loop, which was executed 0 times.

In this PR I have addressed these issues and fixed them. In all cases I am confident that the unit tests were incorrect and CodeCompass worked correctly. There is only a single case where I am uncertain (related to function pointers), but I will mark that case for discussion explicitly soon.

I have also modified that the CI shall build CodeCompass in *Debug* mode for testing in the future, to avoid any such issue like this in the future. (Docker images and tarballs are still built in *Release* mode of course.)

### DIFF
--- a/plugins/cpp/test/src/cppparsertest.cpp
+++ b/plugins/cpp/test/src/cppparsertest.cpp
@@ -142,7 +142,7 @@ TEST_F(CppParserTest, FunctionWithMultipleDeclarations)
   _transaction([&, this]() {
     RCppFunction callees = _db->query<model::CppFunction>(
       QCppFunction::name == "multiFunction");
-    EXPECT_GT(callees.size(), 0);
+    EXPECT_FALSE(callees.empty());
 
     RCppAstNode multiFuncAstNode = _db->query<model::CppAstNode>(
       QCppAstNode::entityHash == callees.begin()->entityHash);
@@ -307,7 +307,7 @@ TEST_F(CppParserTest, Typedef)
       QCppTypedef::name == "Integer");
     RCppAstNode astNodes = _db->query<model::CppAstNode>(
       QCppAstNode::entityHash == integer.entityHash);
-    EXPECT_GT(astNodes.size(), 0);
+    EXPECT_FALSE(astNodes.empty());
 
     for (const model::CppAstNode& n : astNodes)
     {
@@ -461,11 +461,11 @@ TEST_F(CppParserTest, Enum)
   {
     RCppEnum enumerations = _db->query<model::CppEnum>(
       QCppEnum::name == "Enumeration");
-    EXPECT_GT(enumerations.size(), 0);
+    EXPECT_FALSE(enumerations.empty());
 
     RCppAstNode astNodes = _db->query<model::CppAstNode>(
       QCppAstNode::entityHash == enumerations.begin()->entityHash);
-    EXPECT_GT(astNodes.size(), 0);
+    EXPECT_FALSE(astNodes.empty());
 
     for (const model::CppAstNode& n : astNodes)
     {
--- a/plugins/cpp/test/src/cppparsertest.cpp
+++ b/plugins/cpp/test/src/cppparsertest.cpp
@@ -13,6 +13,8 @@
 #include <model/cppnamespace-odb.hxx>
 #include <model/cpprecord.h>
 #include <model/cpprecord-odb.hxx>
+#include <model/cpptypedef.h>
+#include <model/cpptypedef-odb.hxx>
 #include <model/cppvariable.h>
 #include <model/cppvariable-odb.hxx>
 #include <model/file.h>
@@ -31,8 +33,11 @@ using QCppEnum = odb::query<model::CppEnum>;
 using QCppEnumConstant = odb::query<model::CppEnumConstant>;
 using QCppNamespace = odb::query<model::CppNamespace>;
 using QCppRecord = odb::query<model::CppRecord>;
+using QCppTypedef = odb::query<model::CppTypedef>;
 using QCppVariable = odb::query<model::CppVariable>;
 using QFile = odb::query<model::File>;
+using RCppEnum = odb::result<model::CppEnum>;
+using RCppFunction = odb::result<model::CppFunction>;
 using RCppAstNode = odb::result<model::CppAstNode>;
 
 class CppParserTest : public ::testing::Test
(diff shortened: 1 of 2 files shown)


plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -135,10 +140,12 @@ TEST_F(CppParserTest, FunctionDeclarationOnly)
 TEST_F(CppParserTest, FunctionWithMultipleDeclarations)
 {
   _transaction([&, this]() {
-    model::CppFunction callee = _db->query_value<model::CppFunction>(
+    RCppFunction callees = _db->query<model::CppFunction>(
       QCppFunction::name == "multiFunction");
+    EXPECT_GT(callees.size(), 0);
```
mcserep, line 145: Here the `multiFunction` is declared 2 times and defined 1 time, so all together 3 records in `CppFunction` will be found.

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -296,10 +303,11 @@ TEST_F(CppParserTest, FunctionCall)
 TEST_F(CppParserTest, Typedef)
 {
   _transaction([&, this]() {
-    model::CppEnum integer = _db->query_value<model::CppEnum>(
-      QCppEnum::name == "Integer");
+    model::CppTypedef integer = _db->query_value<model::CppTypedef>(
+      QCppTypedef::name == "Integer");
```
mcserep, line 307: Here `Integer` was not an enum, but a typedef in the parsed source code.

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -526,10 +538,15 @@ TEST_F(CppParserTest, Enum)
           break;
       }
     }
+  });
+}
 
-    model::CppFunction fieldFunction = _db->query_value<model::CppFunction>(
+TEST_F(CppParserTest, Fields)
```
mcserep, line 544: Some tests cases were put in the `CppParserTest::Enums` test case, but they were really checking class fields. So I created this new  `CppParserTest::Fields` test case for them.

plugins/cpp/test/src/cppparsertest.cpp
mcserep, line 145: Here the multiFunction is declared 2 times and defined 1 time, so all together 3 records in CppFunction will be found.

plugins/cpp/test/src/cppparsertest.cpp
mcserep, line 307: Here `Integer` was not an enum, but a typedef in the parsed source code.

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -449,11 +457,15 @@ TEST_F(CppParserTest, Record)
 
 TEST_F(CppParserTest, Enum)
 {
-  _transaction([&, this] {
-    model::CppEnum enumeration = _db->query_value<model::CppEnum>(
+  _transaction([&, this]
+  {
+    RCppEnum enumerations = _db->query<model::CppEnum>(
       QCppEnum::name == "Enumeration");
+    EXPECT_GT(enumerations.size(), 0);
```
mcserep, line 464: There are multiple records for `Enumeration` in the `CppEnum` table.

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -526,10 +538,15 @@ TEST_F(CppParserTest, Enum)
...
+}
 
-    model::CppFunction fieldFunction = _db->query_value<model::CppFunction>(
+TEST_F(CppParserTest, Fields)
+{
+  _transaction([&, this] {
+    model::CppVariable fieldFunction = _db->query_value<model::CppVariable>(
       QCppFunction::name == "fieldFunction");
-    astNodes = _db->query<model::CppAstNode>(
+    RCppAstNode astNodes = _db->query<model::CppAstNode>(
```
mcserep, line 549: This is the case where I was uncertain about my fix. This `fieldFunction` is a defined like this:

```cpp
struct Derived : public MyClass
{
  MyClass fieldVariable;
  MyClass (*fieldFunction)(const MyClass&);
  
  // etc...
};
```

So it is a class member field, as a function pointer. It gets parsed into the `CppVariable` table, not the `CppFunction`, but that is correct that way, right @bruntib ? It is a variable, not a function semantically.

intjftw: I checked the above code, the function pointer is considered a `fieldDecl`.
image

intjftw (approved):

PR #524, opened 2021-03-23
Only the desired sql engine is installed into the container.

filbeofITK:
This way we reduce the runtime image's space. The minimal Readme i constructed clears up usage.
Also I don't know how  this image has worked before because an `apt-get update` was missing from before the git installation, so I included that too.


mcserep:
> Also I don't know how this image has worked before because an `apt-get update` was missing from before the git installation, so I included that too.

I guess it worked because the runtime image depends on the `codecompass:dev` image, but doing a fresh `apt-get update` makes it more independent.

docker/runtime/Dockerfile
```diff
@@ -13,7 +13,8 @@ ENV CC_DATABASE ${CC_DATABASE}
 ARG CC_BUILD_TYPE=Release
 ENV CC_BUILD_TYPE ${CC_BUILD_TYPE}
 
-RUN apt-get install -y git
+RUN apt-get update && \
```
mcserep, line 16: The `-qq` flag is missing to be consistent with other updates in Dockerfiles.

docker/runtime/Dockerfile
```diff
@@ -2,7 +2,7 @@
 #-----------------------------    BUILD STAGE   ------------------------------#
 ###############################################################################
 
-FROM codecompass:dev as builder
+FROM modelcpp/codecompass:dev as builder
```
mcserep, line 5: Why changing this to the DockerHub tag name? If someone is building the runtime images, I think that he/she would build the dev image, too.

mcserep (changes requested):
Good point, only some minor things.

docker/runtime/Dockerfile
filbeofITK: This ensures that the official runtime images is always built from the official developer image, and can not be mixups with local containers while building.

mcserep: That could raise issues, because if a user would like to build custom images, then building the runtime image would not depend on the locally built dev image. Also, with this modification the walkthrough in `docker/README.md` would not work.

filbeofITK: That is fair. I'm reverting this.

mcserep (approved):

PR #525, opened 2021-04-05
Remove mentions and guides for Ubuntu 16.04 "Xenial Xerus"

whisperity:
Ubuntu 16.04 has ended its 5-year long standard support lifetime in April.


.travis.yml
```diff
@@ -6,88 +6,6 @@ services:
 
 # ---------------------------- APT configuration -----------------------------
 
-.apt_xenial_config: &apt_xenial_config
```
whisperity, line 9: Shouldn't we just delete the entire `.travis.yml` file given that we no longer use Travis?

.gitlab/ci.yml
```diff
@@ -100,13 +100,13 @@ tarball suse-42.1:
         # show GLIBC verison
         - *show-glibc-version
 
-tarball ubuntu-16.04:
+tarball ubuntu-18.04:
```
mcserep, line 103: Tarballs were explicitly created for supporting older OS versions to be used internally in Ericsson.
I know for sure that we are using `suse-42.1`, I am not sure about `ubuntu-16.04`.

If we are using this tarball, we shall not update it. If we are not using it, maybe just dropping it would be fine.
I recommend to not modify the`.gitlab/ci.yml` file for now.

.travis.yml
mcserep: I agree. We can drop this file, as we have switched to GitHub Actions, so the Travis build script will gradually get out of sync. If ever needed, it will be available in the version history anyway.

mcserep (changes requested):

mcserep (approved):
Looks good, let's merge it! 🚀

PR #526, opened 2021-04-11
Google Analytics event tracking

intjftw:
I added a couple more event trackers to collect data about the main functionalities and send them to Google Analytics, if it is enabled. The following requests are recorded:

- diagrams
- showing AST
- documentation
- git blame
- clicking on node
- type and query of search
- code bites

The type of request, the name of the project, file and/or node are recorded by the event trackers.

### DIFF
--- a/plugins/search/webgui/js/searchResult.js
+++ b/plugins/search/webgui/js/searchResult.js
@@ -71,8 +71,8 @@ function (ObjectStoreModel, BorderContainer, declare, Memory, Observable, topic,
           if (that._currentQueryData) {
             that._pagerChanged = true;
             topic.publish(
-                'codecompass/search',
-                that._currentQueryData);
+              'codecompass/search',
+              that._currentQueryData);
           }
         }
       });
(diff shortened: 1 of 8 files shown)


mcserep (changes requested):

plugins/search/webgui/js/searchResult.js
```diff
@@ -33,7 +34,8 @@ function (ObjectStoreModel, BorderContainer, declare, Memory, Observable, topic,
   });
 
   var SearchResult = declare(BorderContainer, {
-    constructor : function () {
+    searchTypes : model.searchservice.getSearchTypes(),
```
mcserep, line 37: I get `Uncaught TypeError: Cannot read property 'getSearchTypes' of undefined` for this line.

mcserep (changes requested):

plugins/search/webgui/js/searchResult.js
intjftw: I could not reproduce this error, however, I put the initialization of `searchTypes` into the constructor. Please check if it resolved the error or not.

mcserep: Seems to work for me with this modification.

mcserep (approved):

PR #527, opened 2021-04-13
Conditional compilation of plugins

intjftw:
I added a `WITH_PLUGIN` and a `WITHOUT_PLUGIN` option to the cmake command. By using `WITH_PLUGIN`, the user can specify the plugins they want to build. By using `WITHOUT_PLUGIN`, the user can specify the plugins they don't want to build. The options cannot be used together at the same time. If neither is specified, all plugins are to be built.


doc/deps.md
```diff
@@ -303,3 +303,5 @@ relevant during compilation.
...
 | `CODECOMPASS_LINKER` | The path of the linker, if the system's default linker is to be overridden. |
+| `WITH_PLUGIN` | The names of the plugins to be built. Possible values are **cpp**, **cpp_reparse**, **dummy**, **git**, **metrics**, **search**. The metrics and search plugins are fundamental, they will be compiled even if not included. Keep in mind that the cpp_reparse plugin cannot be compiled without the cpp plugin. `WITH_PLUGIN` **cannot** be used together with `WITHOUT_PLUGIN`. Example: `-DWITH_PLUGIN="cpp;git"` This will compile the cpp, git, metrics and search plugins. |
+| `WITHOUT_PLUGIN` | The names of the plugins to be skipped at build. Possible values are **cpp**, **cpp_reparse**, **dummy**, **git**, **metrics**, **search**. The metrics and search plugins are fundamental, they will be compiled even if included. Keep in mind that the cpp_reparse plugin cannot be compiled without the cpp plugin. `WITHOUT_PLUGIN` **cannot** be used together with `WITH_PLUGIN`. Example: `-DWITHOUT_PLUGIN="dummy;git;search"` This will compile the cpp, cpp_reparse, metrics and search plugins. |
```
whisperity, line 307: It the option cannot be used for `metrics` and `search` (*), I think it is enough if a too eager user is told by the CMake error message. No need to mention in the "beginner's guide" something that shouldn't happen and we can also reasonably guard against anyways.

(*): Although, **why** can't it? I understand the reasoning behind not having the ability to disallow `search`, but why is `metrics` such a core thing? I hardly ever cared about the output of that plugin, ever...)

whisperity: Same with `cpp_reparse`: there's a warning or error if `cpp_reparse` is on and `cpp` isn't, or at least there *should* be. This superfluous explanation makes the documentation hard to follow.

plugins/CMakeLists.txt
```diff
@@ -2,9 +2,44 @@
...
+
 # Add all subdirectories to the build
 file(GLOB plugins RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "${CMAKE_CURRENT_SOURCE_DIR}/*")
 
+# Remove subdirectories according to the provided WITH_PLUGIN or WITHOUT_PLUGIN variables
+if(WITH_PLUGIN)
+  if(NOT ("metrics" IN_LIST WITH_PLUGIN)
+      OR NOT ("search" IN_LIST WITH_PLUGIN))
+    message(WARNING "The search and metrics plugins cannot be left out, "
+            "they will be compiled anyway.")
```
whisperity, line 17: Emphasise they are options:

```suggestion
    message(WARNING "The 'search' and 'metrics' plugins cannot be left out, "
            "they will be compiled anyway.")
```

plugins/CMakeLists.txt
```diff
@@ -2,9 +2,44 @@
...
+       AND NOT ("${_plugin}" STREQUAL "search")
+       AND IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${_plugin}")
+      list(REMOVE_ITEM plugins ${_plugin})
+    endif()
+  endforeach()
+elseif(WITHOUT_PLUGIN)
+  if("metrics" IN_LIST WITHOUT_PLUGIN
+     OR "search" IN_LIST WITHOUT_PLUGIN)
+    message(WARNING "The search and metrics plugins cannot be left out, "
+            "they will be compiled anyway.")
```
whisperity, line 31:
```suggestion
    message(WARNING "The 'search' and 'metrics' plugins cannot be left out, "
            "they will be compiled anyway.")
```

plugins/CMakeLists.txt
```diff
@@ -2,9 +2,44 @@
...
+  endif()
+  foreach(_plugin ${plugins})
+    if(("${_plugin}" IN_LIST WITHOUT_PLUGIN)
+       AND NOT ("${_plugin}" STREQUAL "metrics")
+       AND NOT ("${_plugin}" STREQUAL "search")
+       AND (IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${_plugin}"))
+      list(REMOVE_ITEM plugins ${_plugin})
+    endif()
+  endforeach()
```
whisperity, line 40: Can we factor this loop out from the `if()`s? The hardcoded keeping of `search` and `metrics` and everything else is duplicated in here.

doc/deps.md
intjftw: There is a warning message if someone tries to exclude them. However, if you find this description too much, I'll remove the explanations and merge these two lines into one line in the table.

We talked about it a couple weeks ago, and @zporky said that metrics could be considered a baseline plugin. I agreed because it's such a small and universal plugin, it won't hurt if we make it obligatory. I thought that we could include search, too.

plugins/CMakeLists.txt
intjftw: I refactored it, however, I think it is less readable this way.

doc/deps.md
whisperity: Alright, understandable. Not saying it hurts, just that it was surprising at first glance!

mcserep (approved):
Looks good for me with the modifications.

PR #529, opened 2021-04-25
LDAP authentication

### LINKED ISSUE #521: LDAP authentication
It is a frequently re-emerging topic when CodeCompass is presented to a team if there is an option for LDAP-based authentication. We should upgrade the current authentication to use LDAP.

intjftw:
I added a new authenticator plugin that supports LDAP authentication. The plugin depends on the [ldap-cpp](https://github.com/AndreyBarmaley/ldap-cpp) wrapper of OpenLDAP. The `authentication.json` file should be complemented with the LDAP server configuration as described in `authentication.md`.

The auth is working, however, I plan on adding a new CMake flag to be able to compile any authenticator plugin on demand, and some small changes in the webgui to show who's logged in (and maybe a logout function) thus the Ready-OnHold status. 

Closes #521.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -19,6 +19,7 @@ RUN set -x && apt-get update -qq \
   libgit2-dev \
   libgraphviz-dev \
   libgtest-dev \
+  libldap2-dev \
   libmagic-dev \
   libsqlite3-dev \
   libssl-dev \
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -49,6 +49,7 @@ ENV DATABASE=sqlite \
     INSTALL_DIR=/CodeCompass/install \
     SOURCE_DIR=/CodeCompass/CodeCompass \
     TEST_WORKSPACE=/CodeCompass/test_workspace \
-    TEST_DB="sqlite:database=$TEST_WORKSPACE/cc_test.sqlite"
+    TEST_DB="sqlite:database=$TEST_WORKSPACE/cc_test.sqlite" \
+    WITH_AUTH="plain;ldap"
 
 ENV PATH="$INSTALL_DIR/bin:$PATH"
--- a/docker/runtime/Dockerfile
+++ b/docker/runtime/Dockerfile
@@ -27,6 +27,7 @@ RUN mkdir /CodeCompass-build && \
     -DDATABASE=$CC_DATABASE \
     -DCMAKE_INSTALL_PREFIX=/CodeCompass-install \
     -DCMAKE_BUILD_TYPE=$CC_BUILD_TYPE && \
+    -DWITH_AUTH="plain;ldap" \
   make -j $(nproc) && \
   make install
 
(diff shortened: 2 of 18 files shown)


docker/dev/Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN set -x && apt-get update -qq \
   libgit2-dev \
   libgraphviz-dev \
   libgtest-dev \
+  libldap2-dev \
```
mcserep, line 22: You add the LDAP dependency, but then do not configure CodeCompass with the ldap auth plugin.

docker/runtime/Dockerfile
```diff
@@ -51,6 +51,7 @@ RUN set -x && apt-get update -qq \
     libgit2-dev \
     libssl1.1 \
     libgvc6 \
+    libldap-2.4-2 \
```
mcserep, line 54: You add the LDAP dependency, but CodeCompass is built without the ldap auth plugin.

docker/web/Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN set -x && apt-get update -qq \
     postgresql-server-dev-12 \
     default-jre \
     libgit2-dev \
+    libldap-2.4-2 \
```
mcserep, line 26: You add the LDAP dependency, but CodeCompass is built without the ldap auth plugin.

webserver/authenticators/ldap/src/plugin.cpp
```diff
@@ -0,0 +1,112 @@
+#include <webserver/authenticator.h>
+#include <util/logutil.h>
+
+#include <boost/property_tree/ptree.hpp>
+
+#include <set>
+
+#include "../ldap-cpp/cldap.h"
```
mcserep, line 8: Should be `#include <cldap.h>`, as that folder is on the include path.

webserver/authenticators/ldap/src/plugin.cpp
```diff
@@ -0,0 +1,112 @@
+#include <webserver/authenticator.h>
+#include <util/logutil.h>
+
+#include <boost/property_tree/ptree.hpp>
+
+#include <set>
```
mcserep, line 6: See the *"Order of the inclusion of headers"* in the coding conventions: 

> First include standard C++ headers, then Boost headers, then other supporting library headers (LLVM/Clang, ODB, SQLite, etc.), then your implementing headers.

mcserep (changes requested):
Looks good, only some minor change requests.

Also, we shall build the tarballs with LDAP support, too!

docker/web/Dockerfile
intjftw: Does this mean that I only need to add `-DWITH_AUTH="plain;ldap"` into the runtime container, since its content is copied in this Dockerfile?

docker/dev/Dockerfile
mcserep: Although maybe we don't need that in the dev image.

docker/web/Dockerfile
mcserep: Yes.

mcserep (approved):
Great work, LGTM! 🚀

PR #536, opened 2021-07-05
Disable git blame menu when blame is not available.

### LINKED ISSUE #232: "Team -> Blame" (Git) options should only be visible if its applicable
Currently, this options is always visible. It should be shown either:

 * if the Git parser module could actually parse at least one repository from the inputs (easier)
 * for files that are actually versioned (harder)

intjftw:
Git blame menu (Team --> Blame) does not appear in the right-click menu on clicking a node or a file if:
- There are no repositories available in the workspace directory.
- The file in question is unversioned.
Closes #232.

### DIFF
--- a/plugins/git/service/git.thrift
+++ b/plugins/git/service/git.thrift
@@ -209,5 +209,11 @@ service GitService
     1:string repoId_,
     2:string hexOid_,
     3:string path_,
-    4:string localModificationsFileId_)
+    4:string localModificationsFileId_),
+
+   /**
+   * Check whether there is at least one repository
+   * in the workspace directory.
+   */
+    bool isEnabled()
 }
--- a/plugins/git/service/git.thrift
+++ b/plugins/git/service/git.thrift
@@ -215,5 +215,5 @@ service GitService
    * Check whether there is at least one repository
    * in the workspace directory.
    */
-    bool isEnabled()
+    bool isRepositoryAvailable()
 }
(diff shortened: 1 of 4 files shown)


plugins/git/service/git.thrift
```diff
@@ -209,5 +209,11 @@ service GitService
...
     2:string hexOid_,
     3:string path_,
-    4:string localModificationsFileId_)
+    4:string localModificationsFileId_),
+
+   /**
+   * Check whether there is at least one repository
+   * in the workspace directory.
+   */
+    bool isEnabled()
```
whisperity, line 218: I think `enabled` is too generic and bad for what this function tries to convey. We can selectively disable plugins at some places, some plugins (such as cpp-reparse) have a manual "disable" flag that even if the plugin *is loaded* it should not serve the clients, etc. Let's turn this into a more descriptive name of what the API is trying to convey.

intjftw: I renamed it to `isRepositoryAvailable`.

whisperity: Thanks! Much more clearer this way.

whisperity (approved):
LGTM. Huh, I didn't know about this `isInRepository` flag before. Certainly made it much easier to implement, it seems.

intjftw:
Yes! Although I knew the necessary functions in the `libgit2` library, but I'm glad I found our implementation before trying to implement my own solution. :sweat_smile:

mcserep (approved):
LGTM 🚀

PR #538, opened 2021-08-05
Fix for git blame issue with submodules

Fardas:


mcserep:
> I have only found one issue: if a file in a subrepository is right clicked in the _File Manager_, then a _"[ERROR] Getting tree entry failed: -3"_ error message is logged by the webservice. Source of error: https://github.com/Fardas/CodeCompass/blob/master/plugins/git/service/src/gitservice.cpp#L766 (Git Blame works nevertheless for subrepo files.)

The source of the issue is that in `getRepositoryByProjectPath` [all repositories are iterated over](https://github.com/Fardas/CodeCompass/blob/master/plugins/git/service/src/gitservice.cpp#L212) and while it is checked that the path of the examined repository is the prefix of the file's path, the path of the main repository will most likely prefix **ALL** files.

mcserep:
> The source of the issue is that in `getRepositoryByProjectPath` [all repositories are iterated over](https://github.com/Fardas/CodeCompass/blob/master/plugins/git/service/src/gitservice.cpp#L212) and while it is checked that the path of the examined repository is the prefix of the file's path, the path of the main repository will most likely prefix **ALL** files.

Fixed this in 112c0ee0fe75e35a4d61454e946379998ad0f279. It is strange though you have not faced this issue @Fardas .

mcserep (approved):

PR #546, opened 2021-11-16
Use semantic version requirements for npm dependencies

Wrathan:
Change "marked" package version to 2.1.3, the last one, which supports Node 10.19.0 (the last stable Node version on Ubuntu 20.04) to prevent npm installation and webgui errors.


mcserep:
Instead of a specified version (e.g. 2.1.3), semantic versioning should be used, with [tilde or caret ranges](https://docs.npmjs.com/cli/v6/using-npm/semver), so possible future fixes will be applied.

Using the `>=`  range operator is also not a good practice, as major version upgrades can break functionality, as it happened in this case. These operators should also be refactored sooner or later, but that could be another PR.

mcserep (changes requested):
Instead of a specified version (e.g. 2.1.3), semantic versioning should be used, with [tilde or caret ranges](https://docs.npmjs.com/cli/v6/using-npm/semver), so possible future fixes will be applied.

Using the `>=`  range operator is also not a good practice, as major version upgrades can break functionality, as it happened in this case. These operators should also be refactored sooner or later, but that could be another PR.

mcserep:
Updated to use semantic version requirements for all npm dependencies. See 83b3b9919bfba194588933a36f30e0cac145c9cc.

mcserep (approved):

PR #549, opened 2021-11-29
Introduce Python plugin

rmfcnb:


intjftw (changes requested):

plugins/python/parser/src/pythonparser.cpp
```diff
@@ -0,0 +1,976 @@
...
+                    std::cout << "Unknown status: " << boost::python::extract<int>(status) << std::endl;
+            }
+
+            model::BuildActionPtr buildAction(new model::BuildAction);
+            buildAction->command = "";
+            buildAction->type = model::BuildAction::Other;
+            try{
+                util::OdbTransaction transaction{ctx.db};
+
+                transaction([&, this] {
```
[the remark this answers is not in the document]
mcserep: It is a common practice in the `cppplugin`, too.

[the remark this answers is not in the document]
intjftw: All right, then I revoke this.

plugins/python/parser/src/scripts/cc_python_parser/common/unique_list.py
```diff
@@ -0,0 +1,23 @@
+from typing import TypeVar, List, Iterable
+
+T = TypeVar('T')
+
+
+class UniqueList(List[T]):
```
bruntib, line 6: This class is not used, so it could be removed.

plugins/python/parser/src/scripts/cc_python_parser/function_symbol_collector.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class TemporaryScope(FunctionScope):
+    pass
+
+
+class FunctionSymbolCollector(SymbolCollector, IFunctionSymbolCollector):
+    def __init__(self, symbol_collector: SymbolCollector, tree: Union[ast.FunctionDef, ast.AsyncFunctionDef],
+                 arguments: List[Tuple[DeclarationType, Optional[str]]]):
+        SymbolCollector.__init__(self, ParserTree(tree), symbol_collector.current_file,
```
bruntib, line 23:
```suggestion
        SymbolCollector.__init__(self, tree, symbol_collector.current_file,
```

plugins/python/parser/src/scripts/cc_python_parser/import_preprocessor.py
```diff
@@ -0,0 +1,179 @@
...
+            if m is not None and m.origin is not None:
+                self.import_paths.add(PurePath(m.origin))
+
+    @staticmethod
+    def convert_ast_import_to_import(node: ast.Import, is_local: bool = False) -> List[Import]:
+        imports = []
+        for module in node.names:
+            module_parts = module.name.split('.')
+            module_name = module_parts[-1]
+            module_parts.remove(module_name)
```
bruntib, line 104:
```suggestion
            del module_parts[-1]
```
If the module name appears several times in the full module name then the first occurrence would be removed.

plugins/python/parser/src/scripts/cc_python_parser/import_preprocessor.py
```diff
@@ -0,0 +1,179 @@
...
+
+    @staticmethod
+    def convert_ast_import_from_to_import(node: ast.ImportFrom, is_local: bool = False) -> List[Import]:
+        assert len(node.names) > 0
+        is_module = True
+        try:
+            if node.level > 0 and node.module is None:
+                a = util.find_spec(node.names[0].name)
+                if a is None:
+                    pass    # print()
```
bruntib, line 117:
```suggestion
                if util.find_spec(node.names[0].name) is None:
                    pass    # print()
```

plugins/python/parser/src/scripts/cc_python_parser/main.py
```diff
@@ -0,0 +1,72 @@
...
+                return project_roots
+            for project in config['projects']:
+                if project['status']:
+                    project_roots.append(project['project'])
+    return project_roots
+
+def main():
+    def directory_exception(path: PurePath) -> bool:
+        directory = os.path.basename(os.path.normpath(str(path)))
+        return directory.startswith('.') or directory == 'venv'
```
bruntib, line 47: These shouldn't be hardcoded in the source code. We should support a "skipfile" instead, where the user can provide a set of files and directories to skip. It is not necessarily true that a virtual environment is located in "venv" directory. I'm also not sure if we need to skip 3pp dependencies by default.

plugins/python/parser/src/scripts/cc_python_parser/parser.py
```diff
@@ -0,0 +1,155 @@
...
+            elif len(dependency_file_info) == 1:
+                if dependency_file_info[0].status is ProcessStatus.WAITING:
+                    self.parse_file(dependency_file_info[0])
+                    current_file = file_info.get_file()
+                    logger.debug(f"\nFILE: {file_info.get_file_name()}\n=========================================")
+                    db_logger. \
+                        debug(f"\nFILE: {file_info.get_file_name()}\n=========================================")
+            else:
+                assert False, 'Multiple file occurrence: ' + dependency.get_file()
+        sc = SymbolCollector(ParserTree(tree), file_info.path, file_info.preprocessed_file,
```
bruntib, line 113:
```suggestion
        sc = SymbolCollector(tree, file_info.path, file_info.preprocessed_file,
```

plugins/python/parser/src/scripts/cc_python_parser/parser.py
```diff
@@ -0,0 +1,155 @@
...
+                 exceptions: Optional[ParseException] = None):
+        self.directories: List[str] = directories
+        self.other_modules: Set[PurePath] = set()
+        self.persistence: ModelPersistence = persistence
+        if exceptions is None:
+            exceptions = DefaultParseException()
+        self.exceptions: ParseException = exceptions
+        self.files: List[FileInfo] = []
+        self.other_module_files: List[FileInfo] = []
+        self.parsing_started_files = []
```
bruntib, line 41: These two members are not used.

plugins/python/parser/src/scripts/cc_python_parser/scope_manager.py
```diff
@@ -0,0 +1,501 @@
...
+    def get_qualified_scope_name_from_current_scope(self, scopes: Optional[List[Scope]] = None) -> str:
+        qualified_name_parts = []
+        if scopes is None:
+            for scope in reversed(self.scopes):
+                if isinstance(scope, LifetimeScope) and not isinstance(scope, GlobalScope):  # TODO: ExceptionScope
+                    qualified_name_parts.append(scope.name)
+        else:
+            for scope in reversed(scopes):
+                if isinstance(scope, LifetimeScope) and not isinstance(scope, GlobalScope):  # TODO: ExceptionScope
+                    qualified_name_parts.append(scope.name)
```
bruntib, line 485:
```suggestion
        if scopes is None:
            scopes = self.scopes
        for scope in reversed(scopes):
            if isinstance(scope, LifetimeScope) and not isinstance(scope, GlobalScope):  # TODO: ExceptionScope
                qualified_name_parts.append(scope.name)
```

plugins/python/parser/src/scripts/cc_python_parser/symbol_collector.py
```diff
@@ -0,0 +1,1130 @@
...
+from cc_python_parser.type_deduction import TypeDeduction
+from cc_python_parser.type_hint_data import TypeHintType
+from cc_python_parser.variable_data import VariableDeclaration, ImportedVariableDeclaration, \
+    ModuleVariableDeclaration, StaticVariableDeclaration, NonlocalVariableDeclaration, GlobalVariableDeclaration, \
+    FunctionVariableDeclaration, TypeVariableDeclaration, MethodVariableDeclaration
+from cc_python_parser.common.metrics import metrics
+
+
+class SymbolCollector(ast.NodeVisitor, SymbolFinder, SymbolCollectorBase):
+    def __init__(self, tree: ParserTree, current_file: PurePath, preprocessed_file: PreprocessedFile,
```
bruntib, line 40:
```suggestion
    def __init__(self, tree, current_file: PurePath, preprocessed_file: PreprocessedFile,
```

bruntib (changes requested):
First, I'd like to tell that this Python parser is a great work, I appreciate it. I think we should continue the review process and this will be a good language plugin.

mcserep (changes requested):
I have tested functionally the Python plugin with the following single-file Python script:
https://gist.github.com/mcserep/9df2edbb6767b9190e438f9ed7f3a944

Remarks for the parser:
 - Parsing went smoothly, although took more time than expected. (`requests` and `datetime` modules were imported.)
 - During parsing a lot of debug messages were shown, which should be hidden by default.

Remarks for the service:
 - Multiline documentation comments are not handed properly and treated as Python code (*Invalid PythonAstNode ID* message shown).
 - Clicking on the `audit_submission` function works properly, but when clicking on the `get_user_name` function results in segfault.
 - Clicking on imported symbols (e.g. `requests` or `requests.get`) also displays an *Invalid PythonAstNode ID*.
 - CodeBites diagram is not shown properly.

PR #556, opened 2022-02-03
Filter C/C++ compile commands

intjftw:
Filter compile commands from `compile_commands.json` so C++ plugin won't try to parse everything in case of a multi-language project.

### DIFF
--- a/plugins/cpp/parser/src/cppparser.cpp
+++ b/plugins/cpp/parser/src/cppparser.cpp
@@ -748,20 +748,16 @@ bool CppParser::parseByJson(
 
   //--- Read the compilation commands compile database ---//
 
-  std::vector<clang::tooling::CompileCommand> tempCompileCommands =
+  std::vector<clang::tooling::CompileCommand> compileCommands =
     compDb->getAllCompileCommands();
-  std::vector<clang::tooling::CompileCommand> compileCommands;
 
-  const std::vector<std::string> cppExts{
-    ".c", ".cc", ".cpp", ".cxx", ".o", ".so", ".a"};
-
-  std::copy_if(tempCompileCommands.begin(), tempCompileCommands.end(),
-    std::back_inserter(compileCommands), [&](clang::tooling::CompileCommand c)
-    {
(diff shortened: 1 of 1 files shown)


plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -748,8 +748,21 @@ bool CppParser::parseByJson(
 
   //--- Read the compilation commands compile database ---//
 
-  std::vector<clang::tooling::CompileCommand> compileCommands =
+  std::vector<clang::tooling::CompileCommand> tempCompileCommands =
     compDb->getAllCompileCommands();
+  std::vector<clang::tooling::CompileCommand> compileCommands;
+
+  const std::vector<std::string> cppExts{
+    ".c", ".cc", ".cpp", ".cxx", ".o", ".so", ".a"};
```
bruntib, line 756: There is a function already which checks whether a file is a cpp file:
https://github.com/Ericsson/CodeCompass/blob/master/plugins/cpp/parser/src/cppparser.cpp#L157-L166

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -748,8 +748,21 @@ bool CppParser::parseByJson(
...
+
+  std::copy_if(tempCompileCommands.begin(), tempCompileCommands.end(),
```
bruntib, line 758: We could remove elements with `remove_if` or something similar in order to avoid a copy of the compile commands.

bruntib (changes requested):

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -748,8 +748,21 @@ bool CppParser::parseByJson(
...
+    std::back_inserter(compileCommands), [&](clang::tooling::CompileCommand c)
```
bruntib, line 759:
```suggestion
    std::back_inserter(compileCommands), [&](const clang::tooling::CompileCommand& c)
```

intjftw:
Thanks @bruntib, I committed all changes you requested.

bruntib (approved):

PR #557, opened 2022-02-09
Java plugin

Wrathan:


intjftw:
@Wrathan Please provide a detailed user guide in `doc/usage.md` on how to parse Java projects.

webgui/package.json
```diff
@@ -14,7 +14,7 @@
     "codemirror"       : ">=5.19.0",
     "jquery.fancytree" : ">=2.19.0",
     "svg-pan-zoom"     : ">=2.19.0",
-    "marked"           : ">=0.3.6",
+    "marked"           : "2.1.3",
```
intjftw, line 17: This should be a minimal version, not an exact one.

plugins/dummy/parser/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ include_directories(
   ${PROJECT_SOURCE_DIR}/parser/include)
 
 add_library(dummyparser SHARED
-  src/dummyparser.cpp)
+        src/dummyparser.cpp)
```
intjftw, line 8: This change is unnecessary.

plugins/java/test/src/javaparsertest.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+TEST_F(JavaParserTest, simpleJavaParserTest)
+{
+  ASSERT_EQ(1,1);
+}
+
+TEST_F(JavaParserTest, simpleJavaParserTest2)
+{
+  ASSERT_TRUE(true);
+}
```
intjftw, line 29: Actual tests should be written for the plugin, or this file is unnecessary.

plugins/java/test/src/javaservicetest.cpp
```diff
@@ -0,0 +1,24 @@
...
+   */
+  virtual void TearDown() override
+  {  
+  }
+};
+
+TEST_F(JavaServiceTest, simpleJavaServiceTest)
+{
+  EXPECT_EQ(7,7);
+}
```
intjftw, line 24: Actual tests should be written for the plugin, or this file is unnecessary.

plugins/java/test/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_executable(dummytest
+        src/javaparsertest.cpp
+        src/javaservicetest.cpp)
+
+target_link_libraries(javatest ${GTEST_BOTH_LIBRARIES} pthread)
+
+# Add a test to the project to be run by ctest
+add_test(allJavaTest javatest)
```
intjftw, line 8: Actual tests should be written for the plugin, or this file is unnecessary.

plugins/java/service/src/javaservice.cpp
```diff
@@ -0,0 +1,263 @@
...
+        return;
+
+      std::istringstream s(file->content->content);
+      std::string line;
+      while (std::getline(s, line))
+        content.push_back(line);
+  });
+
+  javaQueryHandler.getSyntaxHighlight(return_, range_, content);
+  */
```
intjftw, line 250: Dead code should be deleted.

intjftw (changes requested):

webgui/package.json
mcserep: Especially semantic versioning should be used.

> **Tilde Version Range (~)**
The ~ operator is best explained by example: ~1.2 is equivalent to >=1.2 <2.0.0, while ~1.2.3 is equivalent to >=1.2.3 <1.3.0. As you can see it is mostly useful for projects respecting [semantic versioning](https://semver.org/). A common usage would be to mark the minimum minor version you depend on, like ~1.2 (which allows anything up to, but not including, 2.0). Since in theory there should be no backwards compatibility breaks until 2.0, that works well. Another way of looking at it is that using ~ specifies a minimum version, but allows the last digit specified to go up.
> 
> **Caret Version Range (^)**
The ^ operator behaves very similarly, but it sticks closer to semantic versioning, and will always allow non-breaking updates. For example ^1.2.3 is equivalent to >=1.2.3 <2.0.0 as none of the releases until 2.0 should break backwards compatibility. For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0 and ^0.0.3 as >=0.0.3 <0.0.4.
> This is the recommended operator for maximum interoperability when writing library code.

scripts/CodeCompass_logger
mcserep, line 1: This file should stay executable.

scripts/remover.sh
mcserep, line 1: This file should stay executable.

plugins/dummy/service/src/plugin.cpp
```diff
@@ -1,6 +1,6 @@
 #include <webserver/pluginhelper.h>
 
-#include <service/dummyservice.h>
+#include <service/javaservice.h>
```
mcserep, line 3: This should not be changed,

plugins/dummy/service/src/javaservice.cpp
```diff
@@ -1,4 +1,4 @@
-#include <service/dummyservice.h>
+#include <service/javaservice.h>
```
mcserep, line 1: This should not be changed,

plugins/dummy/service/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ add_library(dummythrift STATIC
 target_compile_options(dummythrift PUBLIC -fPIC)
 
 add_library(dummyservice SHARED
-  src/dummyservice.cpp
+        src/javaservice.cpp
```
mcserep, line 36: This should not be changed,

plugins/dummy/service/include/service/javaservice.h
mcserep, line 1: This should not be renamed, instead the dummyservice should copied as the new javaservice.

mcserep (changes requested):

Issue #535, opened 2021-05-21
GitHub artifact upload is slow


mcserep:
Uploading the artifacts to GitHub is slow in the CI pipeline. A warning message tells us:
> There are over 10,000 files in this artifact, consider create an archive before upload to improve the upload performance.

We indeed upload over 20,000 files per CodeCompass build job, so this recommendation should be considered.

PR #561, opened 2022-03-02
Archive artifacts before upload

andocz:
Closes https://github.com/Ericsson/CodeCompass/issues/535.

This reduces the time the artifact upload takes from ~20-30 minutes to ~40 seconds. Compare https://github.com/Ericsson/CodeCompass/actions/runs/1886786933 and https://github.com/andocz/CodeCompass/actions/runs/1919240367.

A consequence of this change is that each artifact is now a zip file containing a zip file when downloaded. [This is a limitation of how downloading artifacts from GitHub works](https://github.com/actions/upload-artifact#zipped-artifact-downloads), so there's no way around it. The contents of each inner zip remains the same, though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,26 +196,25 @@ jobs:
         working-directory: ${{github.workspace}}/build
         run: make test ARGS=-V
 
-      - name: Archive CodeCompass binaries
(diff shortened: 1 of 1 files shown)


intjftw (approved):

mcserep (approved):
Looks good to me!

PR #564, opened 2022-03-07
docs: add missing postgresql dependency and use apt everywhere

andocz:
Following the instructions in the docs to set up PostgreSQL on Lubuntu 20.04, I got an error when running `psql`, saying a server wasn't running. It was not until I installed the `postgresql` package that the instructions worked. This package isn't installed by default on [Ubuntu 20.04](http://releases.ubuntu.com/focal/ubuntu-20.04.4-desktop-amd64.manifest) or [18.04](http://releases.ubuntu.com/bionic/ubuntu-18.04.6-desktop-amd64.manifest) desktop either.

I also changed leftover occurrences of `apt-get` to `apt` while I was at it.

### DIFF
--- a/doc/deps.md
+++ b/doc/deps.md
@@ -81,7 +81,7 @@ package manager.
 sudo apt install libsqlite3-dev
 
 # For PostgreSQL database systems:
-sudo apt install postgresql postgresql-server-dev-<version>
+sudo apt install postgresql-server-dev-<version>
 ```
 
 ##### Ubuntu 20.04 ("Focal Fossa") LTS
@@ -91,7 +91,7 @@ sudo apt install postgresql postgresql-server-dev-<version>
 sudo apt install libodb-sqlite-dev libsqlite3-dev
 
 # For PostgreSQL database systems:
-sudo apt install postgresql libodb-pgsql-dev postgresql-server-dev-<version>
+sudo apt install libodb-pgsql-dev postgresql-server-dev-<version>
 ```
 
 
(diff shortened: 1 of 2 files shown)


mcserep:
The `deps.md` contains the guide on how to **compile** CodeCompass. Installing a functional PostgreSQL server is not required for that, only the development package. Instead, the `usage.md`, the *User Guide* describes that a PostgreSQL server should be installed before usage.

Therefore I think the `postgresql` package was omitted deliberately from `deps.md`. What do you think @intjftw , @whisperity , @bruntib ?

andocz:
Ah, my bad, I somehow missed that `usage.md` already told you to install `postgresql-<version>`, even though I edited the command for it...
But in my defense, *all* other command snippets in the documentation are inside code blocks, I think that's why I skimmed over it. Maybe we could change it to be a code block too to save a bit of confusion?

And I'm okay with keeping it out of `deps.md`. Someone writing CI scripts might use `deps.md` as reference, for example, and in that case installing it is unnecessary.

whisperity:
Yes if the main package that also runs a server isn't needed, we should skip it. However, *some* sort of postgres is needed to build and generate the ODB bindings, no? Most likely that is only a dev package.

If the user manual for running wasn't clear, I'd suggest improving the visuals in that.

bruntib:
Yes, I also think that `deps.md` is describing build dependencies only.
Using `apt` and `apt-get` in a uniform manner is a good point.

bruntib (approved):

mcserep:
Renamed the PR to reflect its updated content.

PR #566, opened 2022-03-07
Improve error handling of CppServiceHandler

andocz:
Sometimes a segfault happens when opening "This calls" in the info tree. For example at [`tinyxml2::XMLDocument::SetError`](https://codecompass.net/demo/#wsid=TinyXML2&fid=13444780443610084403&select=2466|1|2491|2&center=text).

I fixed this by adding a check against the query matching nothing, like how it's already done with other queries in the code.
The query is empty if the AST node has no corresponding CPP entity. In TinyXML2's case this happens with `__va_list_tag::__va_list_tag()`. I assume this is an anomaly, so I added a warning log message.

I also updated an extremely outdated comment.

### DIFF
--- a/plugins/cpp/service/src/cppservice.cpp
+++ b/plugins/cpp/service/src/cppservice.cpp
@@ -1615,13 +1615,14 @@ CppServiceHandler::getTags(const std::vector<model::CppAstNode>& nodes_)
 
         FuncResult funcNodes = _db->query<cc::model::CppFunction>(
           FuncQuery::entityHash == defNode.entityHash);
-        if(!funcNodes.empty())
+        if (!funcNodes.empty())
         {
           const model::CppFunction& funcNode = *funcNodes.begin();
 
           for (const model::Tag& tag : funcNode.tags)
             tags[node.id].push_back(model::tagToString(tag));
-        } else
+        }
+        else
           LOG(warning) << "Database query result was not expected to be empty. "
                        << __FILE__ << ", line #" << __LINE__;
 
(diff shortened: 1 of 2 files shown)


intjftw (changes requested):
Please modify according to the C++ coding convention (space between `if` and parentheses, `else` goes to new line after bracket), otherwise it's fine.

plugins/cpp/service/src/cppservice.cpp
```diff
@@ -1611,14 +1611,20 @@ CppServiceHandler::getTags(const std::vector<model::CppAstNode>& nodes_)
...
+          const model::CppFunction& funcNode = *funcNodes.begin();
 
-        for (const model::Tag& tag : funcNode.tags)
-          tags[node.id].push_back(model::tagToString(tag));
+          for (const model::Tag& tag : funcNode.tags)
+            tags[node.id].push_back(model::tagToString(tag));
+        }
+        else
+          LOG(warning) << "Database query result was not expected to be empty. "
+                       << __FILE__ << ", line #" << __LINE__;
```
whisperity, line 1627: I'd rely on the log message being legible to human beings and unique enough to be searched if needed (i.e. "Database query result".. which query? what database?) instead of relying on `__FILE__` and `__LINE__` being defined in the preprocessor stack at all (as opposed to `std::source_location` and its "backend", `__builtin_line()` which is also conditionally portable) and especially being defined to the value that is legible...

Seeing `/var/lib/ci/worker/job/89/workspace/foo/bar/baz/qux.cpp, line# 43892983` won't really tell the user it is an issue with what exact query in the C++ plugin...

(Moreover, all these additional strings just increase the size of the generated library, even in a release build.)

whisperity: "Unexpected empty result when querying tags of C++ function [function name]"?

mcserep:
@whisperity, @intjftw: Is there any reason this is not merged yet?

Seems like all requested changes have been made and it looks good to me.

PR #569, opened 2022-03-09
Fix recursive caller nodes

### LINKED ISSUE #544: Recursive visualization of Caller node in Info Tree
In some cases, inside the Caller node's children, the same Caller node can be found recursively.

Structure (shown in the image):
Caller -> /filename/ -> /function/ -> Caller -> /filename/ -> ...

image

### LINKED ISSUE #562: Template specializations cause duplicate nodes in the Info Tree
Sometimes there are duplicate nodes in the info tree.

For example, [`tinyxml2::DynArray::Push`](https://codecompass.net/demo/#wsid=TinyXML2&fid=203832585766346198&select=226|5|231|6&center=text) has the following code:
```cpp
    void Push( T t ) {
        TIXMLASSERT( _size < INT_MAX );
        EnsureCapacity( _size+1 );
        _mem[_size] = t;
        ++_size;
    }
```

But in the info tree, it is shown as calling `EnsureCapacity` 9 times (should be 1), and being called by `tinyxml2::MemPoolT::Alloc` 5 times (should also be 1). Also, all the usages are listed twice. All of these duplicates have the same line number.

<details>
  <summary>Info Tree</summary>
  
  image
  
</details>

~~This might be related to https://github.com/Ericsson/CodeCompass/issues/544.~~ Actually, they're unrelated.

andocz:
Fixes #544, which was caused by the children of the child caller nodes being set incorrectly. Recursions in the tree are also avoided now.

There is one issue remaining that I need some feedback on. When the child caller nodes are created, the reference count after their name is appended to instead of replaced, which means reference counts keep piling up as we go deeper in the tree. See "Caller(4)(8)(1)" in this screenshot.

This is not intentional, is it? In that case I can fix it too.

### DIFF
--- a/plugins/cpp/webgui/js/cppInfoTree.js
+++ b/plugins/cpp/webgui/js/cppInfoTree.js
@@ -119,9 +119,11 @@ function (model, viewHandler, util) {
     return res;
   }
 
-  function loadReferenceNodes(parentNode, nodeInfo, refTypes) {
+  function loadReferenceNodes(parentNode, nodeInfo, refTypes, scratch) {
     var res = [];
     var fileGroupsId = [];
+
+    scratch = scratch || {};
 
     var references = model.cppservice.getReferences(
       nodeInfo.id,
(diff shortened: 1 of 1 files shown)


mcserep (approved):
Looks good, I don't know why it wasn't reviewed until now :confused: 

Let's get this merged now, anyway! :rocket:

Issue #571, opened 2022-03-22
Relative include flags with spaces cause parsing failure


andocz:
When parsing [Scintilla 3.10.6](https://sourceforge.net/p/scintilla/code/ci/rel-3-10-6/tree/), errors like this occur:
```
/mnt/linux/hg/scintilla-code/lexlib/Accessor.cxx:11:10: fatal error: 'ILexer.h' file not found
#include "ILexer.h"
```
This is because Scintilla uses spaces in the include flags, like so (in `gtk/makefile`):
```
INCLUDEDIRS=-I $(srcdir)/../include -I $(srcdir)/../src -I $(srcdir)/../lexlib
```
This causes an issue due to the combination of the following two things:
1. `CodeCompass_logger` doesn't convert include flags with spaces to absolute paths.
2. `CodeCompass_parser` can only handle absolute include paths.

Removing the space after every `-I` in `gtk/makefile` fixed the issue, confirming this explanation.

mcserep:
Does CodeChecker's logger handle this correctly? If so, can we update CodeCompass' logger with the one in CodeChecker?

andocz:
CodeChecker's logger keeps the include paths as-is, it never converts a relative path to an absolute one or vice-versa.

I think the proper fix would be to make the C++ parser use the compile command's recorded directory as the working directory when running `ClangTool`. We shouldn't rely on `compile_commands.json` to only contain absolute paths. Even if `CodeCompass_logger` always successfully converted all paths to absolute, other tools may generate compilation databases with relative paths (see CodeChecker).

whisperity:
> I think the proper fix would be to make the C++ parser use the compile command's recorded `directory` as the working directory when running `ClangTool`.

This does seem like the right call, but let's ask @bruntib to give us a heads-up related to this issue and what steps were taken in CodeChecker. I have vague memories of issues like this constantly looming around us...

PR #572, opened 2022-03-28
Change working directory of C++ parser

andocz:
This PR makes Clang tool invocations use the working directories of each compilation command. This fixes relative paths in arguments not being handled properly (#571), as previously the tool ran in the default working directory of `.`.

Callers of `SourceManager::getFile` in `cppparser` were changed to provide an absolute path.

Additionally, inaccurate comments in `SourceManager::getFile` were corrected.

### DIFF
--- a/parser/include/parser/sourcemanager.h
+++ b/parser/include/parser/sourcemanager.h
@@ -51,8 +51,8 @@ public:
    * based on the given path_. The object is read from a cache. If the file is
    * not in the cache yet then a model::File entry is created, persisted in the
    * database and placed in the cache. If the file doesn't exist then it returns
-   * nullptr.
-   * @param path_ The file path to look up.
+   * a file with no contents.
+   * @param path_ The absolute file path to look up.
    */
   model::FilePtr getFile(const std::string& path_);
 
--- a/parser/src/sourcemanager.cpp
+++ b/parser/src/sourcemanager.cpp
@@ -166,7 +166,7 @@ model::FilePtr SourceManager::getFile(const std::string& path_)
   boost::filesystem::path canonicalPath
     = boost::filesystem::canonical(path_, ec);
 
-  //--- If the file can't be found on disk then return nullptr ---//
+  //--- If the file can't be found on disk then return a blank file ---//
 
   bool fileExists = true;
   if (ec)
--- a/plugins/cpp/parser/include/cppparser/filelocutil.h
+++ b/plugins/cpp/parser/include/cppparser/filelocutil.h
@@ -79,23 +79,31 @@ public:
   }
 
   /**
-   * This function returns the file path in which loc_ location takes place. The
-   * location is meant to be the expanded location (in case of macro expansion).
-   * If the file can't be determined then empty string returns.
+   * This function returns the absolute path of the file in which loc_ location
+   * takes place. The location is meant to be the expanded location (in case of
+   * macro expansion).
+   * If the file can't be determined then an empty string is returned.
    */
   std::string getFilePath(const clang::SourceLocation& loc_)
   {
-    clang::SourceLocation expLoc = _clangSrcMan.getExpansionLoc(loc_);
-    clang::FileID fid = _clangSrcMan.getFileID(expLoc);
+    return getFilePath(
+      _clangSrcMan.getFileID(_clangSrcMan.getExpansionLoc(loc_)));
+  }
 
-    if (fid.isInvalid())
+  /**
+   * This function returns the absolute path of the file identified by fid_.
+   * If the file can't be determined then an empty string is returned.
+   */
+  std::string getFilePath(const clang::FileID& fid_)
+  {
+    if (fid_.isInvalid())
       return std::string();
 
-    const clang::FileEntry* fileEntry = _clangSrcMan.getFileEntryForID(fid);
+    const clang::FileEntry* fileEntry = _clangSrcMan.getFileEntryForID(fid_);
     if (!fileEntry)
       return std::string();
 
-    return fileEntry->getName();
+    return fileEntry->tryGetRealPathName();
   }
 
 private:
--- a/plugins/cpp/parser/src/cppparser.cpp
+++ b/plugins/cpp/parser/src/cppparser.cpp
@@ -308,7 +308,8 @@ int CppParser::parseWorker(const clang::tooling::CompileCommand& command_)
     clang::tooling::FixedCompilationDatabase::loadFromCommandLine(
       argc,
       commandLine.data(),
-      compilationDbLoadError));
+      compilationDbLoadError,
+      command_.Directory));
 
   if (!compilationDb)
   {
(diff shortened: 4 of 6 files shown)


whisperity:
> It turns out there's a bug I didn't catch: if the build directory doesn't exist, the parsing fails, because `RealFileSystem` requires that the `chdir` target exists. To fix this, I'll have to reimplement `RealFileSystem` without this check.

I do not understand this. How is it possible to have a build generated with CMake (which creates directory structures) or logged (in which case the build was executed and should have created the directories) but do not have the "build directory" (what does this refer to?) exist?

Is it meaningful to parse a file in such a context? If the directory where the build was executed (is this what "build directory" refers to?) doesn't even exist, how was the build executed? We'll end up having issues with (potentially generated?) files gone missing...

In general, I would prefer if we tried not to reimplement things from LLVM too much, unless really needed. We're way behind the upstream and all these reimplements will cause problems when we try supporting newer releases of Ubuntu even.

Perhaps we could either create the directory, or give a hard error and have the user reconsider what they messed up with their build system...

whisperity:
Okay, I tried giving a second read to the patch itself but I am still confused. What do we need the `chdir` for, and why do we try `chdir`ing to that directory? Or is that `chdir` needed by Clang instead? What about `VirtualFileSystem`? Can't we create a `VirtualFileSystem` that always says _"Yes, sir, this directory surely exists!"_ in case it doesn't really exist on the real device, and overlay that on top of the RFS in the VFS stack the `ClangTool` uses and then make `clang` believe that directory is there?

andocz:
> How is it possible to have a build generated with CMake (which creates directory structures) or logged (in which case the build was executed and should have created the directories) but do not have the "build directory" (what does this refer to?) exist?

It can happen if the build directory is selectively transferred between computers, like what happens on the CI where we zip up just the source files. The *build directory* (where CMake was invoked) exists here actually, but the *compilation directory* (`"directory"` in `compile_commands.json`) of some commands doesn't, since directories containing no source files go missing.

> We'll end up having issues with (potentially generated?) files gone missing...

We can have the compilation dependencies all intact with only compilation directories containing none missing, which is what happened on the CI.

> In general, I would prefer if we tried not to reimplement things from LLVM too much, unless really needed.

Oh, I didn't mean reimplementing it from scratch, just creating a wrapper `FileSystem` subclass that adjust arguments before passing them on to the methods of a `RealFileSystem` instance, converting relative paths to absolute ones. It would work like [`ProxyFileSystem`](https://github.com/llvm/llvm-project/blob/release/10.x/llvm/include/llvm/Support/VirtualFileSystem.h#L366-L405).

> Or is that `chdir` needed by Clang instead?

Yes, it `chdir`s to the directory of each compilation command before running the tool [here](https://github.com/llvm/llvm-project/blob/release/10.x/clang/lib/Tooling/Tooling.cpp#L513-L514). This is what will make future references to relative paths get resolved correctly.

> Can't we create a `VirtualFileSystem` that always says [...]

I don't think that would work. [`RealFileSystem` uses `llvm::sys::fs::is_directory`](https://github.com/llvm/llvm-project/blob/release/10.x/llvm/lib/Support/VirtualFileSystem.cpp#L324-L327) to verify that the work dir exists; the `ClangTool`'s stack isn't used at this level.

But even if it worked, I think my "redirecting wrapper class" idea is more straightforward.

whisperity:
Oh, there is `ProxyFileSystem` now... neat. Back when I implemented reparse, that wasn't a thing yet. The setCWD would've worked though, because that's also what reparse uses:

https://github.com/Ericsson/CodeCompass/blob/83a6a9a724892f458060e1a39b89a5383d381102/plugins/cpp_reparse/service/src/databasefilesystem.cpp#L248-L261

And in this case, almost none (apart from Clang's own intrinsic headers that install with CodeCompass...) of the directories exist on the server. And the overlay is passed to `ClangTool`:

https://github.com/Ericsson/CodeCompass/blob/83a6a9a724892f458060e1a39b89a5383d381102/plugins/cpp_reparse/service/src/reparser.cpp#L174-L176

andocz:
The reason missing directories never caused an issue with reparse is because it doesn't set the working directory properly. It's using the default value of `.` from `FixedCompilationDatabase::loadFromCommandLine`. It's a good thing you mentioned reparse, because I realized it's still broken with relative paths.

`DatabaseFileSystem` doesn't actually use its `_currentWorkingDirectory` for anything. It passes the paths it receives as-is to `getFile`. Also, `setCurrentWorkingDirectory` doesn't work properly if `path_` is absolute; the `path_` gets appended to the work dir instead of being used in itself. These will have to be corrected.

andocz:
The reparser properly supports relative paths now. I had to add a new table to the database (`BuildDirectory`) for this.

I decided to scrap the `FakeWorkDirFileSystem`, because to make it work properly, it would need to resolve `..`s in paths lexically, which is too messy for my liking. (I can think of symlinks as a case when this would produce wrong results.) Now if the build directory is missing, a warning is emitted and `/` is used as a fallback.

For parsing to fail, a build command needs to have its directory missing *and* use a relative path. I don't think it's worth going out of our way to recover from this scenario given how unlikely it is. You suggested creating the build directory, but I think the parser shouldn't modify its **in**puts.

model/include/model/builddirectory.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace cc
+{
+namespace model
+{
+
+struct BuildAction;
+
+#pragma db object
+struct BuildDirectory
```
whisperity, line 19: I'm not sure why this has to be a brand new table, foreign key, and all that jazz for the storage. `BuildAction` already contains `string command;`, for example. If `"directory"` is just another member of the entries in a compilation database JSON (just like `"command"`), then why isn't this reflected in the schema? Every build action **MUST** have a build directory (it just may or may not exist or point to anything meaningful).

This struct, as it is now, won't do uniqueing of build directories either, because it is the build directory that refers to **one** build action each. If we want a separate record to only store one directory ("string") exactly once, then it is the buildaction that has to refer into this relation's key.

andocz: I wanted to keep backwards compatibility with existing databases, that's why I didn't want to change `BuildAction`'s schema. Is that not a priority? In that case you're right, storing it in `BuildAction` would make more sense.

Deduplicating the strings is not worth the hassle, in the CodeCompass SQLite database `BuildAction` only takes up 0.012% of the space.

mcserep: We can easily reparse projects, backward-compatibility with existing databases is not a high priority.

mcserep:
The PR should also handle relative paths for the `file` in the compilation command source.

As part of #637, it was discover that in case the `directory` is `/source/`, the `file` is `file.cpp` and the execution directory of  `CodeCompass_parser` is `/cc/`, then instead of `/source/file.cpp`, the non-existent file `/cc/file.cpp` is being parse, leading to error.

PR #574, opened 2022-04-04
fix(build): (NFC) Fix some compiler warnings emitted when compiling with Clang 13.0

whisperity:
When trying to compile CodeCompass with LibTooling built on top of Clang 13.0, some warnings were emitted that broke the build. This patch fixes these.

 - `-Wc++11-narrowing`
 - `-Wmismatched-tags`
 - `-Wrange-loop-construct`

### DIFF
--- a/plugins/cpp/parser/src/cppparser.cpp
+++ b/plugins/cpp/parser/src/cppparser.cpp
@@ -469,7 +469,7 @@ void CppParser::markModifiedFiles()
   // Detect changed files through C++ header inclusions.
   util::OdbTransaction {_ctx.db} ([&]
   {
-    for (const model::FilePtr file : filePtrs)
+    for (const model::FilePtr& file : filePtrs)
     {
       if(file)
       {
@@ -711,7 +711,7 @@ void CppParser::initBuildActions()
   });
 }
 
-void CppParser::markByInclusion(model::FilePtr file_)
+void CppParser::markByInclusion(const model::FilePtr& file_)
 {
   auto inclusions = _ctx.db->query<model::CppHeaderInclusion>(
     odb::query<model::CppHeaderInclusion>::included == file_->id);
--- a/plugins/cpp/test/src/cppparsertest.cpp
+++ b/plugins/cpp/test/src/cppparsertest.cpp
@@ -361,9 +361,10 @@ TEST_F(CppParserTest, Record)
     {
       EXPECT_EQ(n.symbolType, model::CppAstNode::SymbolType::Type);
 
+      using LineType = decltype(n.location.range.start.line);
       switch (n.location.range.start.line)
       {
-        case -1: 
+        case static_cast<LineType>(-1):
           EXPECT_TRUE(
             // TODO: investigate the type of this. It is possibly the parameter
             // of a compiler generated copy constructor or assignment operator.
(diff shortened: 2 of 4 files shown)


plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -469,7 +469,7 @@ void CppParser::markModifiedFiles()
   // Detect changed files through C++ header inclusions.
   util::OdbTransaction {_ctx.db} ([&]
   {
-    for (const model::FilePtr file : filePtrs)
```
whisperity, line 472:
```

/local/repo/CodeCompass/plugins/cpp/parser/src/cppparser.cpp:472:31: error: loop variable 'file' creates a copy from type 'const model::FilePtr' (aka 'const shared_ptr<cc::model::File>') [
-Wrange-loop-construct]
    for (const model::FilePtr file : filePtrs)
                              ^
```

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -711,7 +711,7 @@ void CppParser::initBuildActions()
   });
 }
 
-void CppParser::markByInclusion(model::FilePtr file_)
+void CppParser::markByInclusion(const model::FilePtr& file_)
```
whisperity, line 714: `FilePtr` is actually a `shared_ptr` and this function only uses it once to read from it to construct more `shared_ptr`s

webserver/src/mainrequesthandler.h
```diff
@@ -12,7 +12,7 @@ namespace webserver
 class Session;
 class SessionManager;
 
-struct MainRequestHandler
```
whisperity, line 15:
```

/local/repo/CodeCompass/webserver/include/webserver/session.h:23:1: warning: class 'MainRequestHandler' was previously declared as a struct; this is valid, but may result in linker errors un
der the Microsoft C++ ABI [-Wmismatched-tags]
class MainRequestHandler;
^
/local/repo/CodeCompass/webserver/src/mainrequesthandler.h:15:8: note: previous use is here
struct MainRequestHandler
       ^
```

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -361,9 +361,10 @@ TEST_F(CppParserTest, Record)
     {
       EXPECT_EQ(n.symbolType, model::CppAstNode::SymbolType::Type);
 
+      using LineType = decltype(n.location.range.start.line);
       switch (n.location.range.start.line)
       {
-        case -1:
```
whisperity, line 366:
```

/local/repo/CodeCompass/plugins/cpp/test/src/cppparsertest.cpp:366:14: error: case value evaluates to -1, which cannot be narrowed to type 'cc::model::Position::PosType' (aka 'unsigned long'
) [-Wc++11-narrowing]
        case -1:
             ^
```

plugins/cpp/test/src/cppparsertest.cpp
```diff
@@ -653,10 +654,13 @@ TEST_F(CppParserTest, Variable)
     astNodes = _db->query<model::CppAstNode>(
       QCppAstNode::entityHash == memberVariable.entityHash);
 
+    using LineType =
+        decltype(std::declval<model::CppAstNode>().location.range.start.line);
     for (const model::CppAstNode& n : astNodes)
       switch (n.location.range.start.line)
       {
-        case -1: // Access by compiler generated constructors.
```
whisperity, line 659:
```

/local/repo/CodeCompass/plugins/cpp/test/src/cppparsertest.cpp:659:14: error: case value evaluates to -1, which cannot be narrowed to type 'cc::model::Position::PosType' (aka 'unsigned long'
) [-Wc++11-narrowing]
        case -1: // Access by compiler generated constructors.
             ^
```

mcserep (approved):
Looks good to me!

mcserep:
Are we planning to use conventional commits from now on, or just personal preference? 😄

whisperity:
> Are we planning to use conventional commits from now on, or just personal preference? smile

@mcserep Ah, just reflex. I was committing to like 5 different repositories this morning.

PR #579, opened 2022-05-24
Finding CodeCompass binary path

intjftw:
Multiple plugins need to know the absolute path of the CodeCompass binary that is being run, in places where they don't reach `argv[0]`.

### DIFF
--- a/util/include/util/filesystem.h
+++ b/util/include/util/filesystem.h
@@ -16,6 +16,12 @@ namespace util
  */
 std::string binaryPathToInstallDir(const char* path);
 
+/**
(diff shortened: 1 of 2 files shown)


bruntib:
What do plugins need this information for?

intjftw:
> What do plugins need this information for?

It's usually that the plugin needs to run another binary on a relative path to the CC binary (e.g. the competence plugin wants to run JPlag, Java and C# plugins need to run their own parser components in their respective languages), and this can't be done if the CC binary is not run from the install directory.

mcserep (approved):
Looks fine for me :+1:

PR #582, opened 2022-09-16
Support for analyzing YAML files

intjftw:
This is a new language plugin to extend CC with. Ericsson products have been transforming their projects from monolithic to service oriented software architecture so they are now built up of microservices. Since projects are very large and sprawling, there is a wide demand in Ericsson for a static analysis tool that will help understanding the network of microservices within a project.

This plugin consists of 

- a model of various db tables
- a parser that integrates [yaml-cpp](https://github.com/jbeder/yaml-cpp): it is capable of analyzing YAML files, and finding microservices in Helm charts. It comes with a relation collector to find connection points between previously found microservices.
- a web service: there are some basic table-form visualizations of YAML data, and more elaborate visualizations of microservices.

I'll keep expanding this plugin with the huge amount of feature requests that I received from Ericsson, but I think that the plugin can be officially added to CC at this point with the current feature set.

The basis of this work was done by @rizwankadhar as his BSc thesis. Thanks Rizwan for your work!


intjftw:
> (Just a side note: if we are embedding business logic into this module that is specific to specific YAML schemae (I see Helm mentioned in the original post) then maybe calling the plugin "YAML Parser" is a bit too generic.)

The plugin became more and more specific to Helm and Kubernetes from the initial point of plain YAML parsing. I plan to rename it to "Microservice plugin" or something like that.

intjftw:
I added the most important modifications to the PR. It is now in a state where parsing is correct and thorough enough to provide useful functionality. I removed the WIP flag, as the PR is ready for review.

mcserep:
Resolved merge conflicts in 71047a8a5ef555785672fffb280b460a308fe8d1.

plugins/helm/yaml-cpp-config.cmake
```diff
@@ -0,0 +1,16 @@
...
+
+set_and_check(YAML_CPP_INCLUDE_DIR "@PACKAGE_CMAKE_INSTALL_INCLUDEDIR@")
+
+# Our library dependencies (contains definitions for IMPORTED targets)
+include(@PACKAGE_CONFIG_EXPORT_DIR@/yaml-cpp-targets.cmake)
+
+# These are IMPORTED targets created by yaml-cpp-targets.cmake
+set(YAML_CPP_LIBRARIES "@EXPORT_TARGETS@")
+
+check_required_components(@EXPORT_TARGETS@)
```
mcserep, line 16: What is the purpose of this file? I don't see it being used at all.

plugins/helm/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+find_package(yaml-cpp REQUIRED)
```
mcserep, line 1: This should be moved to the `CMakeLists.txt` file of the parser component of the plugin, as it is only used there.

docker/runtime/Dockerfile
```diff
@@ -71,6 +71,7 @@ RUN set -x && apt-get update -qq && \
     libldap-2.4-2 \
     libmagic-dev \
     libthrift-dev \
+    ibyaml-cpp-dev \
```
mcserep, line 74: Typo.

plugins/helm/service/include/service/yamlservice.h
```diff
@@ -0,0 +1,189 @@
...
+#include <model/yamlastnode-odb.hxx>
+
+#include <projectservice/projectservice.h>
+#include <util/graph.h>
+#include <util/legendbuilder.h>
+#include <util/odbtransaction.h>
+#include <util/util.h>
+#include <webserver/servercontext.h>
+
+//#include <LanguageService.h>
```
mcserep, line 29: Commented out code.

plugins/helm/parser/include/parser/yamlparser.h
```diff
@@ -0,0 +1,105 @@
...
+
+#include <util/parserutil.h>
+
+#include "yaml-cpp/yaml.h"
+
+namespace cc
+{
+namespace parser
+{
+class YamlParser : public AbstractParser
```
mcserep, line 17: While the plugin directory was renamed to `helm`, the parser and service classes kept the name `YamlParser` and `YamlService`. This is highly confusing.

plugins/helm/parser/src/templateanalyzer.h
```diff
@@ -0,0 +1,140 @@
...
+    std::map<std::string, YAML::Node>& fileAstCache_);
+
+  ~TemplateAnalyzer();
+
+  void init();
+  uint64_t getTemplateCounter() { return templateCounter; }
+
+private:
+  bool visitKeyValuePairs(
+    std::string path_,
```
mcserep, line 38: Code smell: string parameter passed by value, instead of const reference. Is it required to make a copy here?

plugins/helm/parser/src/valueanalyzer.h
```diff
@@ -0,0 +1,71 @@
...
+    std::map<std::string, YAML::Node>& fileAstCache_,
+    uint64_t templateIdCounter);
+
+  void init();
+
+  ~ValueAnalyzer();
+
+private:
+  YAML::Node findValue(
+    std::string value_,
```
mcserep, line 36: Code smell: string parameter passed by value, instead of const reference. Is it required to make a copy here?

plugins/helm/parser/src/valueanalyzer.h
```diff
@@ -0,0 +1,71 @@
...
+    ParserContext& ctx_,
+    std::map<std::string, YAML::Node>& fileAstCache_,
+    uint64_t templateIdCounter);
+
+  void init();
+
+  ~ValueAnalyzer();
+
+private:
+  YAML::Node findValue(
```
mcserep, line 35: Function not defined, only declared?

plugins/helm/webgui/js/yamlDiagram.js
```diff
@@ -0,0 +1,78 @@
...
+
+      var diagramTypes = model.yamlservice.getMicroserviceDiagramTypes(serviceInfo);
+      for (let diagramType in diagramTypes)
+        submenu.addChild(new MenuItem({
+          label   : diagramType,
+          type    : diagramType,
+          onClick : function () {
+            var that = this;
+
+            //topic.publish('codecompass/openFile', { fileId : serviceInfo.id });
```
mcserep, line 57: Commented out code.

plugins/helm/webgui/js/yamlMenu.js
```diff
@@ -0,0 +1,151 @@
...
+        }
+      });
+    }
+  };
+
+  viewHandler.registerModule(getdefintion, {
+    type : viewHandler.moduleType.TextContextMenu,
+    service : model.yamlservice
+  });
+*/
```
mcserep, line 44: Commented out code.

plugins/helm/webgui/js/yamlMenu.js
```diff
@@ -0,0 +1,151 @@
...
+        }
+      });
+    }
+  };
+
+  viewHandler.registerModule(infobox, {
+    type : viewHandler.moduleType.TextContextMenu,
+    service : model.yamlservice
+  });
+*/
```
mcserep, line 105: Commented out code.

webgui/scripts/codecompass/viewHandler.js
```diff
@@ -31,6 +31,7 @@ function (lang, Deferred, model) {
       InfoTree               :  6, /** Info tree item (eg. CppInforTree)**/
       FileManagerContextMenu :  7, /** File manager context menu (eg. Diagram directory)**/
       InfoPage               :  8, /** Info page (eg. Credit, User Guide) **/
+      MicroserviceContextMenu:  9  /** Microservice navigator context menu **/
```
mcserep, line 34: Again, `helm` vs. `yaml` vs. `microservice`. Three different naming for the same thing. Confusing.

webgui/index.html
```diff
@@ -9,6 +9,7 @@
 
     <!-- CSS -->
 
+    <link rel="stylesheet" href="style/codecompass.css" />
```
mcserep, line 12: Why is this required? It should be bundled and loaded by Dojo.

plugins/helm/parser/src/templateanalyzer.cpp
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    case model::MSResource::ResourceType::MEMORY:
+    case model::MSResource::ResourceType::STORAGE:
+    {
+      int m = amount_.find('M');
+      if (m != std::string::npos)
+      {
+        amount_.erase(m, 2);
+        LOG(info) << std::stof(amount_);
```
mcserep, line 420: This and other log commands produces some strange and verbose output, e.g.:

> 2023-08-10 15:44:30 [INFO] 50Gi
> 2023-08-10 15:44:30 [INFO] 50
> 2023-08-10 15:44:30 [INFO] 256
> 2023-08-10 15:44:30 [INFO] 8Gi
> 2023-08-10 15:44:30 [INFO] 8
> 2023-08-10 15:44:30 [INFO] 8Gi
> 2023-08-10 15:44:30 [INFO] 8

There should be more context and the messages either debug or trace level.

mcserep (changes requested):
The code changeset generally meets the coding guidelines, but multiple naming conventions (*helm*, *yaml* and *microservice*) creates confusion.

When testing on the gitlab Helm chart, I received a segmentation fault, therefore I could not test the functionality.

Also, the update for the tarball creation (new dependency) is missing from the PR.

webgui/scripts/codecompass/viewHandler.js
intjftw: In this case, I would leave it like this. The primary purpose of the Helm chart analysis is to identify microservices and their connections, so I think this naming is correct here. I'm unifying it in other places.

plugins/helm/yaml-cpp-config.cmake
intjftw: `find_package(yaml-cpp)` uses this file to set the `YAML_CPP_INCLUDE_DIR` and `YAML_CPP_LIBRARIES` variables which are used in the `CMakeLists.txt` file of the parser component.

mcserep: `yaml-cpp-config.cmake` is installed with the `libyaml` library, for me is available on the path `/usr/lib/x86_64-linux-gnu/cmake/yaml-cpp/yaml-cpp-config.cmake`.

That is the used version. This local version is not referenced at all and is not used as I have seen.

webgui/scripts/codecompass/viewHandler.js
mcserep: As far as I understand this plugin is capable for Kubernetes microservice analysis based on Helm charts. Naming it simply microservice analysis in general is a bit like naming the "C++ plugin" as "Programming language plugin", as that would be a much wider concept.

There are other microservice orchestration frameworks beside Kubernetes, like Docker Swarm, Helios, Amazon ECS or Mesos, just to name a few. This plugin does not deal with those systems, instead it is fundamentally coupled with Helm charts for Kubernetes. Thus, in my opinion, the naming should use the `helm` keyword mostly and also towards the end user it should be named something like that.

PR #584, opened 2022-10-04
Support for C# static analysis

intjftw:
The official support for C# static analysis. Uses the Roslyn Semantic API to parse source code.

This plugin has been made by @Borisz42 as his TDK thesis. Thanks Dávid for your work!


plugins/csharp/service/src/csharpservice.cpp
```diff
@@ -0,0 +1,303 @@
...
+      FileQuery::id == std::stoull(fpos_.file));
+  });
+  _csharpQueryHandler.getAstNodeInfoByPosition(return_, file->path, fpos_.pos);
+}
+
+void CsharpServiceHandler::getSourceText(
+        std::string& return_,
+        const core::AstNodeId& astNodeId_)
+{
+  LOG(info) << "getSourceText";
```
intjftw, line 94: All unnecessary debug comments should be deleted.

plugins/csharp/service/src/csharpservice.cpp
```diff
@@ -0,0 +1,303 @@
...
+namespace service
+{
+namespace language
+{
+typedef odb::query<cc::model::File> FileQuery;
+namespace fs = boost::filesystem;
+namespace bp = boost::process;
+namespace pt = boost::property_tree;
+
+int CsharpServiceHandler::_thriftServerPort = 9091;
```
intjftw, line 18: Let us make sure that the port is free and if it's not, search for a free one.

plugins/csharp/parser/src_csharp/AstVisitor.cs
```diff
@@ -0,0 +1,1050 @@
...
+
+            int len = data_.Length;
+            for (int i = 0; i < len; ++i)
+            {
+                hash ^= data_[i];
+                hash *= 1099511628211;
+            }
+
+            return hash;
+        }
```
intjftw, line 59: Helper functions like these might be placed in a separate file / class.

plugins/csharp/parser/src_csharp/AstVisitor.cs
```diff
@@ -0,0 +1,1050 @@
...
+
+        private CsharpAstNode AstNode(SyntaxNode node, AstSymbolTypeEnum type)
+        {            
+            return AstNode(node, type, AstTypeEnum.Declaration);
+        }
+
+        public override void VisitUsingDirective(UsingDirectiveSyntax node)
+        {
+            base.VisitUsingDirective(node);
+            //Adatbázisban nem kell feltétlenül tárolni, inkább csak azt kell biztosítani hogy amiket meghívunk vele azok is be legyenek járva
```
intjftw, line 112: Unnecessary comments should be deleted.

plugins/csharp/service/src_csharp/CSharpQueryServer.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        public static ILogger<T> CreateLogger<T>() => LogFactory.CreateLogger<T>();
+    }
+
+    public class CSharpQueryServer
+    {
+        private static readonly ILogger Logger = LoggingHelper.CreateLogger<CSharpQueryServer>();
+        private static readonly TConfiguration Configuration = new TConfiguration();
+        private static int port = 9091;
```
intjftw, line 46: Although I rewrote this part to not clash with the cpp service, and to work with multiple parsed C# projects, port numbering should really be given some thought. The plugin should look for the next free port if the current one is taken.

intjftw (commented):
The plugin generally works fine, however, there is much space for improvement. We should run some tests on larger projects to find critical errors.

lib/csharp/thrift_netstd/Benchmarks/Thrift.Benchmarks/CompactProtocolBenchmarks.cs
```diff
@@ -0,0 +1,79 @@
+// Licensed to the Apache Software Foundation(ASF) under one
```
intjftw, line 1: We should investigate whether this large pack of Thrift files is really needed as source in the plugin. It seems unncessary.

plugins/csharp/parser/src/csharpparser.cpp
```diff
@@ -0,0 +1,168 @@
...
+  command.append(paths_[1]);
+  command.append("' '");
+  command.append(csharp_path.string());
+  command.append("' ");
+  command.append(std::to_string(_ctx.options["jobs"].as<int>()));
+  LOG(debug) << "CSharpParser command: " << command;
+
+  ch::steady_clock::time_point begin = ch::steady_clock::now();
+  
+  int result = bp::system(command, bp::start_dir(csharp_path), bp::std_out > log);
```
ervin7mk, line 78: CodeCompass parser will fail in case of calling it using a relative path. Not finding the C# binary, due to the relative path might cause this issue. The following usage will cause a failure:
 `~/CodeCompass_install/bin/CodeCompass_parser -f -w ~/[...] -n unit2 -i ~/[...] -i ~/[...] -d "[...]"`
This error message will be shown:
`termiante called after throwing an instance of 'boost::process::process_error
what():executive failed: No such file or directory 
Aborted (core dumped)`

plugins/csharp/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_subdirectory(parser)
+#add_subdirectory(test)
```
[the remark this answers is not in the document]
mcserep: In other plugins, there are no `test` folder if no valid tests are provided. I suggest to follow this convention.

plugins/csharp/service/src_csharp/csharpservice.csproj
```diff
@@ -0,0 +1,32 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\..\model\*.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\..\lib\csharp\thrift_netstd\Thrift\Thrift.csproj" />
+    <ProjectReference Include="..\..\..\..\..\..\..\codecompass\intjftw-build\plugins\csharp\service\src_csharp\gen-netstd\gen-netstd.csproj" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
mcserep: Tough question, I see the following options:
1. Generated, language-specific Thrift interfaces should be generated into the source folder and version controlled. These files rarely change and it is also an issue for the web frontend that the required JS/TS interfaces are not available in the source folder of CodeCompass. As a first attempt we could do this for the C# Thrift interfaces to see how this work. However, if shall refactor this later to the complete project. (Preferably in another PR.)
2. Copy the complete C# source code into the build folder, edit the .csproj file there, and do the compilation there. Basically that is what we do for the current and the new (#590) webgui, in the installation folder. Kinda ugly to be honest.
3. During building copy the `csharpservice.csproj` inside the source directory, e.g. `csharpservice_generated.csproj` and edit the copied version. Use that for building the C# project. Place `csharpservice_generated.csproj` in the gitignore list.

In my opinion, option 2 is what we currently do, but it is not really nice. Option 1 should be the proper solution, but it requires refactoring for the complete project to be a proper solution. Option 3 requires minimal extra effort. Although it creates a new generated file in the source directory, that could easily be ignored from VC. It is not really the proper and general solution, but can be done easily.

So maybe do Option 3 now and later refactor as suggested in Option 1 in a separate PR?

plugins/csharp/parser/src/csharpparser.cpp
intjftw: Fixed by 48f9983212824cff5f9aea9bc118c25149d03b71.

mcserep:
I have updated this PR with the changes of #817 and #818. Some changes requested by @intjftw are still necessary before we can merge this.

mcserep:
@matecsad On second thought, please open a separate PR targeting the `feature/csharp_plugin` branch with the fixes for the remaining open issues.

PR #589, opened 2023-01-11
Reducing the number of db connections at runtime

### LINKED ISSUE #485: Plugins should share their database connection pool
Connection to SQLite / PostgreSQL database is performed by the **`cc::util::connectDatabase()`** function.
The function contains a simplistic database connection caching technique: existing connections are stored in the **`std::map<std::string, std::shared_ptr<odb::database>> databasePool`** static variable and can be reused later if the connection string matches precisely:
https://github.com/Ericsson/CodeCompass/blob/173a3cf15e15be6de38b476fa7ff5eb258829e8c/util/src/dbutil.cpp#L291-L304

However the util module is statically linked to all plugin shared objects (and to the `CodeCompass_webserver` and `CodeCompass_parser` binaries), thus the `databasePool` map is not shared between them, but instead all plugins have their separate database connection pool.

While [not documented properly](https://www.codesynthesis.com/pipermail/odb-users/2014-November/002242.html), we also know that we don't need to deal with creating separate database connections per thread, because it is done by ODB automatically.
> If you try to access the database from multiple threads simultaneously, then ODB will create and use several connections.

This is the reason why we experienced that the CodeCompass webserver maintains approximately `threads * plugins` amount of database connections, which can easily go over the default maximum limit of 100 in case of PostgreSQL.
While we definitely need separate connections per threads, **the database pool between shared objects and libraries shall be shared!**

intjftw:
In this PR I made the `util` library shared instead of static. This way each parsed project shall open one connection to the database at webserver runtime instead of each plugin opening its own connection per project.

@mcserep thank you for your help with this.

Fixes #485.

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -6,7 +6,7 @@ include_directories(
 include_directories(SYSTEM
   ${ODB_INCLUDE_DIRS})
 
-add_library(util STATIC
+add_library(util SHARED
   src/dbutil.cpp
   src/dynamiclibrary.cpp
   src/filesystem.cpp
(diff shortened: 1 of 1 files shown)


mcserep (approved):
LGTM! 🚀

PR #591, opened 2023-02-10
Incorrectly handled foreach style loops

### LINKED ISSUE #509: Incorrectly handled foreach style loops
It seems CodeCompass handles *foreach* style `for` loops incorrectly when clicking on the enumerated variable.

**How to reproduce:**
- In the PCL source code click on `centroids_`  in line 97.  or  on `centroid` in line 101. In the *InfoTree*, the complete functions is shown, and the goto definition feature also work in correctly.
https://codecompass.net/demo/#wsid=PCL&fid=16855117887643076224&select=97|1|98|1&center=text
- In the CodeCompass source click `_ctx.fileStatus` in line 358. to get the very same issue:
https://codecompass.net/demo/#wsid=CodeCompass&fid=3022161984525021070&select=358|1|359|1&center=text

zsofiaschell:
Fix #509 
TraverseCXXForRangeStmt and VisitCXXForRangeStmt added to handle foreach style loops correctly.


plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -141,6 +141,50 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
...
+
+  bool VisitCXXForRangeStmt(clang::CXXForRangeStmt *forRangeStmt)
+  {
+    model::CppAstNodePtr astNode = std::make_shared<model::CppAstNode>();
+    
+    clang::DeclRefExpr* declRefExpr = llvm::dyn_cast<clang::DeclRefExpr>(forRangeStmt->getRangeInit());
+
+    clang::Stmt* stmt = llvm::dyn_cast<clang::Stmt>(forRangeStmt->getRangeInit());
+
+    clang::CallExpr* callExpr = llvm::dyn_cast<clang::CallExpr>(stmt);
```
mcserep, line 157: When we have a function or method call in the range.based for loop, the `forRangeStmt->getRangeInit()` will be a `ExprWithCleanups` statement, and one of its descendant in the AST will be the `CallExpr` or `CXXMemberCallExpr`.

Eg. for a function call:
image

And for a method call:
image

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -141,6 +141,50 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
...
+    const clang::VarDecl* vd = llvm::dyn_cast<clang::VarDecl>(forRangeStmt->getRangeStmt()->getSingleDecl());
+    astNode->entityHash = util::fnvHash(getUSR(vd));
+
+    astNode->symbolType = model::CppAstNode::SymbolType::Variable;
+
+    astNode->astType = model::CppAstNode::AstType::Read;
+
+    astNode->id = model::createIdentifier(*astNode);
+
+    _astNodes.push_back(astNode);
```
mcserep, line 184: Since you also call `VisitDeclRefExpr`, this seems to be redundant.

mcserep (changes requested):
Regular variables seems to be handled correctly now, but iterating over function and method call results not.

I have used the following sample code for testing:
https://github.com/mcserep/cc-tests/blob/master/foreach/main.cpp

plugins/cpp/parser/src/clangastvisitor.h
mcserep: @whisperity: is there an easy way to request the first descendant of a node which matches a certain type? E.g. to get the first `CallExpr` descendant of the `ExprWithCleanups` node in the AST?

whisperity: Unfortunately, I think no. Every node is implemented with different fields and accessor methods. The whole `RecursiveASTVisitor` is using lots of generated code in the background to facilitate the traversals. We could try using the _AST Matchers_ library (that Clang-Tidy also uses...) that has a generic `has()` matcher which can bind a child node.

plugins/cpp/parser/src/clangastvisitor.h
zsofiaschell: Removed.

Issue #580, opened 2022-05-26
Support Ubuntu Jammy 22.04


mcserep:
Ubuntu 22.04 (Jammy Jellyfish) is out, we shall support it officially.
We shall still continue to support Ubuntu 18.04, as the EOL is 2023 April and the current ELX release still depends on it as far as I know.

intjftw:
A couple issues with this:
- LLVM 10 is not available to install from apt in Ubuntu 22.04. LLVM 11 to 14 are available.
- LLVM 10+ had some modifications in the methods we use so upgrading requires some updates to our code as well.
- The problem with ODB in Ubuntu 18.04 re-emerged in 22.04. We need to manually compile ODB.
- Only Thrift 0.16.0 is available in Ubuntu 22.04 which is - of course - not backward-compatible, so we need to manually compile Thrift 0.13.0 as well.
- There are a few missing header inclusions in our files which g++ 11 refuses to complement, so those should be added.

mcserep:
My suggestions:
- Drop support for Ubuntu 18.04. Its EOL is already near anyway and one problem less here.
- Use LLVM 11, which can be installed from the standard APT repository both for Ubuntu 20.04 and 22.04.
- If we need manual compilation for ODB on Ubuntu 22.04, it is a shame, but we have no other choice if the packaged version in the APT repository is incorrect.
- We shall stay with Thrift 0.13.0 for now and compile it manually on Ubuntu 22.04. Most of the developers still use Ubuntu 20.04. Later, in a separate ticker, we can upgrade to Thrift 0.16.0, update the codebase and require manual compilation on Ubuntu 20.04.

mcserep:
The Ubuntu 18.04 CI jobs failed both for #582 and #584 recently. It seems Thrift 0.13.0 does not compile anymore on Ubuntu 18.04:
> CSC : error CS7027: Error signing output with public key from file 'thrift.snk' -- File not found. [/home/runner/thrift-0.13.0/lib/netstd/Thrift/Thrift.csproj]

Another reason to drop support for Ubuntu 18.04 😄

tiaanwessels:
I'd like to experiment with CodeCompass on Ubuntu 22 but get the following when i build it:
[ 20%] Building CXX object service/authentication/CMakeFiles/authenticationthrift.dir/gen-cpp/authentication_constants.cpp.o
[...]

PR #593, opened 2023-02-23
Update PostgreSQL user permission in Usage Guide

mcserep:
CodeCompass does not require `SUPERUSER` privileges in PostgreSQL, only `CREATEDB`.

Also updating documentation links to more up to date version (`9.5` -> `12`).

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -38,7 +38,7 @@ psql
 In the PostgreSQL command-line shell, type:
 
 ```sql
-CREATE USER compass WITH SUPERUSER LOGIN PASSWORD '<mypassword>';
+CREATE USER compass WITH CREATEDB LOGIN PASSWORD '<mypassword>';
 ```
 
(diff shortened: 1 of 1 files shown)


intjftw:
LGTM!
Related question: should we also update the Postgres version in the CI and the tarball as well? It's kind of inconsistent. For Ubuntu 18.04 we install v. 10 (I know we'll stop supporting it soon but still), and also we build v.10.x in the tarball.

mcserep:
> LGTM! Related question: should we also update the Postgres version in the CI and the tarball as well? It's kind of inconsistent. For Ubuntu 18.04 we install v. 10 (I know we'll stop supporting it soon but still), and also we build v.10.x in the tarball.

I don't think we should spend any effort on updating the Ubuntu 18.04 scripts, as we will drop support for it soon anyways (see #580). Also, in the default package repository of Ubuntu 18.04, PostgreSQL version 10 is available, 12 is not.

Issue #573, opened 2022-03-28
Non-existent files are resolved as the working directory file


andocz:
AST nodes that aren't located in any file are reported in the database as being located inside the working directory file.

As a result, the directory that the parser was called from appears in the file manager in the web UI.

This happens because `cc::parser::FileLocUtil::getFilePath(const clang::SourceLocation&)` returns an empty string if no file containing the SourceLocation can be found. This "path" is directly passed to `cc::parser::SourceManager::getFile(const sdt::string&)` in numerous places in the code. This function resolves it against the working directory, returning the same directory.

My planned solution is to make `SourceManager::getFile` return `nullptr` if a relative path (which an empty string is) is passed as the argument. This means the `location_file` field for these nodes will be `null` in the database. Does that sound like a good format? 

I wanted to discuss this before implementing a fix as it requires a (small) change to the database schema.

mcserep:
@andocz Can you gives us a MWE to test and validate this issue on? (An AST node without location?)
Where did you find this unintended behaviour?

andocz:
TinyXML2 is one example. I parsed it by invoking `CodeCompass_parser` while cd'd to CodeCompass's  `bin` directory, and these nodes had the `bin` directory as their location:
![](https://user-images.githubusercontent.com/100377984/196568169-25873adb-cd77-4ee6-ac98-ca28e998671b.png)

But any C++ code should do the trick. The key is to invoke the parser from a directory unrelated to the project being parsed.

mcserep:
@andocz Thanks for the quick response, we will reproduce this first.

However I don't think that the correct behaviour would be to add `null` value for the `location_file` field for these records, only if the related `astValue` really belongs to no file. On the screenshot you shared this is not the case for many occurrences, e.g. the  `main(int, const char **)::TestUtil::~TestUtil()` should belong to a file, I guess to the one containing the `main` function.

So the source if this faulty behaviour should be examined first.

ervin7mk:
I've succesfully reproduced the issue brought up by @andocz . I created a directory unrelated to the project named ParseFromHere and I invoked the parser while inside this directory. As seen in the attached screenshots, there are nodes related to this directory. 
![ParseFromHere_found_in_File](https://user-images.githubusercontent.com/84744394/208257386-d43f0a26-133d-4a08-93ae-328d8a50a312.png)
[...]

PR #594, opened 2023-03-06
Fix non-existent files resolving as working directory

ervin7mk:
See issue [https://github.com/Ericsson/CodeCompass/issues/573](url) for details

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -1246,13 +1246,16 @@ private:
     if (start_.isInvalid() || end_.isInvalid())
     {
       fileLoc.file = getFile(start_);
-      const std::string& type = fileLoc.file.load()->type;
-      if (type != model::File::DIRECTORY_TYPE && type != _cppSourceType)
+      if (fileLoc.file)
       {
-        fileLoc.file->type = _cppSourceType;
-        _ctx.srcMgr.updateFile(*fileLoc.file);
+        const std::string& type = fileLoc.file.load()->type;
+        if (type != model::File::DIRECTORY_TYPE && type != _cppSourceType)
+        {
+            fileLoc.file->type = _cppSourceType;
+            _ctx.srcMgr.updateFile(*fileLoc.file);
+        }
       }
-      return fileLoc;
+        return fileLoc;
     }
 
     clang::SourceLocation realStart = start_;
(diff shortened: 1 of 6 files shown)


plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -1246,13 +1246,16 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
...
-        _ctx.srcMgr.updateFile(*fileLoc.file);
+        const std::string& type = fileLoc.file.load()->type;
+        if (type != model::File::DIRECTORY_TYPE && type != _cppSourceType)
+        {
+            fileLoc.file->type = _cppSourceType;
+            _ctx.srcMgr.updateFile(*fileLoc.file);
+        }
       }
-      return fileLoc;
+        return fileLoc;
```
bruntib, line 1258: Unnecessary indentation.

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -1268,14 +1271,16 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
...
+      const std::string& type = fileLoc.file.load()->type;
+      if (type != model::File::DIRECTORY_TYPE && type != _cppSourceType)
+      {
+        fileLoc.file->type = _cppSourceType;
+          _ctx.srcMgr.updateFile(*fileLoc.file);
+      }
     }
-
-    return fileLoc;
+      return fileLoc;
```
bruntib, line 1283: Unnecessary indentation.

plugins/cpp/parser/src/ppincludecallback.cpp
```diff
@@ -77,39 +77,48 @@ void PPIncludeCallback::InclusionDirective(
 
   std::string includedPath = searchPath_.str() + '/' + fileName_.str();
   model::FilePtr included = _ctx.srcMgr.getFile(includedPath);
```
bruntib, line 79: Instead of adding extra indentation everywhere, maybe it was simpler to bring the `getFile()` calls from line 79 and 94 to the beginning before line 76 and make an early return if either of them is null. It would indicate better that no action is needed in that case.

plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -374,17 +379,20 @@ std::vector<std::vector<std::string>> CppParser::createCleanupOrder()
       {
         auto file = _ctx.srcMgr.getFile(item.first);
 
-        auto inclusions = _ctx.db->query<model::CppHeaderInclusion>(
-          odb::query<model::CppHeaderInclusion>::included == file->id);
-
-        for (const auto& inclusion : inclusions)
+        if (file)
```
bruntib, line 382: Consider using the early return technique if the nesting is too deep.

bruntib (changes requested):

Issue #590, opened 2023-01-16
New web frontend


mcserep:
The design concept of the current web frontend is considered outdated in some terms. The task is to completly redesign and implement a new web frontend for CodeCompass from scratch. Technologies and libraries to use:
 - React
 - Material Design

The new frontend shall use the existing RPC-based Thrift backend and its provided features of CodeCompass.

mdeme01:
The new frontend is currently in development [here](https://github.com/mdeme01/CodeCompass/tree/new-frontend/webgui-new).

So far, the following features were implemented:
- Switching between workspaces
- Viewing files in the file manager and file contents in the editor
- Searching and displaying search results
- Generating metrics and diagrams for files/directories

Let me know if there are any problems, or if the design should be different.

mcserep:
@mdeme01 If you think that your development is ready for a first round of review, please open a pull request. It is easier to review the code that way and leave remarks.

You can still push new commits to your branch of course and the PR will auto-update.

mdeme01:
Pull request is ready [here](https://github.com/Ericsson/CodeCompass/pull/595).

mcserep:
With @intjftw, we have deployed the current version of the new frontend for review:  
https://codecompass.net/demo/new/

intjftw:
Closed by #595.

PR #595, opened 2023-03-27
New frontend in React

mdeme01:
#590

New frontend for CodeCompass in Next.js (React) - TypeScript, using the Material UI library.

As there are a lot of commits and changes, it is possible that there are some bugs left in the codebase. When I was testing it, it worked fine, but I might have missed something. Also, the UI/UX was created according to me, so if anything should be different, let me know.

If you want to test it, please refer to the README I added - the necessary steps can be found there.

Update: Now all features have been implemented.


whisperity:
How does the server know which UI to serve? I'm trying to find that change in the diff but so far didn't have any luck.

intjftw:
Hi Deme,

We looked at the PR. In my opinion, you've done a very good job so far. :)

A couple notes:
- When we select a node, e.g. a function in a cpp file, the entire code in the definition of the node is highlighted, and the cursor jumps to the end of the code. This could be fine when the code snippet is short, but rather bothersome when it's long.
- The visualization and animations of code metrics is confusing. Sometimes the squares cover each other during the animation, and it's not obvious which file or unit a square represents.
- The icons in the info tree, rev control navigator etc. are missing.
- The "File references" menu is a great idea (I don't remember having it from the old GUI). However, the same node is present many times.

All in all, great work, let's have a meeting together with @mcserep.

mcserep:
The PR should have been opened from a feature branch, instead of `master`.
This way it will mess up your history in your fork after this gets merged.

.gitignore
```diff
@@ -6,3 +6,44 @@ nbproject/
 *.swp
 build/
 build_*/
+
+# Next.js project for webgui
```
mcserep, line 10: These new rules are specific to the `webgui-new` folder, and thus should be placed there.

webgui-new/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
```
mcserep, line 2: Is this enough? Does this include React specific linting rules? Maybe `plugin:react/recommended` should be added as well.

webgui-new/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "eslint": "8.31.0",
+    "eslint-config-next": "13.1.1",
+    "typescript": "4.9.4"
+  },
+  "dependencies": {
+    "@codemirror/lang-cpp": "^6.0.2",
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
+    "@mui/icons-material": "^5.11.0",
+    "@mui/lab": "^5.0.0-alpha.116",
```
mcserep, line 27: An alpha version dependency seems suspicious.

mdeme01: The @mui/lab package is part of the Material UI library, and is used to add more components to the main MUI package. From this package, I'm using the TreeView component, which is used to display the parts of the UI that need a tree-style view (for instance the file tree), so it is mandatory to use in the project. Because it's in alpha version, I carefully checked and it does not cause any problems and is safe to use in this case.

Furthermore, the package has constant maintenance and an active community, so it can be updated when it is out of the alpha version. Also, it is one of the most popular UI libraries, so I think we can make an exception in this case.
https://snyk.io/advisor/npm-package/@mui/lab

mdeme01:
> * The "File references" menu is a great idea (I don't remember having it from the old GUI). However, the same node is present many times.

The reason why some of the nodes are duplicated is because functions have both a declaration and a definition (and they're indentical texts). However, it is also like this in the old UI in the file outline menu. Should I remove the function declarations here, or leave it as it is?

webgui-new/README.md
```diff
@@ -0,0 +1,76 @@
...
+You can specify the properties of the webserver by creating a `.env` file at the root of the project. This is only required for development.\
+The default values are: `localhost` for host, `8080` for port, `false` for https, and path will be an empty string.\
+Note: you need to set the `WEBSERVER_PORT`, if you choose to use another port for the webserver, the other properties are not that important considering you run a local webserver.
+
+```py
+# .env
+WEBSERVER_HOST=
+WEBSERVER_PORT=
+WEBSERVER_HTTPS=
+WEBSERVER_PATH=
```
mcserep, line 49: It could be a simplification to replace this with a simple parameter, e.g. `BACKEND_URL`. Then with the [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) type you could create an object from it to easily get the schema, the domain, the port and the path.

webgui-new/README.md
```diff
@@ -0,0 +1,76 @@
+# New Next.js (React) based webgui for CodeCompass
```
mcserep, line 1: We could name this *New Web GUI development guide* and focus only on the parts on how to set up a development environment for the new web GUI. If someone would not like to develop the web GUI, it should be enough to read just the regular development guide.

mcserep (changes requested):

mcserep:
@mdeme01 Please add a new option to the *Menu* button in the top right corner to switch to the new UI! It is required to set up a convenient test environment on codecompass.net.
Screenshot from 2023-04-24 23-19-57

webgui-new/InstallReact.cmake
```diff
@@ -0,0 +1,46 @@
...
+  execute_process(
+    COMMAND npm install
+    WORKING_DIRECTORY ${INSTALL_WEBROOT_REACT_DIR}/app)
+endif()
+
+message("Installation of npm packages are finished.")
+
+# Generate TypeScript from Thrift
+execute_process(
+  COMMAND /bin/sh -c "chmod +x thrift-codegen.sh"
```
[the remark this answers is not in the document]
mdeme01: I tried using the `file(CHMOD ...)` command, as written in the documentation. But it seems it is only supported in version 3.19 of CMake, and my version is 3.16.3 - which is what was installed by following the build guide. Is it only my version, or is this the one used for CodeCompass?

webgui-new/README.md
```diff
@@ -0,0 +1,76 @@
...
+Make sure you are using at least version 18 of [Node.js](https://nodejs.org/en/).\
+You can also download [nvm](https://github.com/nvm-sh/nvm), and run `nvm install` at the root of the project to use the specified Node version (18.13.0).
+
+## Development
+
+For easier development, first run the webserver (CodeCompass_webserver), then run the development server.\
+Since the webserver and development server are different origins, you have to use the [CORS extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=en) to resolve CORS issues.
+
+Alternatively, you could add this change to the `MainRequestHandler::begin_request_handler(struct mg_connection* conn_)` method in `mainrequesthandler.cpp`, then build the webserver again.\
+(Note: Do not commit this change, as it is unnecessary in production.)
```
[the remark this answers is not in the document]
mdeme01: The problem with the proxy is that it only works with actual API routes in Next.js, which is not the case in this application, so for now I'm not seeing a solution for this.

I will remove the other solution from the README to modify the main request handler, because I agree it is messy. However, I think for development, using the CORS extension I linked is perfectly fine, because it is easy to use (you just have to toggle a button in the browser), and is only required for development.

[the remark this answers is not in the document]
mcserep: I am not that familiar with Next.JS, but I think the `http-proxy-middleware` package should not be affected by it, since it is a lower level package below the framework, modifying the middleware.

The browser extension option could be mentioned, but that is a 3rd party, Chrome-specific solution to this problem. I think the application itself should provide a solution for this, proxying in the development server is a widely used approach in my experience. (Such a solution was used in all Angular and React projects I have worked so far.)

webgui-new/README.md
```diff
@@ -0,0 +1,76 @@
...
+npm install
+```
+
+Generate TypeScript files from Thrift by running the following command:
+
+```bash
+chmod +x thrift-codegen.sh && ./thrift-codegen.sh --thrift-source ../
+```
+
+You can specify the properties of the webserver by creating a `.env` file at the root of the project. This is only required for development.\
```
mcserep, line 33: Usually the `.env` file is under version control, defining a default configuration. Users can set up a `.env.local` file to override it with local configuration, and that file if ignored from version control. We should also follow this directive.
https://create-react-app.dev/docs/adding-custom-environment-variables/

webgui-new/InstallReact.cmake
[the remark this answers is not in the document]
mcserep: I did not check that `file(CHMOD ...)` is not available in CMake 3.16 yet. Then disregard this comment and stick to that version, as that is the available version in the Ubuntu 20.04 APT repository.

webgui-new/README.md
mdeme01: Now with the suggested changes, there is only one `.env` variable, `BACKEND_URL`, which already has a default, fallback value set in `next.config.js`, which is `http://localhost:8080` (because the default port for the webserver is `8080`).

Users can override this by creating a `.env` file and setting the `BACKEND_URL` variable. I understand why using a `.env.local` file may be a good idea, but I don't think it is really necessary in this case, as there is only one variable, which has a fallback value and only required for development.

Of course, if we really want to follow this directive, I can look into this, but in my opinion it is not that necessary in this case.

webgui-new/README.md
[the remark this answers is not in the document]
mdeme01: Next.js provides an option in the `next.config.js` file, called `rewrites`. However, this can only be used to proxy client-side routes. The `http-proxy-middleware` package you mentioned can only be used for Node.js webservers.

Unfortunately, you cannot set the request header of a webserver from a client-side application. For this, you would have to modify the webserver's request handler (like the way I mentioned earlier in the README).

webgui-new/README.md
mcserep: Thanks for refactoring this! 👍 

The goal of the `.env` file is to define the default values, which can be overridden. Of course if you hard-code the default values into `next.config.js`, it becomes an empty file; but the conventional solution would be to place the default value (`http://localhost:8080`) to the `.env` file.

It is a minor issue now and you can leave like this if you want, but most likely it will be refactored later anyway when more environment variables are introduced for future features. Then it would be much better to have all those default values in a single, separate file.

webgui-new/README.md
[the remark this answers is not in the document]
mcserep: > Next.js provides an option in the `next.config.js` file, called `rewrites`. However, this can only be used to proxy client-side routes. The `http-proxy-middleware` package you mentioned can only be used for Node.js webservers.

In development mode, Next.js runs a Node.js webserver as well. Otherwise, it won't be able to serve the frontend.
That is the reason of the CORS error, since this Node.js webserver runs on a different port than the backend.
https://nextjs.org/docs/deployment#nodejs-server

We will only use the proxy in development mode. In production a static build is deployed on the same host and port as the backend, so no CORS issue will be raised.

> If I understand correctly, you cannot set the request header of a webserver from a client-side application. For this, you would have to modify the webserver's request handler (like the way I mentioned earlier in the README).

The proxy server standing between the frontend and the backend will solve this, that is the purpose of it.

webgui-new/README.md
mcserep: @mdeme01 Additional remark: the official Next.js documentation also recommends *NOT* to use runtime configuration in `next.config.js`, like you do, but instead purely environment variables.

https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration
> Note: This feature is considered legacy and does not work with [Automatic Static Optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization), [Output File Tracing](https://nextjs.org/docs/advanced-features/output-file-tracing#automatically-copying-traced-files), or [React Server Components](https://nextjs.org/docs/advanced-features/react-18/server-components). Please use [environment variables](https://nextjs.org/docs/basic-features/environment-variables) instead in order to avoid initialization overhead.

mdeme01: Okay, I removed the public runtime config and commited the `.env` file. Now the variables can be overriden by creating a `.env.local` file and setting the variables inside it.

Because `.env` variables cannot be accessed in the browser, I used the built-in feature of Next.js to expose them, prefixing them with `NEXT_PUBLIC_`.

To determine which URL to use for the configuration, the app needs to know if it is in a production environment or not - I used the `NODE_ENV` variable previously for this, but because I cannot access it in the browser now, I had to create another variable, `NEXT_PUBLIC_DEV_ENV`, which is `false` by default, so in the `.env.local` file, you also have to set this variable to `true` for development.

webgui-new/README.md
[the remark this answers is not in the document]
mdeme01: Then I misunderstood something. I will look into this, but for know I don't know how would I implement something like this in my case.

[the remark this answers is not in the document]
mdeme01: Update:

Apparently, there is a way to override the Next.js server and use a custom Node.js server, which I tried, and it works. Here I can set the proxy, and the CORS issues are resolved.

However, I need to use the `NEXT_PUBLIC_BACKEND_URL` `.env` variable for the proxy, and only Next.js provides an option to override `.env` variables with a `.env.local` file, Node.js does not. And because the Node.js server starts before the Next.js app initializes, the overrided `.env` variables won't be available. Therefore, `.env.local` is not a valid option anymore if we want a proxy server.

webgui-new/README.md
mcserep: I have used the `http-proxy-middleware` package with environment files for Node.js based proxying to avoid CORS issues in multiple projects, it works.

Such an example is the [TMS](https://gitlab.com/tms-elte/frontend-react) submission management project developed at the ELTE FI in React. See `.env` files and the `src/setupProxy.js` for reference.

I will get on to review the other parts of the PR. If you are stuck with how to implement this, we shall get back to this sometime later. Let's keep this thread open as a reminder.

webgui-new/README.md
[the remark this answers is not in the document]
mdeme01: Update:

I found out that you can define a custom Node.js webserver that overrides the Next.js webserver. This way, the proxy can be created. However, there are still some issues I need to resolve.

webgui-new/README.md
mdeme01: Thank you, I will take a look, but I think you replied to the wrong thread. 🙂

mcserep: Yes, I replied from mobile, and not to the proper thread then :sweat_smile:

webgui-new/README.md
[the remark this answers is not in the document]
mcserep: This was my original answer, but I replied to another thread, so I will just copy it here for future reference:

> I have used the `http-proxy-middleware` package with environment files for Node.js based proxying to avoid CORS issues in multiple projects, it works.
>
> Such an example is the [TMS](https://gitlab.com/tms-elte/frontend-react) submission management project developed at the ELTE FI in React. See `.env` files and the `src/setupProxy.js` for reference.
>
> I will get on to review the other parts of the PR. If you are stuck with how to implement this, we shall get back to this sometime later. Let's keep this thread open as a reminder.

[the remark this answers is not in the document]
mdeme01: I've made little progess. I found out that it is possible to overwrite the Next.js development server with a custom Node.js server, where you can use the `http-proxy-middleware` package.

The proxy works, however, it proxies the contents of the whole webserver. This means that when running the development server, the contents of the proxied webserver will be visible, and not the Next.js application.

I understand that there is a way to only proxy certain requests, like `/api`, just as in the example you provided, at TMS. But in my case, there are no separated API requests. I tried filtering out all the requests that end in `Service`, because only the requests like `/WorkspaceService` and `/ProjectService` would need to be proxied, the others would need to be handled by Next.js, but so far, I didn't have any luck.

I commited these changes to a separate branch on my fork, `proxy-server`, because it think it is a good start.
https://github.com/mdeme01/CodeCompass/commit/e4b8ed7b67ac6a97ba517e4cd4471fcb1bc7ec4a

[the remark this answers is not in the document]
mcserep: > I've made little progess. I found out that it is possible to overwrite the Next.js development server with a custom Node.js server, where you can use the http-proxy-middleware package.

Sounds good. If we would like to use the `Next.JS` development server, there also seems to be a `next-http-proxy-middleware` package for the framework, which can add the required proxy functionality to Next.JS.

> I understand that there is a way to only proxy certain requests, like /api, just as in the example you provided, at TMS. But in my case, there are no separated API requests. I tried filtering out all the requests that end in Service, because only the requests like /WorkspaceService and /ProjectService would need to be proxied, the others would need to be handled by Next.js, but so far, I didn't have any luck.

Maybe the easiest solution for this to really serve the Thrift requests with a path prefix like `/api`. Regardless of the development environment, this logical separation of the backend and frontend can be beneficial anyway. (E.g. it enables to deploy the frontend and the backend separately later.)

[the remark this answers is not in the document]
mdeme01: Update:

I managed to solve it. I modifed the config to use the current URL for the Thrift service at all times, not just in production. This way the `/Service` paths can be proxied too to the webserver. Then I used a `pathFilter` to only proxy the paths ending with `/Service`, the others are handled by Next.js.

It was a little unclear to me how would I create the proxy with the `next-http-proxy-middleware` package (because it is dependant on the built-in `/api` routes of Next.js), so I sticked with a custom Node.js server.

It is also worth mentioning that the `pathFilter` option is only available in the v3 beta verison of `http-proxy-middleware`, but it is really mandatory to use in this case. However, I read that the v3 will be out of beta soon.

And also another note, this way, the proxying also works with `rewrites` in the `next.config.js`. However, the requests were doubled because of the rewrites and the app was slower, so I decided not to go with this option.

mcserep:
**[Jump to label and jump to root]**

The old web UI displays shortcuts to jump to the labels (defined at parsing, stored in the project workspace) and an option to jump to the root folder. These features are missing from the new web GUI currently.
image

mcserep:
**[Persistent URLs]**

In CodeCompass, the support of persistent URLs is an important feature. Persistent URLs mean that every navigating action on the GUI (opening a project, opening a file, navigating in a file, etc.) modifies the URLs. This way the URL can be copied and sent to colleagues, who will see the same view upon clicking it. This should be supported in the new UI as well.

webgui-new/InstallReact.cmake
```diff
@@ -0,0 +1,45 @@
...
+execute_process(
+  COMMAND npm run build
+  WORKING_DIRECTORY ${INSTALL_WEBROOT_REACT_DIR}/app)
+
+# Move build directory out of application directory
+file(REMOVE_RECURSE ${INSTALL_WEBROOT_REACT_DIR}/out)
+file(RENAME ${INSTALL_WEBROOT_REACT_DIR}/app/out ${INSTALL_WEBROOT_REACT_DIR}/out)
+
+# Create symbolic link
+execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${INSTALL_WEBROOT_REACT_DIR}/out ${INSTALL_WEBROOT_DIR}/new)
```
mcserep, line 38: This should use a relative path to be more robust. (E.g. the installation folder can be moved or renamed.)

webgui-new/next.config.js
```diff
@@ -0,0 +1,18 @@
...
+  reactStrictMode: true,
+  modularizeImports: {
+    '@mui/icons-material': {
+      transform: '@mui/icons-material/{{member}}',
+    },
+  },
+  trailingSlash: true,
+  ...(process.env.NODE_ENV === 'production' && {
+    assetPrefix: '/new',
+    basePath: '/new',
```
mcserep, line 14: It is not guaranteed that the new web GUI will be served on the `/new` absolute path, e.g. it can be served on `/demo/new` here:
https://codecompass.net/demo/new/

This required the local modification of this file. I am thinking about whether NextJs could handle relative paths here or not?

webgui/scripts/codecompass/view/header.js
```diff
@@ -54,6 +54,14 @@ function (cookie, topic, declare, DropDownMenu, DropDownButton, Select,
         }
       }));
 
+      menu.addChild(new MenuItem({
+        label     : 'New Web GUI',
+        onClick   : function () {
+          window.open(
+            `${window.location.origin}/new`);
```
mcserep, line 61:
```suggestion
            `${window.location.pathname}/new`);
```

mcserep: So it is compatible with URLs where the CodeCompass GUI not served under the root folder.

mdeme01:
> [Persistent URLs]

A part of this feature is already supported. You can right click on a source element in the code and click the 'Get permalink to selection' option. This way, you can send it to someone, and the app will initialize the states based on the URL, jumping to that project, file and selection.

I handled persistence using the local storage, and not the URL, and I realized later on that the old UI uses the URL for this. To fully support this feature, it would mean a lot of refactorings in the code.

Time is getting short before my thesis deadline (less than 2 weeks), so I would like to not start into larger refactorings, only fix smaller issues, and focus on writing my thesis.

mdeme01:
> [Jump to label and jump to root]

Okay, I added these options.

webgui-new/.eslintrc.json
mdeme01: They have been added. See the `eslint.config.js` file. The most important React rules are also configured by default by the plugin. The `.eslintrc.json` file was generated automatically when running `lint`. I'm assuming it is required for Next.js.

webgui-new/next.config.js
mdeme01: This is only a prefix for Next.js requests and it does not effect the webserver path. It is required because the exported contents will be in the `/new` folder of the webserver (in this case a symbolic link). So Next.js needs to know to look for the files in this folder, and not at the root of the webserver.

mcserep:
> > [Jump to label and jump to root]
> 
> Okay, I added these options.

Thanks, your work fast, I will check them today 😃

webgui-new/next.config.js
mcserep: No, it does effect the webserver path. Eg. on the *codecompass.net* server we are serving the new frontend on the https://codecompass.net/demo/new/ path. If these variables have the value `/new` before `npm run build`, then the frontend will not work, because it assumes that the base path is `https://codecompass.net/new/`, which is incorrect. Therefore we had to modify the values of `assetPrefix` and `basePath` to `/demo/new` and rebuild the frontend.

Therefore the values here should not be hardcoded. If a relative path could be supported, that would be the best. Otherwise it should be configurable through an environment variable (usually named `PUBLIC_URL` in React apps).

webgui-new/InstallReact.cmake
mcserep: Try building the Docker images 😃 

According to the workflow, first, CodeCompass will be installed into the `/CodeCompass-install` folder during the build stage:
https://github.com/Ericsson/CodeCompass/blob/master/docker/runtime/Dockerfile#L31

Then, it will be moved to the `/codecompass` folder when the runtime image is constructed:
https://github.com/Ericsson/CodeCompass/blob/master/docker/runtime/Dockerfile#L82

Therefore the `new` symlink in the `webgui` folder will be invalid, because it was created using an absolute path. I had to fix that manually in the Docker image, so it works on codecompass.net. Using a relative path for the *symlink* would solve this issue. (Of course in the long run, there will be only one web GUI, but until then.)

mcserep:
**[Docker image build]**

The docker images cannot be built currently, as a newer NodeJS version is required to build the new frontend. @intjftw fixed it during the deployment to *codecompass.net* and promised to push her changes to your branch. I open this thread, so we won't forget about it.

webgui-new/next.config.js
mdeme01: I see. I could not test it with a webserver that contains additional paths, so I did not know about this. In that case, I will add another `.env` variable for this, I don't think relative paths are supported here.

webgui-new/InstallReact.cmake
mdeme01: I see, thanks for clarifying. Altough I'm a little unsure how I would use a relative path here. Could you elaborate on that?

webgui-new/.eslintrc.json
mcserep: I see, I was not familiar with this *new* ESLint configuration file format.
https://eslint.org/docs/latest/use/configure/configuration-files-new

Then, it is fine, but I think, the old config file (`.eslintrc.json`) should be removed.
https://eslint.org/docs/latest/use/configure/configuration-files

It is strange that there are 2 config files (old and new format) with different content. The `next/core-web-vitals `ruleset` should be added to the new config file as well I guess.

webgui-new/src/pages/project.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Welcome } from 'components/welcome/welcome';
+import { UserGuide } from 'components/user-guide/user-guide';
+
+interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  value: number;
+}
+
+const OuterContainer = styled('div')({
```
mcserep, line 21: This will be a general remark for the whole codebase.
I see you use this `styled()` function [provided by the MUI library](https://mui.com/system/styled/). My concern is that this way the styling is defined in the TypeScript code. I think separating the CSS styling into separate files would be a better, more robust and easier to maintain approach.

As I see, there is  [interoperability](https://mui.com/material-ui/guides/interoperability/) provided by the MUI library with e.g. CSS modules. Would it be possible to refactor the codebase to use CSS modules instead of inline CSS stylesets?

webgui-new/src/service/git-service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+export const getBlobOidByPath = async (repoId: string, hexId: string, path: string) => {
+  if (!client) {
+    return '';
+  }
+  try {
+    return await client.getBlobOidByPath(repoId, hexId, path);
+  } catch {
+    return '';
+  }
```
mcserep, line 136: This will be a general remark for the complete codebase.
I see that in the service layer you basically hide all errors. If anything goes wrong a default empty string, empty array or a false value is returned.

I think some kind of better error handling should be implemented: there is no much point for exceptions if we are simply catching them all and do nothing. Of course the webapp should not crash if e.g. a network request failed due to some network outage, but the user should be notified in some way that the requested data was e.g. not loaded and returned empty result is not the real result.

webgui-new/src/enums/search-enum.ts
```diff
@@ -0,0 +1,62 @@
+export enum SearchMethods {
+  EXPRESSION = 'Search by expression',
+  FILE_REGEX = 'File name filter regex',
+  PATH_REGEX = 'Path filter regex',
+}
```
mcserep, line 5: Using the English labels as the enum values is not really a nice solution here. Currently only the English language is supported and we have no near future plans for supporting other languages, but it would have been still a nicer solution to place these into a real translation file and use [react-i18next](https://react.i18next.com/) for example for internationalization.

webgui-new/src/pages/project.tsx
mdeme01: The main advantage of using Material UI's `styled` function is that you can utilize the styles and colors declared in the `theme.ts` file, and the colors automatically update if the theme is switched to dark/light. So you don't have to check seperately that the theme is light or dark, you can just use for example `theme.colors.primary`, because it is defined for both light and dark modes using MUI's `createTheme()` (see `theme.ts`).

As this is not possible to do with regular CSS, I don't think it is a good idea to use CSS modules instead of MUI's built-in functions, as MUI styled components provide a better abstraction for styling, in my opinion.

However, I can move all of these elements created with `styled` to a separate file for each component, e.g. `styled-components.ts`, give them a more consistent name, and import them from there for each component. This way they would be better separated, like CSS modules. At some places, I used the `sx` prop for styling, which is the same as the inline `style` prop in CSS, but you can use all of MUI's themes and functionalities. In some cases, this is mandatory to use, like when states determine some of the styling, so these cannot be separated.

webgui-new/src/service/git-service.ts
mdeme01: Yes, this was only a temporary solution to avoid crashes, which I wanted to address later. The `react-toastify` package (https://www.npmjs.com/package/react-toastify) could be used to show an error message on the page for the user, while logging the error message to the console using `console.error`.

webgui-new/.eslintrc.json
mcserep: > It seems Next.js can only use the .eslintrc.json config file, because it was automatically created when I ran lint. The new eslint.config.js file was overriden by this file, and I couldn't see the errors. So I added the react plugin to the .eslintrc.json file, and removed the new config file.

Good thing this turned out in time! The existence of 2 config files was strange fortunately 😄 

> This way, the errors are visible. There were errors in almost every file, because ESLint did not recognize React and the JSX namespace. As a fix I had to import 'React', and remove the 'JSX.Element' type enforcement (it was not required anyway, because they are already set to that type).

Cannot this be fixed by adding `"JSX": true` to the `globals`? See e.g. here:
https://gitlab.com/tms-elte/frontend-react/-/blob/412cf5a123249db0b0ddc4585743050f5385bdf6/.eslintrc.json#L60
I think that should be the real fix, not removing the valid type enforcement from the code!

Also maybe `@typescript-eslint` and `plugin:`react-hooks` should be considered for linting. (Or some of their rules.)

webgui-new/README.md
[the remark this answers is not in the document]
mcserep: Great! I will take a look on this later, whether it could work with NextJS's own webserver, but the provided solution is fine already.

webgui-new/README.md
mcserep: The value for `NODE_ENV` should be injected during production build or when running the webserver. (Or at least it works like that with vanilla React without NextJS.)

How come it was accessible previously but not now?
Can you link the related part of the codebase?

webgui-new/InstallReact.cmake
mcserep: Well, instead of `${INSTALL_WEBROOT_REACT_DIR}/out`, it should be simply `../webgui-new/out`.

mcserep:
> > [Persistent URLs]
> 
> A part of this feature is already supported. You can right click on a source element in the code and click the 'Get permalink to selection' option. This way, you can send it to someone, and the app will initialize the states based on the URL, jumping to that project, file and selection.
>
> I handled persistence using the local storage, and not the URL, and unfortunately, I realized later on that the old UI uses the URL for this. To fully support this feature, it would mean a lot of refactorings in the code.
> 
> Time is getting short before my thesis deadline (less than 2 weeks), so I would like to not start into larger refactorings, only fix smaller issues, and focus on writing my thesis. Altough I will check how many changes it would require and try to do something about it if I have the time.

Okay, it sounds like an acceptable solution for now, at least for your thesis submission.
One remark: upon opening such a permalink, the selected part of the code is at the very bottom of the codespace. I think it would be better if it would be on top, because people will most likely look there at first.

webgui-new/src/service/git-service.ts
mcserep: That sounds good!

webgui-new/src/pages/project.tsx
mcserep: While mixing TypeScript and CSS in such a way is a bit odd for me, I accept that it is the preferred and advised solution by MUI. The suggested separation to `styled-components.ts` files should be done though, so that at least there is a logical separation.

webgui-new/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "ts-node src/server.ts",
+    "build": "next build && next export",
+    "start": "NODE_ENV=production ts-node src/server.ts",
+    "lint": "next lint",
+    "thrift-ts": "thrift-typescript --fallbackNamespace none"
+  },
+  "devDependencies": {
+    "@creditkarma/thrift-typescript": "^3.7.6",
```
mcserep, line 13: This is a note to myself, to check whether the already used (and *official*) Thrift library why cannot generate the expected TypeScript files.

This should not affect your thesis submission, though, submitting with this solution is also fine.

webgui-new/README.md
mdeme01: Previously, it was accessible because it was passed as a runtime variable using `publicRuntimeConfig` in `next.config.js`. As that had to be removed, it could no longer be injected, because the client side of the application cannot access environment variables (it will throw the `cannot resolve module` error if you try to use `dotenv`).

However, Next.js has a solution for this, the variables prefixed with `NEXT_PUBLIC_` will be accessible. But you cannot prefix `NODE_ENV`, as it is a built-in variable.

But this no longer applies, as now the webserver URL is set as the current URL both in production and development, because the requests are proxied to the actual webserver in development.

webgui-new/src/pages/project.tsx
mdeme01: I separated them to different files. For the pages, I created files in the `theme` folder and named them accordion to the page. (e.g. `project-styles.tsx`)

webgui-new/InstallReact.cmake
mdeme01: I see, and should `${INSTALL_WEBROOT_DIR}/new` also be changed to `../webgui/new` then?

webgui-new/.eslintrc.json
mdeme01: Yes I tried setting `globals` before, apparently, I set `jsx` and not `JSX`, now it works, thanks for providing an example. I added back the `JSX.Element` type enforcements, and also added the `@typescript-eslint` and `react-hooks` plugins and fixed the warnings related to them.

webgui-new/InstallReact.cmake
mcserep: No, `${INSTALL_WEBROOT_DIR}/new` is a correct, valid absolute path, which will only be used when creating the symlink. Only `${INSTALL_WEBROOT_REACT_DIR}/out` should be changed to a relative path, expecting that the `webgui` and the `webgui-new` folder will always be next to each other, so it will work when their parent directory is moved / renamed.

webgui-new/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+        public src
+        DESTINATION ${INSTALL_WEBROOT_REACT_DIR}/app
+        USE_SOURCE_PERMISSIONS
+        FILES_MATCHING PATTERN "[^.]*")
+
+install(FILES
+        tsconfig.json
+        next.config.js
+        thrift-codegen.sh
+        DESTINATION ${INSTALL_WEBROOT_REACT_DIR}/app)
```
mcserep, line 11: The `.env` file should also be copied, as it contains the `PUBLIC_URL`.

mcserep:
@mdeme01  I have fixed and refactored the outdated CI in #596, please rebase this PR onto master (or merge it), so the pipeline can be executed properly.

mdeme01:
> @mdeme01 I have fixed and refactored the outdated CI in #596, please rebase this PR onto master (or merge it), so the pipeline can be executed properly.

@mcserep For some reason, the CI wants to parse everything inside `node_modules` when parsing CodeCompass.

mcserep:
> Maybe because it parses CodeCompass from the `install` directory, which will contain `node_modules` by the time the CI gets to that step? Shouldn't the CodeCompass project be cloned to a separate directory, like the others, and parsed from there?

No, it parses CodeCompass from the `${{github.workspace}}` directory, but the `build` and the `install` directory are subdirectories for that. We could move the `build` and the `install` directory outside, but it should also be checked how this affects detecting the build source and target relations during C++ parsing?

(Making a separate clone of CodeCompass does not really help, because in that case we have the same problem: where to build it: into a subdirectory or not?)

mcserep:
> No, it parses CodeCompass from the `${{github.workspace}}` directory, but the `build` and the `install` directories are subdirectories for that. We could move the `build` and the `install` directory outside, but it should also be checked how this affects detecting the build source and target relations during C++ parsing? (It could still work based on the `compile_commands.json` file.)

@mdeme01 After some discussion, it should work with moving the build and install directory outside the project folder in the CI. The build targets needs to exist for a full, proper C++ parse, but it will find them based on the compilation commands.
@intjftw started to refactor it accordingly.

mcserep:
> @mdeme01 After some discussion, it should work with moving the build and install directory outside the project folder in the CI. The build targets needs to exist for a full, proper C++ parse, but it will find them based on the compilation commands. @intjftw started to refactor it accordingly.

@mdeme01 Please update again, as #602 have been merged.

intjftw:
@mdeme01 The user documentation in your thesis is good enough to be included in the official documentation as a user guide for the UI (which is currently missing). Please add the text to the docs.

mdeme01:
> @mdeme01 The user documentation in your thesis is good enough to be included in the official documentation as a user guide for the UI (which is currently missing). Please add the text to the docs.

Do you mean add it to the [README](https://github.com/Ericsson/CodeCompass/blob/master/doc/usage.md) as well? Because I already added it to the [UI](https://codecompass.net/demo/new/project/), in the `User guide` tab.

mdeme01:
Updates on development:

- Now persistent URL parameters are fully supported, global states are set accordingly to the URL parameters.
- Refactored git blame to not be in the gutter, but in a separate container next to the code, as the functionality of the gutter was unpredictable and faulty. Also added tooltips to the commit messages.
- Refactored CodeBites: the layout is now using a more consistent tree layout, the nodes can now be deleted, and the code is a bit more polished.
- Fixed a few smaller bugs

mcserep (approved):
Looks good, the remaining few issues could be fixed in upcoming separate tickets.
Congrats! :trophy:

webgui-new/README.md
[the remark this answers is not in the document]
mcserep: @mdeme01 Well, I also found it hard to integrate `next-http-proxy-middleware`, so your proposed solution with a [Custom Server](https://nextjs.org/docs/pages/building-your-application/configuring/custom-server) should be fine, it is an officially documented and advised solution.

webgui-new/package.json
mcserep: Created #606 to address this later.

webgui-new/src/enums/search-enum.ts
mcserep: @mdeme01 Created #607 to address this refactoring issue.

Issue #613, opened 2023-07-08
Make labels dynamic on New Frontend


mcserep:
When parsing a project, users can define arbitrary labels (e.g. source). They can also be added later through the `project_info.json` file from each project. Currently the *New Frontend* has a hardcoded "Jump to source" label, instead of dynamically reading and displaying all labels. (Which even displayed and does not work, when there is no source label at all.)

mdeme01:
Are all labels used to jump to a directory in the file manager? (Also, I thought that the source label is mandatory, because it is required for the git blame to work.)

mcserep:
When executing `CodeCompass_parser`, you may mark any subdirectory as a
label by adding it with a `--label labelname=/path/to/folder` flag.

These are then stored in the `project_info.json` file in the workspace
directory and can be also edited adterwards, if needed.

The source label is not mandatory. Why do you think it is required for the
git blame view? Git repositories are detected by the parser by the presence
of the `.git` folder through an iterative directory walk, and not by the
labels. A project can even contain multiple Git repositories and then the
git plugin will parse all of them. The metadata of the found git
repositories are persisted in the workspace folder.

mdeme01:
I thought you needed the path to the source directory for getting the blame info, but I just found out that there is a `repoPath` property you can use, I fixed it. Thanks for clarifying.

Issue #611, opened 2023-07-08
Welcome page


intjftw:
The welcome page is quite empty right now. It should contain some welcome message with maybe some highlights in CodeCompass.

mcserep:
For now I think the *Flash* highlights could be copied, than later replaced with the *Gershwin* highlights.

Issue #609, opened 2023-07-08
Metrics rectangles shouldn't move


intjftw:
It is quite confusing that upon asking for metrics to a directory, the rectangles float to their place. They should appear in their final place right away.

Issue #608, opened 2023-07-08
Parse status button to show build info


intjftw:
In the old version of the UI, the parse status button on top of an open file could be clicked. On right click, a pop-up window appeared with a table of the warnings and errors given by the compiler during parsing.

mdeme01:
Could you show me an example where I can see how these build logs are shown? So far, all the files I clicked had the 'No build logs for this file' message.

intjftw:
[This file](https://codecompass.net/demo/#wsid=CodeCompass&fid=11852034569623368286&select=1|1|1|1&center=text) has a couple of them (`plugins/cpp/service/src/cppservice.cpp` in CodeCompass).

PR #615, opened 2023-07-10
New frontend remaining issues

mdeme01:
This PR aims to fix the remaning issues related to the new frontend. @mcserep @intjftw 

Issues fixed:

- #613
- #611
- #609
- #608
- #607
- #605
- #604


mcserep:
@mdeme01 Thanks for the quick work, we will review it! For the remaining issues, please create a PR / issue, so the review process is simplified. (The changeset is smaller and it is more clear what was changed and for what reason.)

mdeme01:
> * Metrics are not showing now at all, I get this error in the dev console if I click on metrics for any directory

This problem occurs because you did not specify an 'src' label that is the path to the source directory in the codebase of the project. Because all of the projects had this label on the demo website, I thought this label was mandatory for every project, apparently, it is not.

I used the 'src' label for the Metrics, because if you click the root directory, or any directory above the source folder, it still generates the Metrics for the source folder, as seen on the demo website, and I took examples from there during development.

The problem is, if you click a sub-folder in the treemap (represented by a rectangle), you have to know its full path to get information about it and display the new treemap for it - and the source folder's path is required for this, in my solution. Unfortunately, I cannot provide a better solution without understanding how the Metrics generation in the old frontend works, which I could not figure out from the codebase.

mcserep:
> This problem occurs because you did not specify an 'src' label that is the path to the source directory in the codebase of the project. Because all of the projects had this label on the demo website, I thought this label was mandatory for every project, apparently, it is not.

Unfortunately the existence of the `src` label is not mandatory, as discussed in #613 previously.

> I used the 'src' label for the Metrics, because if you click the root directory, or any directory above the source folder, it still generates the Metrics for the source folder, as seen on the demo website, and I took examples from there during development.
> 
> The problem is that generating metrics when clicking a sub-folder (represented by a rectangle in the treemap) is dependant on the path of the folder, because it works by appending/removing the folder name to the path. And if you generate the metrics for anything outside the source folder, the result will still be the metrics for the source folder, and the path will be out of sync.
> 
> For example, generating it for the root folder `/`, will result in the metrics of `/projects/CodeCompass`, and the path will still be `/`. Now if I click on a sub-folder, e.g. `webgui`, the input for the metrics will be `/webgui`, but it should be `/projects/CodeCompass/webgui`.
> 
> So this is why, when clicking a directory outside the source folder, I appended the source folder path to the current path (stored in the 'src' label), because I did not know how else to solve this. (And I could not figure out how this works in the old frontend.)

To generate the metrics for a file or a directory:
1. First you should fetch the `fileID` based on the path. This could be done with the `getFileInfoByPath()` endpoint in the `projectservice`.
2. Then the `getMetrics()` in `metricsservice` will expect a file ID as a parameter.

This is how it is done in the current UI.

mcserep:
#613 (*Make labels dynamic on New Frontend*): the labels are recognized and displayed correctly on the new UI, but clicking them raises an error in the console log:
> Invalid path: /home/mate/cc/sources/tinyxml2/

The path is valid and correct, the label also works on the old UI correctly.

mcserep (changes requested):
Apart from the mentioned issues with the labels, the changes look good to me and functionally sound.

mdeme01:
> To generate the metrics for a file or a directory:

> First you should fetch the fileID based on the path. This could be done with the getFileInfoByPath() endpoint in the projectservice.
> Then the getMetrics() in metricsservice will expect a file ID as a parameter.
> This is how it is done in the current UI.

Yes, I know that. I explained the problem above. I'm sorry if it wasn't clear, I will try to explain it again.

If you click a directory for metrics, you need its ID, this much is correct and clear, and it works that way correctly. However, after the metrics has been generated, you need to be able to generate metrics for other directories, displayed in the treemap. The treemap is generated from the metrics result, which only contains the **names** of the directories, **not their path**. You should be able to generate metrics for a directory by clicking on it inside the treemap.

This operation requires a new path, for a new metrics, but you can only work with the name of the clicked directory, you don't know its path. But you have the path of the directory you initially generated the metrics for, so you can just append the name of the directory to this path.

For example: you generated metrics for the `CodeCompass` folder, its path is `/projects/CodeCompass`, which is a fully valid path. The treemap will contain the `webgui` directory, represented by a rectangle. Now you want to generate metrics for this directory, and you can do this by appending `webgui` to the current path (`/projects/CodeCompass`), and the new path will be `/projects/CodeCompass/webgui`, which is valid, thus the new metrics can be generated correctly.

The problem comes in when you generate metrics for anything **outside the source folder**, like the `/` folder, or, `/projects`. Appending the names of the directories will not work here, because the paths will be invalid. Using the same example, you want to generate metrics for `webgui`, the current path is `/`, and appending will result in `/webgui`, which is an invalid path, and the metrics generation will fail.

For some reason `getMetrics` returns the result of the source directory for directories outside the source folder. And that is why,  the current path and the result will be out of sync. So I needed to handle this by using the `src` label: I used the source path for directories outside the source folder. What I do not know, is how this is handled in the old frontend, **without** using the `src` label.

mdeme01:
> The labels are recognized and displayed correctly on the new UI, but clicking them raises an error in the console log

I checked it, for me it works correctly. I used the cJSON project and specified 3 labels.

I used the following command: (In my case, the project is located in `/home/deme/CodeCompass/projects`)

```
./CodeCompass_parser \
-d "sqlite:database=~/CodeCompass/database/cjson.sqlite" \
-w ~/CodeCompass/test_ws/ \
-n cJSON/ \
-i ~/CodeCompass/projects/parsed/cjson/compile_commands.json \
-i ~/CodeCompass/projects/cjson/ \
-j 4 \
--label src=/home/deme/CodeCompass/projects/cjson \
--label tests=/home/deme/CodeCompass/projects/cjson/tests \
--label fuzzing_inputs=/home/deme/CodeCompass/projects/cjson/fuzzing/inputs
```

All 3 labels are displayed and clicking them navigates to the correct directory. Are you sure the path you specified is correct for the that project?

mcserep:
> Are you sure the path you specified is correct for the that project?

Yes, it works on the old UI as well. I will try to debug this later.

mcserep:
> Yes, it works on the old UI as well. I will try to debug this later.

@mdeme01  I have debugged this and the source of the error is that in the DB the path is:
`/home/mate/cc/sources/tinyxml2`
while the configured label is:
`/home/mate/cc/sources/tinyxml2/`

Note the trailing slash. Therefore `ProjectServiceHandler::getFileInfoByPath` returns that the path was not found. In the old UI it works regardless of the trailing slash, so it could be checked how it is done there.

mcserep:
> Yes, I know that. I explained the problem above. I'm sorry if it wasn't clear, I will try to explain it again.
> 
> If you click a directory for metrics, you need its ID, this much is correct and clear, and it works that way correctly. However, after the metrics has been generated, you need to be able to generate metrics for other directories, displayed in the treemap. The treemap is generated from the metrics result, which only contains the **names** of the directories, **not their path**. You should be able to generate metrics for a directory by clicking on it inside the treemap.
> 
> This operation requires a new path, for a new metrics, but you can only work with the name of the clicked directory, you don't know its path. But you have the path of the directory you initially generated the metrics for, so you can just append the name of the directory to this path.
> 
> For example: you generated metrics for the `CodeCompass` folder, its path is `/projects/CodeCompass`, which is a fully valid path. The treemap will contain the `webgui` directory, represented by a rectangle. Now you want to generate metrics for this directory, and you can do this by appending `webgui` to the current path (`/projects/CodeCompass`), and the new path will be `/projects/CodeCompass/webgui`, which is valid, thus the new metrics can be generated correctly.
> 
> The problem comes in when you generate metrics for anything **outside the source folder**, like the `/` folder, or, `/projects`. Appending the names of the directories will not work here, because the paths will be invalid. Using the same example, you want to generate metrics for `webgui`, the current path is `/`, and appending will result in `/webgui`, which is an invalid path, and the metrics generation will fail.
> 
> For some reason `getMetrics` returns the result of the source directory for directories outside the source folder. And that is why, the current path and the result will be out of sync. So I needed to handle this by using the `src` label: I used the source path for directories outside the source folder. What I do not know, is how this is handled in the old frontend, **without** using the `src` label.

The `getMetrics` returns the metrics for the files in the specified directory. The result contains the absolute path for the files (as nested JSON objects), therefore I do not really get why we would append `webgui` to `/` in your example.

The old UI simply does a smart move and if the metrics for `/` folder was requested, but all returned metrics are for files in the `/projects/CodeCompass` folder, then in the visualized treemap the `/projects/CodeCompass` becomes the top folder.

mdeme01:
> The result contains the absolute path for the files

Okay, I misunderstood it, and I didn't realize it when I first implemented it. Thanks for clarifying. Now I think I was able to fix it, at least it is working for me, so it does not require the `src` label anymore.

mcserep:
> Okay, I misunderstood it, and I didn't realize it when I first implemented it. Thanks for clarifying. Now I think I was able to fix it, at least it is working for me, so it does not require the `src` label anymore.

Thanks for the update and the fix, it works now for me as well similarly like in the old UI.

> @mdeme01 I have debugged this and the source of the error is that in the DB the path is:
> /home/mate/cc/sources/tinyxml2
> while the configured label is:
> /home/mate/cc/sources/tinyxml2/
> 
> Note the trailing slash. Therefore ProjectServiceHandler::getFileInfoByPath returns that the path was not found. In the old UI it works regardless of the trailing slash, so it could be checked how it is done there.

This is the only remaining issue I see before merging this PR.

mdeme01:
> Note the trailing slash. Therefore ProjectServiceHandler::getFileInfoByPath returns that the path was not found. In the old UI it works regardless of the trailing slash, so it could be checked how it is done there.

I didn't explicitly find out how this is handled in the old UI. However, I added a check for the trailing slash, and if the path contains one, it passes the path without it to the request, and this seems to be working.

mcserep:
> I didn't explicitly find out how this is handled in the old UI. However, I added a check for the trailing slash, and if the path contains one, it passes the path without it to the request, and this seems to be working.

The old UI processes the path by elements, splitting it by the directory separators. See in `fileManager.js`:
```js
      var path = shortcut.split('/');
      path[0] = '/';

      path.forEach(function (directory) {
        if (!directory.length) // <-- this ignores the last element if empty
          return;

        // ... 
      });
```

Since it is not required in the new UI, a simple check as you have implemented should be fine.

mcserep (approved):

PR #624, opened 2023-08-10
Upgrade NodeJS and NPM for tarball build

mcserep:
The new web GUI officially requires NodeJS v18+, while in the Ubuntu 16.04 based tarball build script we have NodeJS 14 installed, which fails.

Unfortunately NodeJS v18 is not compatible with Ubuntu 16.04. (GLIBC mismatch error.)
This PR upgrade NodeJS from v14 to v16 and NPM from v6 to v8 in tarball CI build script and hopefully this will be enough.

@mdeme01 Do we really need NodeJS v18 for the new frontend to build, or was it specified simply as the current LTS version?

@intjftw, @zporky Do we still need these legacy tarball builds in Ericsson?

### DIFF
--- a/.gitlab/build-deps.sh
+++ b/.gitlab/build-deps.sh
@@ -377,12 +377,12 @@ rm -f $PACKAGES_DIR/release-1.10.0.tar.gz
 #######
 
 cd $PACKAGES_DIR
-wget --no-verbose --no-clobber https://nodejs.org/dist/v14.15.0/node-v14.15.0-linux-x64.tar.xz
-tar -xf node-v14.15.0-linux-x64.tar.xz -C $DEPS_INSTALL_RUNTIME_DIR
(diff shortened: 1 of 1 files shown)


mdeme01:
> @mdeme01 Do we really need NodeJS v18 for the new frontend to build, or was it specified simply as the current LTS version?

I tested it and I didn't have any problems installing packages and building the application with Node v16, so it should be fine. I've only written v18 because it was the current LTS version at the time.

mcserep:
> I tested it and I didn't have any problems installing packages and building the application with Node v16, so it should be fine. I've only written v18 because it was the current LTS version at the time.

@mdeme01 Thanks for the quick response, I think we shall keep Node v18 in the documentation as a requirement, as the end of support for v16 is coming soon. But hopefully we can make this tarball build with it to support legacy operating systems in Ericsson.

Let's see whether pipeline succeeds :smile:

PR #627, opened 2023-09-07
LLVM 11 migration

wbqpk3:
- Updated source code to support LLVM 11

In `clang::CreateASTDumper` function `DumpDeclTypes` seems new functionality, thus it was changed to `false`.
See:
https://github.com/llvm/llvm-project/blob/llvmorg-11.1.0/clang/lib/Frontend/ASTConsumers.cpp#L98
https://github.com/llvm/llvm-project/blob/llvmorg-10.0.0/clang/lib/Frontend/ASTConsumers.cpp#L81

- Added missing headers for g++-11 (default g++ on Ubuntu 22.04)
- Updated documentation

### DIFF
--- a/.github/scripts/setup_build_ubuntu-20.04.sh
+++ b/.github/scripts/setup_build_ubuntu-20.04.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 # Install required packages for CodeCompass build
-sudo apt-get install -y git cmake make g++ libboost-all-dev llvm-10-dev clang-10 \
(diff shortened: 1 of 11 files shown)


mcserep:
The CI also needs to be updated to LLVM and Clang 11.
See the following files:
 - `.github/scripts/setup_build_ubuntu-20.04.sh`
 - `.github/scripts/setup_runtime_ubuntu-20.04.sh`
 - `.gitlab/build-deps.sh`

mcserep:
The Dockefiles also had to be updated, related change to this MR in 354895724db9bd5c39ea043a73e9b168574ee960.

PR #634, opened 2023-09-30
Workspace and root files are now ordered by name.

### LINKED ISSUE #626: Sort projects alphabetically
Both on the old and the new UI projects are listed in *some* order, but not sorted in any way. This can make to locate a specific project if there are many parsed. Projects should be listed alphabetically either on the frontend or in the backend, before returning the list. Most likely the latter is the preferred solution.

dbukki:
The issue originally mentions only workspaces not being in the correct order. But I noticed that root files in the file browser pane were also not ordered, and the fix is very similar in both cases (missing std::sort).

Fixes #626.

### DIFF
--- a/service/project/src/projectservice.cpp
+++ b/service/project/src/projectservice.cpp
@@ -124,6 +124,8 @@ void ProjectServiceHandler::getRootFiles(std::vector<FileInfo>& return_)
       return_.push_back(makeFileInfo(f));
     }
   });
+
+  std::sort(return_.begin(), return_.end(), fileInfoOrder);
 }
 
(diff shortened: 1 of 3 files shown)


intjftw:
@dbukki Can you please link the corresponding issue? Writing "Fixes #<issue number>" will automatically link it.

dbukki:
Fixes #626

mcserep:
> Fixes #626

You have to place that into the initial PR description to make it work, I have updated it.

mcserep (approved):
Looks good to me! :+1:

PR #642, opened 2023-10-08
Google analytics integration in the new webgui

### LINKED ISSUE #530: GDPR compliance when Google Analytics is enabled
Using Google Analytics in CodeCompass is optional, but when enabled, we shall fulfill the GDPR regulations.

Summarized:
> To ensure that Google Analytics – its cookies, trackers and statistics tools – run in full compliance with EU’s General Data Protection Regulation (GDPR), you need to:
>  - Ask for and obtain end-user consent for all Google Analytics cookies on your website prior to their activation and operation.
>  - Control each Google Analytics cookie in order to only activate them after your users have given their explicit consent to them.
>  - Provide transparent information in your website’s cookie policy about the details of all Google Analytics cookies in operation – including their provider, technical details, duration and purpose. This is important as consent is only valid under the GDPR if it constitutes an informed choice on behalf of the users.
>  - Compile detailed information in your website’s privacy policy about all Google Analytics cookies on your domain, and what personal data your website processes in general.
>  - Turn on IP anonymization in your Google Analytics account and make sure that it uses pseudonymous identifiers.
 
Source: https://www.cookiebot.com/en/google-analytics-gdpr/

### LINKED ISSUE #616: Google Analytics events in new frontend
The previous version of the UI had Google Analytics integrated in the form of events. These are intended to capture important activity from users anonymously. We use the data for statistical purposes. The new frontend should also integrate the events. They should be featured in every possible function, such as AST node clicks, displaying git blame views, diagrams, documentation, moving between files, etc.

LoremIPsummer:
Google Analytics integration with the [react-ga4](https://www.npmjs.com/package/react-ga4) library.

The data should be collected in the same format as it was declared in #526.

I didn't implement the `info_tree` event, because on the new GUI, clicking on a word in the editor (a.k.a. firing the `click_on_word` event), automatically brings up the Info tree (and there is no dedicated context menu item for it), so it should be considered a duplicated event.

The GA script only loads when the user gives direct consent by clicking on "Accept."
If there is an error during accessing the `ga.txt` file content (it does not exist), the cookie notice won't show up.

fixes #616 
fixes #530

### DIFF
--- a/webgui-new/next.config.js
+++ b/webgui-new/next.config.js
@@ -15,14 +15,6 @@ const nextConfig = {
     assetPrefix: `${process.env.PUBLIC_URL || ''}/new`,
     basePath: `${process.env.PUBLIC_URL || ''}/new`,
   }),
-  async rewrites() {
-    return [
-      {
-        source: '/ga.txt',
-        destination: `${process.env.BACKEND_URL}/ga.txt`,
-      },
-    ];
-  },
 };
 
 module.exports = nextConfig;
--- a/webgui-new/next.config.js
+++ b/webgui-new/next.config.js
@@ -15,6 +15,14 @@ const nextConfig = {
     assetPrefix: `${process.env.PUBLIC_URL || ''}/new`,
     basePath: `${process.env.PUBLIC_URL || ''}/new`,
   }),
+  async rewrites() {
+    return [
+      {
+        source: '/ga.txt',
+        destination: `${process.env.BACKEND_URL}/ga.txt`,
+      },
+    ];
+  },
 };
 
 module.exports = nextConfig;
--- a/webgui-new/src/i18n/locales/en.json
+++ b/webgui-new/src/i18n/locales/en.json
@@ -355,5 +355,34 @@
     "INCLUDE_DEPENDENCY": "Include Dependency",
     "INTERFACE": "Interface Diagram",
     "SUBSYSTEM_DEPENDENCY": "Internal architecture of this module"
+  },
+  "cookie": {
+    "ACCEPT": "Accept",
+    "INTRO_TEXT": "Our website uses cookies to improve your experience.",
+    "LEARN_MORE": "Learn more"
+  },
+  "cookiePolicy": {
+    "TITLE": "Cookie policy",
+    "sections": {
+      "cookies": {
+        "SERVICE": "Service",
+        "PURPOSE": "Purpose",
+        "DURATION": "Duration",
+        "MORE_INFORMATION": "More Information",
+        "COOKIE_NAMES": "Cookie names",
+        "WHAT_COOKIES": "What are cookies?",
+        "WHAT_COOKIES_DESCRIPTION": "Cookies are small text files that websites store on a user's device, primarily for analytics purposes. They enable websites to gather valuable data about user interactions, helping businesses understand user behavior, preferences, and engagement. By analyzing this data, organizations can make informed decisions to improve their websites, content, and marketing strategies. Cookies play a crucial role in optimizing the online experience and tailoring content to better serve users' needs while respecting their privacy through proper consent mechanisms.",
+        "HOW_USE_TITLE": "How we use cookies?",
+        "TABLE": "The table below explains the cookies we use and why:",
+        "CONCLUSION": "By clicking \"Accept\", you agree to the use of cookies on this website, enhancing your browsing experience and enabling us to gather valuable analytics data for continual improvement. You can always withdraw your consent later on by simply removing the corresponding cookies from your device.",
+        "googleAnalytics": {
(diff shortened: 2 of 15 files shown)


webgui-new/next.config.js
```diff
@@ -15,6 +15,14 @@ const nextConfig = {
...
     basePath: `${process.env.PUBLIC_URL || ''}/new`,
   }),
+  async rewrites() {
+    return [
+      {
+        source: '/ga.txt',
+        destination: `${process.env.BACKEND_URL}/ga.txt`,
+      },
+    ];
+  },
```
mcserep, line 25: Other path rewrites for the backend are handled in `src/server.ts`. I suggest to handle this there as well.

LoremIPsummer: Fixed

webgui-new/src/i18n/locales/en.json
```diff
@@ -355,5 +355,33 @@
...
+        "COOKIE_NAMES": "Cookie names",
+        "WHAT_COOKIES": "What are cookies?",
+        "WHAT_COOKIES_DESCRIPTION": "Cookies are small text files that websites store on a user's device, primarily for analytics purposes. They enable websites to gather valuable data about user interactions, helping businesses understand user behavior, preferences, and engagement. By analyzing this data, organizations can make informed decisions to improve their websites, content, and marketing strategies. Cookies play a crucial role in optimizing the online experience and tailoring content to better serve users' needs while respecting their privacy through proper consent mechanisms.",
+        "HOW_USE_TITLE": "How we use cookies?",
+        "CONCLUSION": "By clicking \"Accept\", you agree to the use of cookies on this website, enhancing your browsing experience and enabling us to gather valuable analytics data for continual improvement. You can always withdraw your consent later on by simply removing the corresponding cookies from your device.",
+        "googleAnalytics": {
+          "SERVICE": "Google Analytics",
+          "COOKIE_NAMES": "_ga, _ga_<MEASUREMENT_ID>",
```
mcserep, line 379: Actually, there are 4 cookies placed related to Google Analytics. But do we need to list them all?

webgui-new/src/server.ts
```diff
@@ -16,9 +16,10 @@ const main = async () => {
   const proxyHandler = createProxyMiddleware({
     target: process.env.BACKEND_URL,
     changeOrigin: true,
-    pathFilter: '**/*Service',
+    pathFilter: ['**/*Service', '**/*ga.txt'],
     pathRewrite: {
       '^/[^/]+/(.+Service)$': '/$1',
+      '^/ga.txt' : '/$1'
```
mcserep, line 22: This will not work, as there are no capture groups in `^/ga.txt`, so `$1` is an invalid backreference. It should be either `'^/(ga.txt)': '/$1'`, or simply `'^/ga.txt' : '/ga.txt'`.

webgui-new/src/server.ts
```diff
@@ -16,9 +16,10 @@ const main = async () => {
   const proxyHandler = createProxyMiddleware({
     target: process.env.BACKEND_URL,
     changeOrigin: true,
-    pathFilter: '**/*Service',
+    pathFilter: ['**/*Service', '**/*ga.txt'],
```
mcserep, line 19: Maybe it is enough to filter for `ga.txt` files, and not all files ending with that suffix.

mcserep (changes requested):
The event sending works as expected, some minor remarks.

webgui-new/src/i18n/locales/en.json
LoremIPsummer: Only these two are [set directly from the `gtag.js` file](https://support.google.com/analytics/answer/11397207?hl=en). I double-checked locally and indeed that's the case. 
The other two possible cookies are used for ad interaction tracking, but those are irrelevant for us, since we don't place ads in CodeCompass.

webgui-new/src/server.ts
LoremIPsummer: Fair point. Fixed.

webgui-new/src/server.ts
LoremIPsummer: Strange. If this is not right, then the frontend should not be able to fetch the content of this file at all.

LoremIPsummer: Nevermind, after removing my local cache, I can see it is indeed wrong. Just makes me wonder why I could test the functionality with this in the first place. Fixed.

mcserep: Yes, its indeed strange that it worked, maybe some caching issue based on your previous approach.

webgui-new/src/i18n/locales/en.json
mcserep: Fine then.

mcserep (approved):

PR #643, opened 2023-10-09
Add linting to ci#603

kmkristof:
Added a job to lint the frontend code to the CI workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,11 @@ env:
 
 jobs:
  
+  ## LINTING JOBS  
+  linting:
+    uses: ./.github/workflows/linting.yml
+
+      
   ## BUILD JOBS
   build:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/linting.yml
```diff
@@ -0,0 +1,39 @@
...
+              with:
+                node-version: 20
+            
+            - name: cache npm
+              uses: actions/cache@v3
+              env:
+                cache-name: npm_cache
+              with:
+                path: ~/.npm
+                key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
```
mcserep, line 30: This only caches the global packages data I think (`~/.npm` folder), and could be achieved with the `setup-node` action, as described here:
https://github.com/actions/setup-node#caching-global-packages-data

However, in this case the packages will be installed in the local `node_modules` directory? Shouldn't that be cached as well?

.github/workflows/ci.yml
```diff
@@ -6,7 +6,12 @@ env:
   BUILD_TYPE: Debug
 
 jobs:
+ 
+  ## LINTING JOBS  
+  linting:
+    uses: ./.github/workflows/linting.yml
```
mcserep, line 12: Linting is a separate workflow, it does not expect any parameter from the `ci` workflow, nor does it create any artifact for future usage. Maybe it could be an independent workflow, executed on `push`, `pull_request` and `workflow_dispatch`, the same as the `ci` workflow.

mcserep (changes requested):

Issue #630, opened 2023-09-20
Update Thrift to 0.16


mcserep:
Thrift should be updated to 0.16 (currently 0.13 is used), so we match the version available in the official Ubuntu repositories for 22.04 Jammy.

According to the [CHANGELOG](https://github.com/apache/thrift/blob/master/CHANGES.md), multiple breaking changes occurred between Thrift 0.13 and 0.16.
 - I have made an initial attempt for migration and an *undocumented* issue is that starting from Thrift 0.14, various C++ files (like `*_constants.cpp`, `*_constants.h`, `*_types.cpp`) are [not generated](https://www.mail-archive.com/user@thrift.apache.org/msg03297.html), if they would be practically empty. This results in compilation failure and the CMake build system should be configured accordingly.
- Even after resolving this, there are API changes which should be refactored.

PR #644, opened 2023-10-10
Update docker images to Ubuntu 22.04

### LINKED ISSUE #632: Update Docker images to Ubuntu 22.04
Current Docker images depend on Ubuntu 20.04. Since we will officially support Ubuntu 20.04 for another ca. 1.5 years, it is not urgent to migrate these images to Ubuntu 22.04, but I create an issue for it.

mozesl:
All 3 docker images were updated to Ubuntu 22.04. Thrift and ODB were manually built inside the dev container and then packaged because their official Ubuntu repository versions don't work with the codebase. (Thrift can be installed from the repo once it's updated to 0.16 #630)
Looking for feedback and suggestions on how to handle the relatively big pre-built packages. Also more extensive testing.
Closes #632


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ include(Testing.cmake)
 
 find_package(Boost   REQUIRED COMPONENTS filesystem log program_options regex system thread)
 find_package(Java    REQUIRED)
-find_package(Odb     REQUIRED)
+find_package(ODB     REQUIRED)
```
mcserep, line 15: Why the change? The file is also named `FindOdb.cmake` in the root directory of the project, so it should be at least consistent in the casing.

docker/dev/Dockerfile
```diff
@@ -1,60 +1,80 @@
...
-  && apt-get -y install --no-install-recommends \
-  cmake make \
-  default-jdk \
-  ctags \
-  ca-certificates \
-  curl \
-  gnupg \
-  doxygen \
-  gcc-9 gcc-9-plugin-dev g++-9 \
+RUN set -x \
```
mcserep, line 13: In Docker, each `RUN` statement creates a separate [layer](https://docs.docker.com/build/guide/layers/). So if e.g. in one layer you add a big file to an image, and in the next layer you delete it, the final image size will be big, because a layer contains that file, regardless that it was removed later.

Therefore it was a deliberate decision to install all packages from the Ubuntu package repository in a *single step* and then in the *same* step remove the APT caches (`apt-get clean` and removing `/var/lib/apt/lists/`). This way the image will be significantly smaller, because the caches won't be added to any of the layers.

With your refactoring into separate `RUN` command and thus layers, this will not like this anymore, so I suggest to revert this.

plugins/git/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-find_package(Git2 REQUIRED)
+find_package(libgit2 REQUIRED)
```
mcserep, line 1: I see you have renamed this file to `Findlibgit2.cmake`. If that was to have the original name of the file (I don't know what it was named `FindGit2.cmake`), that should be `FindLibgit2.cmake` and you shall use:

```
find_package(Libgit2 REQUIRED)
```

docker/runtime/Dockerfile
```diff
@@ -38,51 +41,58 @@ RUN mkdir /CodeCompass-build && \
...
+
+RUN apt-get update -qq \
+  && ${INSTALL_COMMAND} \
+    /packages/libodb-2.5.deb \
+    /packages/libodb-dev.deb \
+  && if [ "pgsql" = "${CC_DATABASE}" ]; then \
+    ${INSTALL_COMMAND} \
+      libpq5 \
+      /packages/libodb-pgsql-2.5.deb \
+      /packages/libodb-pgsql-dev.deb \
```
mcserep, line 65: So you compiled these DEB packages on your computer and simply added them to the packages directory?

plugins/cpp/parser/CMakeLists.txt
```diff
@@ -1,5 +1,16 @@
...
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+if (${VERSION_RESULT} EQUAL 0)
+    if (${UBUNTU_VERSION} GREATER_EQUAL 22)
+        message(STATUS "LLVM version override to 11.1 because of Ubuntu 22")
+        set(LLVM_VERSION_TO_FIND 11.1)
+    endif ()
+endif ()
 # Find LLVM
-find_package(LLVM REQUIRED CONFIG)
+find_package(LLVM ${LLVM_VERSION_TO_FIND} REQUIRED CONFIG)
```
mcserep, line 13: This does not seem really robust. A proper solution would be simply the following:
```cmake
find_package(LLVM 11 REQUIRED CONFIG)
```

However, semantic versioning is not handled correctly, so e.g. version 11.1 does not meet the requirement  to match version 11. Since CMake 3.19, there are version ranges, so we can write this:
```cmake
find_package(LLVM 11...<12 REQUIRED CONFIG)
```

But on Ubuntu 20.04, we only have CMake 3.16 from the official package repository. So until we drop support for Ubuntu 20.04, you shall either find a more robust solution, which does not strictly ties the version requirement to `11.1`, as other minor versions are fine as well, or leave this issue (out of scope for the ticket), and the proper LLVM and Clang paths can be given to CMake with the `LLVM_DIR` and `Clang_DIR` flags, as the [build guide](https://github.com/Ericsson/CodeCompass/blob/master/doc/deps.md#build-codecompass) describes.

mcserep (changes requested):

plugins/git/CMakeLists.txt
mcserep: I have renamed it to `FindLibgit2.cmake`, as the `Git2` name could be confusing; and also, all exported symbols are prefixed with *LIBGIT2*.

docker/runtime/Dockerfile
mcserep: @mozesl I gave a second thought to this after our previous meeting, and made a little research on the topic. I found that adding DEB files to Docker images like this is not the usual approach, instead a more favorable solution would be to build ODB and Thrift as part of the Docker image (as a layer).

Of course this could effect, efficiency, but there are caching tools in Docker to tackle this problem:
- When rebuilding the images locally, the layers from the existing previous build can be reused by Docker. Only the first layer, which changes and the ones after it have t be rebuild. See the [Optimizing builds with cache management guide](https://docs.docker.com/build/cache/) for more details if you are unfamiliar with it.
- Of course in CI the individual builds happen independent of each other and the build cache is not preserved between builds. However, with the usage of either the *inline*, the *registry* or the *gha* cache, we could easily reach the same result. The only difference would be that Docker image would not get a full clean build in the CI, with its benefits and drawbacks. (We could still enforce a clean build e.g. on releases.) See the [Cache storage backends guide](https://docs.docker.com/build/cache/backends/) or [this blog post](https://lipanski.com/posts/speed-up-your-docker-builds-with-cache-from) for further details.

docker/dev/install_odb.sh
```diff
@@ -0,0 +1,40 @@
...
+  ;;
+  *)
+    bpkg build odb --yes
+    bpkg build libodb-sqlite --yes
+    bpkg build libodb-pgsql --yes
+  ;;
+esac
+bpkg install --all --recursive
+# Copy to /usr
+cp -rn /odb_install/* /usr
```
mcserep, line 36: Conventionally the `/usr` folder is for files installed through the package manager. Locally compiled programs should go to `/usr/local`.

Therefore, simply `/usr/local` should be given as the install path and no copying is needed.

docker/dev/install_thrift.sh
```diff
@@ -0,0 +1,21 @@
...
+  --enable-libtool-lock --enable-tutorial=no --enable-tests=no    \
+  --with-libevent --with-zlib --without-nodejs --without-lua      \
+  --without-ruby --without-csharp --without-erlang --without-perl \
+  --without-php --without-php_extension --without-dart            \
+  --without-haskell --without-go --without-rs --without-haxe      \
+  --without-dotnetcore --without-d --without-qt4 --without-qt5    \
+  --without-java
+make install -j $(nproc)
+# Copy to /usr
+cp -rn /thrift_install/* /usr
```
mcserep, line 18: Conventionally the `/usr` folder is for files installed through the package manager. Locally compiled programs should go to `/usr/local`.

Therefore, simply `/usr/local` should be given as the install path and no copying is needed. If you don't define a `--prefix` for the `./configure`, that will be the default value.

mcserep (changes requested):
Thanks for updating the MR! Only some minor remarks left.

mcserep:
The DEB files should be removed from the PR. Apart from that, it looks good and ready to merge :rocket:

docker/dev/install_odb.sh
mcserep: We have discussed on the weekly meeting, that this could also be a valid approach. However, I have further doubts, as the install folder could also contain symlinks. If those are not relative, but absolute, copying them would result in invalid links. Therefore installing directly into `/usr/local/` would still be preferred.

docker/dev/install_thrift.sh
mcserep: We have discussed on the weekly meeting, that this could also be a valid approach. However, I have further doubts, as the install folder could also contain symlinks. If those are not relative, but absolute, copying them would result in invalid links. Therefore installing directly into `/usr/local/` would still be preferred.

mcserep (approved):

PR #645, opened 2023-10-11
Ast visitor function for using declarations. 

### LINKED ISSUE #305: "using" declarative is not handled
The namespace "using" declarative is not handled in the C++ parser:
```.cpp
namespace N {}
void f()
{
  using namespace N; // "N" is missing from the database
}
```

schaumb:
Based on https://github.com/Ericsson/CodeCompass/pull/487, it resolves #305 

- [x] Fix tests
- [x] Add more test
- [ ] Variable template ⌧ - postponed #650

Manually:
- [x] Test with multiple namespaces
- [x] Test with multiple files with the same namespace
- [x] Test with nested namespaces
- [x] Using namespace with rename - OK, but  `namespace Cica = ::MyNamespace;` fails
- [x] Using class with rename (works as typedef)


plugins/cpp/model/include/model/cppastnode.h
```diff
@@ -57,6 +57,7 @@ struct CppAstNode
     LocalTypeLoc,
     TypedefTypeLoc,
     InheritanceTypeLoc,
+    Resolution,
```
whisperity, line 60: These nodes are created from Shadow Decls. So why are they called "resolution"?

schaumb:
> In general, I am fine with this. However, there is something I really do not understand in the tests. So, there is a function call to `g()` in the test function. That call _should_(?) resolve to the `void g() {}` as defined on line 21. But nowhere in the tests does the number `21` appear... I might just be seeing ghosts here because something is indexed from 0 instead of 1 (the number `20` appears as a line number in the test!), but it's nevertheless curious.

Hi @whisperity , 
The test is about `using`, but none `using` directives use the global `g` function. 
This free `g` function tests that `using MyNamespace::g`.line does not link to all `g` functions and the global `g` function call is not linked to the `using` line. That's why there are no explicit tests for the `g` function

plugins/cpp/model/include/model/cppastnode.h
schaumb: These are not declarations based on the user.
So the `using MyNamespace::g;` line does not use and does not declare the `g` function.

mcserep: The name *Resolution* seems quite arbitrary, and does not refer in any way that we are handling using statements with them, yet they are storing the location for the using statements. Maybe `UsingLoc` ?

schaumb: I renamed it to UsingLoc.

mcserep (approved):
Great work @schaumb! 🏆

PR #646, opened 2023-10-14
Correct display for implicit and macro-expanded variables

### LINKED ISSUE #597: Local variables are incorrectly listed in info tree
In the C/C++ info tree, the local variables of a method are either incorrectly listed (several false or empty entries), or not listed at all. We should investigate if this is a problem in the web UI or in the backend.

Screenshot from 2023-05-03 21-56-31

dbukki:
Implicit local variables in functions are now marked with the Implicit flag (the same way compiler-generated methods are marked).

Class members and local variables expanded from macros now display the correct text (which is always the type + the name) and jump to the location of the macro's expansion.

Fixes #597.

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -1258,13 +1258,15 @@ private:
 
     clang::SourceLocation realStart = start_;
     clang::SourceLocation realEnd = end_;
-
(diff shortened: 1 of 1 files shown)


mcserep (approved):
Seems good :clap:  During testing, I found another issue with local variables, but verified that is unrelated to this PR, so I will open a separate ticket for it.

Issue #583, opened 2022-10-03
Clicking any node inside a lambda does not work


intjftw:
If you click on a node inside a lambda expression, the lambda declaration will be highlighted and marked (showing the lambda in the info tree) instead of the node itself. See the image for an example with `getFile` clicked.
![Screenshot from 2022-10-03 11-58-18](https://user-images.githubusercontent.com/26124787/193550695-cb942174-d8e1-4de1-b2d7-e1c76a5db495.png)

dbukki:
I found that the issue is with the fact that since commit 3bc3c7f1f69c4a1a0680dc289b50d835d7cf74a1, implicitness is hereditary for child nodes during traversal via the _isImplicit member variable. I guess the original assumption was that once a node is implicit, everything in it must surely also be implicit. But this is simply not the case with lambdas: While the closure type itself is in fact implicit, its operator() is not, and yet it is treated as such. This marks everything inside a lambda point to unspecified locations, because _fileLocUtil.setRange is never called for any of its child nodes.

PR #654, opened 2023-10-22
Destructor usage is now properly registered

### LINKED ISSUE #306: Destructor call is not present on delete expressions.
When allocating an object on the heap, the corresponding constructor call is also registered at that position besides the `operator new()` call. However, in case of a `delete` expression only `operator delete()` is registered, but the destructor isn't.

dbukki:
Currently supported places of usage:
- In delete expressions (for heap-allocated variables).
- At the end of the parent scope (for stack-allocated local variables).
- At the end of the destructor of enclosing record types (for member variables).

Fixes #306.

Only one question so far at line 947: is there a better solution than this?

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -435,15 +435,12 @@ public:
     cppRecord->entityHash = astNode->entityHash;
     cppRecord->name = rd_->getNameAsString();
     cppRecord->qualifiedName = rd_->getQualifiedNameAsString();
-    if (const clang::CXXRecordDecl* crd
-        = llvm::dyn_cast<clang::CXXRecordDecl>(rd_))
-    {
-      cppRecord->isAbstract = crd->isAbstract();
-      cppRecord->isPOD = crd->isPOD();
-    }
 
     if (clang::CXXRecordDecl* crd = llvm::dyn_cast<clang::CXXRecordDecl>(rd_))
     {
+      cppRecord->isAbstract = crd->isAbstract();
+      cppRecord->isPOD = crd->isPOD();
+
       //--- CppInheritance ---//
 
       for (auto it = crd->bases_begin(); it != crd->bases_end(); ++it)
@@ -513,6 +510,28 @@ public:
           friendship->target = cppRecord->entityHash;
           friendship->theFriend = util::fnvHash(getUSR(friendDecl));
         }
+      }
+
+      // --- Destructor --- //
+
+      void* ref = crd;
+      model::FileLoc location = getFileLoc(crd->getEndLoc(), crd->getEndLoc());
+      if (clang::CXXDestructorDecl* dd = crd->getDestructor())
+      {
+        ref = dd;
+        if (crd->hasUserDeclaredDestructor())
+        {
+          clang::Stmt* body = dd->getBody();
+          location = body
+            ? getFileLoc(body->getEndLoc(), body->getEndLoc())
+            : getFileLoc(dd->getEndLoc(), dd->getEndLoc());
+        }
+      }
+
+      for (auto it = crd->field_begin(); it != crd->field_end(); ++it)
+      {
+        clang::FieldDecl* fd = *it;
+        addDestructorUsage(fd->getType(), location, ref);
       }
     }
 
@@ -921,6 +940,32 @@ public:
     return true;
   }
 
+  bool VisitStmt(clang::Stmt* s_)
+  {
+    if (clang::DeclStmt* ds = llvm::dyn_cast<clang::DeclStmt>(s_))
+    {
+      // FIXME: Using the parent map is expensive.
+      // My original idea was to simply use the second top-most statement from
+      // _contextStatementStack as scopeSt, but the stack is always empty
+      // for some reason, except in VisitDeclRefExpr... (why?)
+      auto parents = _astContext.getParentMapContext()
+        .getParents<clang::Stmt>(*s_);
+      const clang::Stmt* scopeSt = parents[0].get<clang::Stmt>();
+      
+      for (auto it = ds->decl_begin(); it != ds->decl_end(); ++it)
+      {
+        if (clang::VarDecl* vd = llvm::dyn_cast<clang::VarDecl>(*it))
+        {
+          model::FileLoc loc =
+            getFileLoc(scopeSt->getEndLoc(), scopeSt->getEndLoc());
+          addDestructorUsage(vd->getType(), loc, vd);
+        }
+      }
+    }
+
+    return true;
+  }
+
   bool VisitNamespaceDecl(clang::NamespaceDecl* nd_)
   {
     //--- CppAstNode ---//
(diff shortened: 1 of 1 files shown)


mcserep:
> Only one question so far at line 947: is there a better solution than this?

@dbukki I am not sure which line you mean, line 947 seems to be unchanged in the GitHub diff viewer. Please mark the respective line with a comment in the PR.

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -921,6 +940,32 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
     return true;
   }
 
+  bool VisitStmt(clang::Stmt* s_)
+  {
+    if (clang::DeclStmt* ds = llvm::dyn_cast<clang::DeclStmt>(s_))
+    {
+      // FIXME: Using the parent map is expensive.
```
dbukki: getParentMapContext is said to be expesnive, and probably not the best approach. We should instead use some sort of statement cache to retrieve the parent of the current statement during traversal.

dbukki:
> @dbukki I am not sure which line you mean, line 947 seems to be unchanged in the GitHub diff viewer. Please mark the respective line with a comment in the PR.

In my first commit, in VisitStmt there was a FIXME comment that we've also discussed during our last weekly meeting. It was about whether there is a better solution to retrieving the parent statement of a statement than using the parent map.

In my second commit on this PR, I solved this by using a similar stack-based approach that we've already had examples for in the code.

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -435,15 +443,12 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
     cppRecord->entityHash = astNode->entityHash;
     cppRecord->name = rd_->getNameAsString();
     cppRecord->qualifiedName = rd_->getQualifiedNameAsString();
-    if (const clang::CXXRecordDecl* crd
-        = llvm::dyn_cast<clang::CXXRecordDecl>(rd_))
+
+    if (clang::CXXRecordDecl* crd = llvm::dyn_cast<clang::CXXRecordDecl>(rd_))
```
whisperity, line 447: Why are we losing the `const` here?

whisperity, line 447: ✏️ (Style nit: `auto*` would be enough here, the type is spelt in the initialiser.)

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -514,6 +519,28 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
           friendship->theFriend = util::fnvHash(getUSR(friendDecl));
         }
       }
+
+      // --- Destructor --- //
+
+      void* ref = crd;
```
whisperity, line 525: why are you doing this? Is this variable used anywhere else later? It's really shocking to so violently get rid of type information...!

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -921,6 +948,27 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
     return true;
   }
 
+  bool VisitDeclStmt(clang::DeclStmt* ds_)
+  {
+    assert(_statements.top() == ds_ &&
+      "ds_ was expected to be the top-level statement.");
```
whisperity, line 954: I'm not sure whether this assertion holds everywhere. Unfortunately, although defining a function within another function is not possible, except for lambdae:

```cpp
int f()
{
    auto L = []() -> int {
        int X;
        return 1;
    }();
    return L;
}
```

here, `VarDecl`s `L` and `X` are both children of `DeclStmt`s, but `X` isn't top-level to the `f` function. However, it is top-level to the `operator()` of the lambda, unless having the `CompoundStmt` there is a problem...

```cpp
int f()
{
  int X = 1;
  {
    int X = 2;
    return X;
  }
}
```

plugins/cpp/parser/src/clangastvisitor.h
dbukki: The variable ref has type void* because the clangPtr_ parameter of addDestructorUsage that we pass it to also expects void*. And that is so because it is then further passed on to insertToCache, which also expects void*. So in essence, I just matched the type of parameters to an already existing pattern of expecting void* in the end at insertToCache. Whether or not this is a good pattern to begin with is of course up for debate, but I did not want to change the signature of insertToCache just for this, without fully knowing why we chose void* there in the first place.

mcserep:
As we discussed on the weekly meeting, even in that case this type
conversion should be done only when passing the argument to the
`insertToCache` function.

plugins/cpp/parser/src/clangastvisitor.h
mcserep: As we discussed on the weekly meeting, even in that case this type conversion should be done only when passing the argument to the `insertToCache` function. 
Don't loose tye information everywhere, just because one function won't need it!

Explicit conversion won't even be required, see other usages of the `insertToCache` function.

plugins/cpp/parser/src/clangastvisitor.h
dbukki: We only really 'lost' the const qualifier for the first two lines (when assigning isPOD and isAbstract). As for the code below //--- CppInheritance ---//: crd was non-const there even before this commit, so no loss there.
What I essentially did here is that I noticed that two consecutive if statements had the exact same dyn_cast condition, so I put them inside one common if statement instead to reduce code duplication and the same dyn_cast overhead twice, and kept the condition without the const qualifier.
We can of course keep the const as well, I just did not want to 'restrict' the code to having to use const-only functionality on an object that we otherwise received as non-const in the function argument. In this particular case, the code seems to compile with crd being const as well, so I'll consider marking it const.

plugins/cpp/parser/src/clangastvisitor.h
dbukki: OK, I have replaced the void* type of ref with the most derived common ancestor of the two types of AST nodes we put into it (see 2d56647ffe4d950ebc1b2a25175b37b78843eb22).

plugins/cpp/parser/src/clangastvisitor.h
dbukki: screen

I have tested your code with the minor change of replacing int with a UDT that has an explicit destructor so we can actually see its usages in the Info Tree. It gets parsed perfectly without the assertion ever breaking.

Note: The 5th usage in the above screenshot at 81:5 is not currently visible in master because of #583. I used the fix in #679 locally to make it visible during testing, but that is a separate unrelated issue.
The point is that no assertions were hit.

mcserep: @whisperity Can you take a look on this?

whisperity: > @whisperity Can you take a look on this?

LGTM. 😉 It's better to have these looked at not to run into odd crashes later.

whisperity (approved):
Apologies for being late with this and that @mcserep had to ping me extensively. I get like 1000 mails from GitHub every day thanks to [llvm/llvm-project](https://github.com/llvm/llvm-project).

I think this is good to go with the changes applied.

Issue #631, opened 2023-09-20
Extend CI for Ubuntu 22.04


mcserep:
The GitHub Actions pipeline should be extended to build and test CodeCompass on Ubuntu 22.04, beside the current Ubuntu 20.04 pipeline.

PR #657, opened 2023-10-27
Update Thrift to version 0.16

### LINKED ISSUE #630: Update Thrift to 0.16

LoremIPsummer:
Update Thrift to version 0.16.

Since version 0.14, Thrift does not generate certain files, causing an error in CMake. To solve this, I made a wrapper function around the generation logic, so if some of the specified files are not generated, we manually create them. 

There was an undocumented breaking change involving the java transport protocol implementations. The indexer service didn't handle all the exceptions correctly, leading to an API mismatch. 

Since version 0.15, the javascript library provided by the `thrift` npm package also ships with the nodejs bindings. For environment detection, it internally uses a package called `browser-or-node`. However, this plugin needs to be bundled with the generated javascript code, making it impossible to directly use it in the old webgui by simply referencing the thrift.js library downloaded from npm. Such change requires the introduction of a build system (like `webpack` or `browserify`), and the proper conversion of CommonJS modules to AMD, so it goes beyond the scope of this PR. However, a traditional browser version of the library can still be acquired by [downloading it directly from the Thrift repository](https://github.com/apache/thrift/blob/master/lib/js/src/thrift.js) with the proper version. I've decided to simply grab it from here and put it into the appropriate `scripts` directory.

The new webgui doesn't produce this behavior, as expected.

fixes #630


mcserep (changes requested):

Functions.cmake
```diff
@@ -125,3 +125,27 @@ function(join _values _glue _output)
   string (REGEX REPLACE "[\\](.)" "\\1" _tmpStr "${_tmpStr}") #fixes escaping
   set (${_output} "${_tmpStr}" PARENT_SCOPE)
 endfunction(join)
+  
+
+# Wrapper function to generate source files from a thrift file with the given arguments.
+# Since Thrift 0.14, some of the files may not be generated, if they would be empty.
+# To overcome this, we manually create the source files, in case the thrift compiler does not generate them.
+function(generate_from_thrift _output_files _thrift_file _additional_args)
```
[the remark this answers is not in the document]
LoremIPsummer: I reverted this functionality based on your suggestion. 
This approach (removing instead of creating) indeed seems to be cleaner. 

The original idea was to "standardize"  the list of the generated files, so the developer doesn't have to worry about which file is generated or not, but such a safeguard shouldn't be necessary.

doc/deps.md
```diff
@@ -180,10 +180,10 @@ avoid using them if it's not necessary.
...
 ```bash
 # Download and uncompress Thrift:
-wget "http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.13.0/thrift-0.13.0.tar.gz" \
-  -O thrift-0.13.0.tar.gz
-tar -xvf ./thrift-0.13.0.tar.gz
-cd thrift-0.13.0
+wget "http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.16.0/thrift-0.16.0.tar.gz" \
+  -O thrift-0.16.0.tar.gz
+tar -xvf ./thrift-0.16.0.tar.gz
+cd thrift-0.16.0
```
mcserep, line 186: The documentation needs further update, because now for Ubuntu 22.04, Thrift can be installed from the official package repositories, while for Ubuntu 20.04, manual compilation is required.

mcserep (changes requested):
The GitHub CI should also be updated to use Thrift 16.

Functions.cmake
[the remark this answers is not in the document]
mcserep: I understand the benefit of your initial approach, but if Thrift decided to drop generating these practically empty files, we should not stick to the old behaviour.

doc/deps.md
mcserep: Sorry I omitted these in the first review round.

LoremIPsummer: I did a somewhat larger refactoring here, that will even help with #631 .

I have reorganized the script folder structure in a way that now it supports compilation with catching abilities. Furthermore, I updated the CI accordingly. The documentation is now also changed.

.github/scripts/ubuntu-20.04/compile_build.sh
```diff
@@ -0,0 +1,22 @@
...
+  --without-haskell --without-go --without-rs --without-haxe        \
+  --without-dotnetcore --without-d --without-qt4 --without-qt5      \
+  --without-java --without-swift
+
+make install -j $(nproc)
+
+
+
+
+
```
mcserep, line 22: Minor remark: few empty lines have been left here.

.github/scripts/ubuntu-20.04/download_build.sh
```diff
@@ -0,0 +1,22 @@
...
+
+wget -O ${DOWNLOAD_PATH}/thrift-0.16.0.tar.gz "http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.16.0/thrift-0.16.0.tar.gz"
+
+# Calculate hash of dependencies for Github Cache Action
+
+dependencies_to_hash=("thrift-0.16.0.tar.gz")
+
+concatenated_data=""
+for file in "${dependencies_to_hash[@]}"; do
+    concatenated_data="${concatenated_data}$(echo -n "$file"; cat "${DOWNLOAD_PATH}/${file}")"
```
mcserep, line 15: Concatenating the complete binary content of all dependencies maybe shouldn't be done, instead just concatenate hashes. (And probably do a second hashing after that, but that could be optional.)

.github/scripts/ubuntu-20.04/download_build.sh
```diff
@@ -0,0 +1,22 @@
+# Download installers for compiled dependencies
+
+mkdir -p ${DOWNLOAD_PATH}
+
+## Thrift 0.16
+
+wget -O ${DOWNLOAD_PATH}/thrift-0.16.0.tar.gz "http://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=thrift/0.16.0/thrift-0.16.0.tar.gz"
```
mcserep, line 7: In the Gitlab CI, we use the `http://archive.apache.org/dist/thrift/0.16.0/thrift-0.16.0.tar.gz` URL.

.github/scripts/ubuntu-20.04/compile_build.sh
```diff
@@ -0,0 +1,22 @@
+# Compile dependencies that are cannot be acquired via the official repositories
```
mcserep, line 1:
```suggestion
# Compile dependencies that cannot be acquired via the official repositories
```

.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on: [push, pull_request, workflow_dispatch]
 
 env:
   BUILD_TYPE: Debug
+  ## For locally compiled dependencies
+  INSTALL_PATH: ${{github.workspace}}/installed
+  ## Temp directory for installers of the downloaded dependencies
+  DOWNLOAD_PATH: ${{github.workspace}}/downloads
```
mcserep, line 10:
```suggestion
  ## For locally compiled dependencies
  INSTALL_PATH: ${{github.workspace}}/dependencies/install
  ## Temp directory for installers of the downloaded dependencies
  DOWNLOAD_PATH: ${{github.workspace}}/dependencies/download
```

mcserep (changes requested):

doc/deps.md
mcserep: Nice caching solution. I have added some remarks, most of them are minor.

.github/scripts/ubuntu-20.04/compile_build.sh
LoremIPsummer: Late night coding shenanigans. Fixed.

.github/scripts/ubuntu-20.04/download_build.sh
LoremIPsummer: Concatenating only the hashes indeed should be enough. Fixed.

.github/scripts/ubuntu-20.04/postcompile_runtime.sh
```diff
@@ -0,0 +1,3 @@
+# Post compilation configuration for runtime (environmental variables, library location settings etc..)
+
+sudo cp ${INSTALL_PATH}/thrift/lib/libthrift.so /usr/lib
```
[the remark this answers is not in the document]
LoremIPsummer: Here, I simply added the compiled thrift lib path to CMAKE_PREFIX_PATH.

.github/scripts/ubuntu-20.04/download_build.sh
LoremIPsummer: Fixed.

.github/scripts/ubuntu-20.04/compile_build.sh
LoremIPsummer: Fixed.

.github/workflows/ci.yml
LoremIPsummer: Fixed.

.github/scripts/ubuntu-20.04/postcompile_build.sh
```diff
@@ -0,0 +1,7 @@
+# Post compilation configuration for building (environmental variables, library location settings etc..)
+
+echo "${INSTALL_PATH}/thrift/bin" >> $GITHUB_PATH
+echo "CMAKE_PREFIX_PATH=${INSTALL_PATH}/thrift:$CMAKE_PREFIX_PATH" >> $GITHUB_ENV
+
+
+sudo cp ${INSTALL_PATH}/thrift/lib/libthrift.so /usr/lib
```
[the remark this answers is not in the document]
LoremIPsummer: For the compilation part, it wasn't even necessary to copy that library, since CMAKE_PREFIX_PATH already had that covered. Fixed

.github/scripts/ubuntu-20.04/postcompile_runtime.sh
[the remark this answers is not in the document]
LoremIPsummer: Fixed.

LoremIPsummer:
@mcserep 
I also removed the not generated files from the new `cppmetrics` plugin from CMake.

mcserep (approved):
Looks good now, nice work!

PR #679, opened 2023-11-11
Nodes inside lambdas are no longer considered inherently implicit

### LINKED ISSUE #583: Clicking any node inside a lambda does not work

dbukki:
Lambda closure types have always been properly visited and parsed into the database with one notable exception: their source location was missing.
The _isImplicit variable prevented nodes rooted under any implicit node (anywhere in the parent chain) from being assigned a source location via _fileLocUtil, so even though the nodes were there, the UI could not locate them.

This fixes #583.

This is more or less a revert (+ some additional necessary refactor) of a much older commit mentioned in the issue (3bc3c7f1f69c4a1a0680dc289b50d835d7cf74a1 by @bruntib). I could not find the issue that this commit was originally supposed to fix, so any insights are welcome. But the assumption that implicit nodes will only have implicit sub-nodes clearly turns out to be incorrect (as shown by lambdas).

### DIFF
--- a/plugins/cpp/parser/src/clangastvisitor.h
+++ b/plugins/cpp/parser/src/clangastvisitor.h
@@ -73,7 +73,8 @@ public:
     clang::ASTContext& astContext_,
     EntityCache& entityCache_,
     std::unordered_map<const void*, model::CppAstNodeId>& clangToAstNodeId_)
-    : _ctx(ctx_),
+    : _isImplicit(false),
+      _ctx(ctx_),
       _clangSrcMgr(astContext_.getSourceManager()),
       _fileLocUtil(astContext_.getSourceManager()),
       _astContext(astContext_),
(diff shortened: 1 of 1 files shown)


mcserep:
@bruntib This the issue I mentioned on Friday. Please take a look when you have time.

mcserep:
@dbukki Reverting that change would'nt mean that e.g. the body of an implicit copy constructor won't be marked as implicit anymore?

dbukki:
> @dbukki Reverting that change wouldn't mean that e.g. the body of an implicit copy constructor won't be marked as implicit anymore?

I am tempted to say that the body of an implicitly generated copy/move constructor/operator should also remain implicit, because none of it is explicitly typed in the code. I can verify that at least the declarations are still implicit using the old UI:
image
But of course what is inside the body is a more interesting question. For example, parsing
```
struct point2
{
	int x;
	int y;
	
	point2(int x, int y) : x(x), y(y) {}
	point2() : point2(0, 0) {}
	
	virtual ~point2() = default;
};
```
yields the follwong AST:
image
where the copy operator and constructor are implicit, but their parameters do not seem to be (or maybe this is so trivial that it's just not included in the string representation?).

By the way: Is this what the original commit eluded to? And is this actually a problem? / In what way can this be a problem?
Because if _isImplicit was introduced solely to resolve the implicitness of the parameters of implicit methods, then perhaps a more targeted solution could be achieved without influencing the inherent implicitness of everything else too.

---

With lambdas, the complexity of the situation (and the reason why _isImplicit is too aggressive) is that:
The closure object that the compiler generates around the lambda body is implicit. It does also have implicit members and constructors and such. But its operator() is explicit, because that part is written in the code.
So we have an explicit AST node with an implicit parent, and using _isImplicit discards these cases.

We can observe that when parsing something like
```
[]() -> IntWrap {
    IntWrap X = 0;
    return X;
}
```
the AST is
image
where the CXXRecordDecl at the top representing the lambda is implicit, but its CXXMethodDecl for operator() is not.

mcserep (approved):
Thanks for finishing this @dbukki. For future reference, the conclusion to resolve this issue was:

> We use implicitness to determine if actual symbol location information should be stored for AST nodes in our database. This differs somewhat from Clang's concept of implicitness. To bridge the gap between the two interpretations, we mostly assume implicitness to be hereditary from parent to child nodes, except in some known special cases (lambdas).

PR #686, opened 2023-11-18
Extend CI for Ubuntu 22.04

### LINKED ISSUE #631: Extend CI for Ubuntu 22.04

LoremIPsummer:
Adds support for ubuntu 22.04 runs in the CI with caching abilities.
Also contains some rework for the ubuntu 20.04 scripts. 

Unfortunately, since matrix actions doesn't support dynamic variables (outputs), the cache keys had to be hard-coded for each OS. In my opinion, considering how many OSes we target, it shouldn't lead to scaling issues.

During the `parse` job, the current `matrix.os` with the combination of one of the output keys ensures that the correct dependencies will be used.  

fixes #631

### DIFF
--- a/.github/scripts/ubuntu-22.04/setup_runtime.sh
+++ b/.github/scripts/ubuntu-22.04/setup_runtime.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-# Install required packages for CodeCompass runtime
-sudo apt-get install -y git cmake make g++ graphviz \
-  libboost-filesystem1.71.0 libboost-log1.71.0 libboost-program-options1.71.0 \
-  libllvm11 clang-11 libclang1-11 libthrift-0.16.0 default-jre libssl1.1 libmagic1 \
-  libgit2-28 ctags googletest libldap-2.4-2
--- a/.github/scripts/ubuntu-22.04/setup_runtime.sh
+++ b/.github/scripts/ubuntu-22.04/setup_runtime.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Install required packages for CodeCompass runtime
+sudo apt-get install -y git cmake make g++ graphviz \
+  libboost-filesystem1.71.0 libboost-log1.71.0 libboost-program-options1.71.0 \
+  libllvm11 clang-11 libclang1-11 libthrift-0.16.0 default-jre libssl1.1 libmagic1 \
+  libgit2-28 ctags googletest libldap-2.4-2
--- a/scripts/install_latest_build2.sh
+++ b/scripts/install_latest_build2.sh
@@ -9,10 +9,23 @@ fi
 
 install_dir=$1
 toolchain_file="toolchain.sha256"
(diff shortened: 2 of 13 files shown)


.github/scripts/ubuntu-22.04/setup_runtime.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Install required packages for CodeCompass runtime
+sudo apt-get install -y git cmake make g++ graphviz \
+  libboost-filesystem1.71.0 libboost-log1.71.0 libboost-program-options1.71.0 \
+  libllvm11 clang-11 libclang1-11 libthrift-0.16.0 default-jre libssl1.1 libmagic1 \
+  libgit2-28 ctags googletest libldap-2.4-2
```
mcserep, line 7: Most of these package versions does not exist on Ubuntu 22.04, they should be updated.

CI only passes, because for some reason the `parse` job calls the `setup_build.sh`, instead of the `setup_runtime.sh`. Maybe because we need to build CodeCompass in the parse job, as that is one of the tests.

In that case if `setup_runtime.sh` is not needed, it could be deleted?

scripts/install_latest_build2.sh
```diff
@@ -9,10 +9,23 @@ fi
 
 install_dir=$1
 toolchain_file="toolchain.sha256"
-wget --no-verbose --no-clobber https://download.build2.org/toolchain.sha256 -O "${toolchain_file}"
+if [ "$1" = "--version" ]; then
+    wget -q --no-verbose --no-clobber https://download.build2.org/toolchain.sha256 -O "${toolchain_file}"
+else
+    wget --no-verbose --no-clobber https://download.build2.org/toolchain.sha256 -O "${toolchain_file}"
+fi
```
mcserep, line 16: What is the diff here? I only see a `-q` flag.

mcserep (changes requested):

mcserep:
Any update on this @LoremIPsummer?

.github/scripts/ubuntu-22.04/setup_runtime.sh
LoremIPsummer: I've totally overlooked on this. It is indeed an unnecessary script in this way. Removed.

scripts/install_latest_build2.sh
LoremIPsummer: The full change is to give the script the ability to print which version would it install in case someone called it with the `--version` argument, without actually installing it. I think it is a better approach to exclude the output of `wget` and only print out the version number in this case. Since the `install_latest_build2.sh` script won't change even a new build2 version is out, I needed a mechanism to cache build2 in a cheap manner, so I use the version output to keep track of it.

mcserep (approved):
Looks good now, LGTM! :rocket:

Issue #653, opened 2023-10-18
Cannot run PostgreSQL server directly from docker containers


mozesl:
The postgresql-12(/14) package is not installed in any of the docker images, therefore it's not possible to run CodeCompass with Postgres directly from those containers and requires additional installs.

PR #687, opened 2023-12-02
Fixed: Cannot run PostgreSQL server directly from docker containers

Chillthrower:


mcserep:
Hi @Chillthrower,  
Thanks for the fix, I think this will resolve #653.

docker/dev/Dockerfile
```diff
@@ -7,6 +7,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 
 # CodeCompass development dependencies.
 RUN set -x && apt-get update -qq \
+  && apt-get install -y postgresql-14 \
   && apt-get -y install --no-install-recommends \
```
mcserep, line 11: Please extend the existing `apt-get install` statement instead. You can add this new package to the end of the list.

mcserep (changes requested):

docker/dev/Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN set -x && apt-get update -qq \
   thrift-compiler libthrift-dev \
   postgresql-server-dev-14 && \
   ln -s /usr/bin/gcc-11 /usr/bin/gcc && \
-  ln -s /usr/bin/g++-11 /usr/bin/g++
+  ln -s /usr/bin/g++-11 /usr/bin/g++ && \
+  apt-get install -y postgresql-14
```
mcserep, line 34: No, I meant that the installation of the `postgresql-14` package should go to the end of the package list for the existing `apt-get install` command, right after ` postgresql-server-dev-14`.

mcserep (changes requested):

mcserep (approved):
Thanks for updating the MR accordingly!

PR #688, opened 2023-12-03
Lack of cohesion metrics

### LINKED ISSUE #681: Lack of Cohesion of Methods at Type Level for C++
The  single responsibility principle states that a class should have only one reason to change and one responsibility. Such a class is said to be cohesive.  
A class has strong *cohesion* if its individual operations use as many of the class's fields as possible. Otherwise, the class is likely to perform multiple roles.

Several metrics exist for calculating a *Lack of Cohesion of Methods* value (hereinafter *LCOM*).
The algorithms used by [CppDepend](https://www.cppdepend.com/documentation/code-metrics#LCOM) to compute LCOM metrics and LCOM HS (Henderson-Sellers variant of LCOM) are the following:  
`LCOM = 1 – (sum(MF)/(M*F))` 
`LCOM HS = (M – sum(MF)/F)/(M-1)`  

Where:
- `M` is the number of methods in class (both static and instance methods are counted);
 - `F` is the number of instance fields in the class;
 - `MF` is the number of methods of the class accessing a particular instance field;
 - `sum(MF)` is the sum of `MF` over all instance fields of the class.

The LCOM takes its values in the range [0-1]. The LCOM HS takes its values in the range [0-2].   
A class is utterly cohesive if all its methods use all its instance fields, which means that `sum(MF)=M*F` and then `LCOM = 0` and `LCOMHS = 0`.

**Recommendations:**  
Consider only types with more than 10 methods and 10 fields.
In such cases an LCOM value larger than 0.8 or an LCOMHS value larger than 1.0 might be problematic.
The constraint related to LCOMHS is stronger (and thus easier to satisfy) than the constraint related to LCOM.

dbukki:
Introduction of Lack of Cohesion (LoC) metrics in the metrics plugin.

* New metrics calculation function: CppMetricsParser::lackOfCohesion
* New metrics types: LACK_OF_COHESION and LACK_OF_COHESION_HS
* Helper function for detecting if the given file path of an AST node is internal to the parsed project: CppMetricsParser::isInInputPath
* New helper database views in model/cppcohesionmetrics.h
* Verbose debug utilities for the real-time monitoring of the parsed types.
* Refactor: Removed a duplicate code block from ClangASTVisitor::getFileLoc.

Fixes #681.


plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -107,11 +123,158 @@ void CppMetricsParser::functionParameters()
...
+    
+    typedef odb::query<model::CohesionCppAstNodeView>::query_columns QNode;
+    const auto& QNodeFilePath = QNode::File::path;
+    const auto& QNodeRange = QNode::CppAstNode::location.range;
+    
+    #ifdef DEBUG_COHESION_VERBOSE
+    std::size_t typecount =
+      _ctx.db->query_value<model::CppRecordCount>().count;
+    std::size_t typeindex = 0;
+    std::size_t checkedcount = 0;
```
intjftw, line 147: `std::size_t` is usually used as a size of something. If it's not a container that has an actual size, it is more natural to use `unsigned` instead of `size_t`.
Also, please follow the coding conventions even if these are log variables (`typeCount` etc.)

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
```diff
@@ -21,14 +24,26 @@ namespace parser
 class CppMetricsParser : public AbstractParser
 {
 public:
+  static constexpr double LOC_SCALING = 1e+4;
```
intjftw, line 27: I'm guessing that this magic number is a multiplier to the actual metrics?

intjftw (changes requested):
Ultimately this PR is fairly good. I tested it on TinyXML2, and it returns believable values. Additionally, @mcserep told me that you agreed on adding automated test cases which I don't see so please @dbukki add them along with my change requests.

intjftw:
Something I noticed running the plugin in debug mode: it seems that one AST node is calculated multiple times:

```txt
2023-12-11 11:41:59 [DEBUG] 498/527	tinyxml2::MemPoolT::Item               0       0
2023-12-11 11:41:59 [DEBUG] 499/527	tinyxml2::MemPoolT::Block              0       0
2023-12-11 11:41:59 [DEBUG] 500/527	tinyxml2::MemPoolT                 0.625  0.6818
2023-12-11 11:41:59 [DEBUG] 501/527	tinyxml2::MemPoolT                0.7083  0.7727
2023-12-11 11:41:59 [DEBUG] 502/527	tinyxml2::MemPoolT                0.7083  0.7727
2023-12-11 11:41:59 [DEBUG] 503/527	tinyxml2::MemPoolT                0.7083  0.7727
2023-12-11 11:41:59 [DEBUG] 504/527	tinyxml2::MemPoolT                0.7083  0.7727
```
@dbukki is this an error or just something with the debug log?

plugins/cpp/parser/src/clangastvisitor.h
```diff
@@ -1339,34 +1339,27 @@ class ClangASTVisitor : public clang::RecursiveASTVisitor<ClangASTVisitor>
...
-      if (type != model::File::DIRECTORY_TYPE && type != _cppSourceType)
-      {
-        fileLoc.file->type = _cppSourceType;
-        _ctx.srcMgr.updateFile(*fileLoc.file);
-      }
-      return fileLoc;
-    }
-
-    clang::SourceLocation realStart = start_;
-    clang::SourceLocation realEnd = end_;
```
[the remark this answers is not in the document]
dbukki: While trying to figure out a way to calculate metrics for only "internal types", I had a brief excursion into this function to see how type locations (specifically: the file paths) are stored.
Here in the old code I noticed that lines 1344-1350 and 1371-1376 were duplicates, making the function harder to interpret than it actually is. So I took the redundancy out.
Anything that was previously not executed because of the early return statement at line 1350 is now in the 'else' block of that conditional statement. What used to be done by lines 1344-1350 is now simply performed by 1371-1376 for both branches of execution. This should result in the same behavior as before, but with less clutter.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -107,11 +123,158 @@ void CppMetricsParser::functionParameters()
...
+  });
+}
+
+bool CppMetricsParser::isInInputPath(const std::string& path_) const
+{
+  std::size_t i = 0;
+  while (i < _inputPaths.size() && path_.rfind(_inputPaths[i], 0) != 0)
+    ++i;
+  return i < _inputPaths.size();
+}
```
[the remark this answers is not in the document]
dbukki: The behavior of this function currently only depends on _inputPaths, which comes from the command line "-i" argument of the metrics plugin.
I can move a generalized version of this function into filesystem.h though - one that also takes the inputs vector as parameter - if that's OK.

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
dbukki: Normally, the lack of cohesion metric is a real value in the range [0,1] (or [0,2] for the HS variant), which can only be stored in our current `unsigned int` value field if we premultiply it by some constant. This is that constant, and I agree that it's fairly arbitrary. I actually raised this concern in the original issue, but for now we agreed to go with this approach (https://github.com/Ericsson/CodeCompass/issues/681#issuecomment-1831904646).

dbukki:
> Something I noticed running the plugin in debug mode: it seems that one AST node is calculated multiple times:
> ...
> @dbukki is this an error or just something with the debug log?

I am currently investigating this strange phenomenon too.
If the same type only ever appeared in the output with the same metrics, I would be very much tempted to say that this is just a printing error. But for some reason, the same type can sometimes appear with different metrics, as is the case with `tinyxml2::MemPoolT` in your example.

My initial guess is that these are actually different template instantiations of the same type, and some of them are specializations that have a different number of fields and methods.
If this is indeed the case, then the fault is to be found in our already existing type extraction logic in the parser: perhaps the fully qualified type name does not yet contain the template arguments?

It could also be that types defined in a header and included by multiple compilation units each have their own copy of this type in the database.

And then it's also possible that both of the above factors play a role in this output. I have yet to figure out a full diagnosis.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
dbukki: Thanks for noticing the variable casing, that was just hasty coding on my part in places that were initially only supposed to be temporary debug-only features. But of course if we intend to keep them under `--loglevel=debug,` they will be refactored.

About `std::size_t`: I went with it because the already existing `model::CppRecordCount::count` field that I used to get the total number of types was already defined as `std::size_t`. From that, to me it followed logically that the type of a "range" and its "indexer" should match (the same way we use `std::size_t` for both the size and the indexing of a `std::vector`).

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
mcserep: @dbukki As discussed on the last meeting in December 2023, make this field a float or a double. It should also be *nullable* to handle extremal values. See `#pragma db null` for ODB.

dbukki: @mcserep The value is now a double. See https://github.com/Ericsson/CodeCompass/pull/688/commits/8d757aec85a6e8a20ca17b07436127f448a9eac7

plugins/cpp_metrics/test/sources/service/mccabe.cpp
```diff
@@ -0,0 +1,148 @@
+void conditionals(int arg1, bool arg2) { // +1
```
intjftw, line 1: I'm not sure why this file and many other code related to PR #689 appear here. @mcserep do you think this will mess up the commit history?

plugins/cpp_metrics/test/sources/parser/lackofcohesion.cpp
```diff
@@ -0,0 +1,223 @@
+////////////////
```
intjftw, line 1: The tests fail in the CI, the PR cannot be merged this way.

intjftw (changes requested):
The most critical point is the test failure in the CI. If that's corrected, I say we merge the PR, and clean up a few things afterwards. @mcserep what do you think?

docker/dev/Dockerfile
```diff
@@ -28,7 +28,8 @@ RUN set -x && apt-get update -qq \
   libssl-dev \
   llvm-11 clang-11 llvm-11-dev libclang-11-dev \
   thrift-compiler libthrift-dev \
-  postgresql-server-dev-14 && \
+  postgresql-server-dev-14 \
+  postgresql-14 && \
```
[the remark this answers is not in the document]
dbukki: I do not know. These commits and their diffs above were generated automatically when I rebased this branch onto the newest master. I needed to rebase because the foundation for the metrics unit tests was in a newer commit that I did not have in my old master. I do not know about the exact contents of all the other commits that are now also merged as a result.

plugins/cpp_metrics/test/sources/parser/lackofcohesion.cpp
dbukki: I cannot figure out why the tests fail though. It's not about the results, it's about the existence of the results at all. It seems as if the types I am looking for are simply not present in the database when the tests run.
I cannot use the test case for McCabe as a reference because that test actually only uses results parsed by the main cpp plugin. In my tests, I require records from the database parsed by the metrics plugin itself.
Is it possible that the tests are not properly configured and the test files that I have included are not even parsed by the metrics parser, only the cpp parser? Is there a way I can check this?

plugins/cpp_metrics/test/sources/service/mccabe.cpp
mcserep: @dbukki mentioned that he rebased the branch. I think something went the wrong way, since code already accepted and merged into `master` shows up in this PR as new changes. This also makes the review not easy.

@dbukki please clean this up and either force push the branch with a clean version history (containing only the changes related to your PR), or create a new PR accordingly.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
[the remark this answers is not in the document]
mcserep: Looks good to me.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -15,13 +17,27 @@
 
 #include <memory>
 
+// Controls whether cohesion metrics are printed to the output
+// immediately as they are being calculated.
+#define DEBUG_COHESION_VERBOSE
```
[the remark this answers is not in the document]
mcserep: I agree with @intjftw , we already use Boost's log library, no need for these macros. If you wouldn't like to include all these messages to the `debug` log, there is even a `trace` log level below it, which can be used.

plugins/cpp_metrics/test/sources/parser/lackofcohesion.cpp
mcserep: @dbukki In that case you should verify the content of the database and see why is there a mismatch?

mcserep: In the tests only the `compile_commands.json` was added to the input path, the source folder itself not. Fixed in [4f43dd7](https://github.com/Ericsson/CodeCompass/pull/688/commits/4f43dd7ee550b552e1e13bf910708082a5ae0f31).

plugins/cpp_metrics/test/sources/service/mccabe.cpp
mcserep: After merging the current `master` into this branch, unrelated changes to this PR was removed from the diff correctly. This issue is resolved.

docker/dev/Dockerfile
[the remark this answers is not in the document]
mcserep: After merging the current `master` into this branch, unrelated changes to this PR was removed from the diff correctly. This issue is resolved.

mcserep (approved):

PR #707, opened 2024-02-03
Maximize disk space in CI

mcserep:
Maximize disk space in **parse** GitHub Action job, as it tends to fail when parsing CodeCompass with PostgreSQL, due to low disk space.

Also remove unused steps from CI.


mcserep:
In the end I did not use the [`easimon/maximize-build-space`](https://github.com/marketplace/actions/maximize-build-disk-space), as it did not work for me. Instead, I simply remove the cached docker images on start.

intjftw (approved):

PR #710, opened 2024-02-05
C++ metrics Thrift API

intjftw:
I added basic Thrift API endpoints to the C++ metrics plugin. Currently there is only one endpoint that can query any type of metrics to a particular AST node. @mcserep what do you think what else I can add?

### DIFF
--- a/plugins/cpp_metrics/service/CMakeLists.txt
+++ b/plugins/cpp_metrics/service/CMakeLists.txt
@@ -1,6 +1,11 @@
 include_directories(
   include
   ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp
+  ${PROJECT_BINARY_DIR}/service/project/gen-cpp
+  ${PROJECT_SOURCE_DIR}/service/project/include
+  ${PROJECT_SOURCE_DIR}/model/include
+  ${PROJECT_BINARY_DIR}/service/language/gen-cpp
+  ${PLUGIN_DIR}/model/include
   ${PROJECT_SOURCE_DIR}/util/include
   ${PROJECT_SOURCE_DIR}/webserver/include)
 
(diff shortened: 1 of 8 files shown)


plugins/cpp_metrics/service/CMakeLists.txt
```diff
@@ -9,35 +14,45 @@ include_directories(SYSTEM
...
 
-add_library(cppmetricsservice SHARED
+add_dependencies(cxxmetricsthrift projectthrift)
+
+add_library(cxxmetricsservice SHARED
   src/cppmetricsservice.cpp
   src/plugin.cpp)
 
-target_compile_options(cppmetricsservice PUBLIC -Wno-unknown-pragmas)
+#target_compile_options(cxxmetricsservice PUBLIC -Wno-unknown-pragmas)
```
mcserep, line 44: Why was this commented out?

plugins/cpp_metrics/service/include/service/cppmetricsservice.h
```diff
@@ -1,11 +1,20 @@
...
 #include <boost/program_options/variables_map.hpp>
 
+#include <odb/database.hxx>
+
+#include <model/cppastnodemetrics.h>
+#include <model/cppastnodemetrics-odb.hxx>
+#include <model/cppastnode.h>
+#include <model/cppastnode-odb.hxx>
+
+//#include <projectservice/projectservice.h>
```
mcserep, line 16: Commented out code.

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -0,0 +1,20 @@
...
+{
+  ParameterCount = 1,
+  McCabe = 2,
+  LackOfCohesion = 3,
+  LackOfCohesionHS = 4
+}
+
+service CppMetricsService
+{
+  double getCppMetricsForAstNode(
```
mcserep, line 17: I am not sure whether it would be efficient or useful in the future to query metrics one by one for AST nodes. Maybe return all metrics for the queried node instead?

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -0,0 +1,20 @@
...
+
+enum CppMetricsType
+{
+  ParameterCount = 1,
+  McCabe = 2,
+  LackOfCohesion = 3,
+  LackOfCohesionHS = 4
+}
+
+service CppMetricsService
```
mcserep, line 15: A function returning possible metrics (id and name pairs) would also be needed.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -16,6 +16,35 @@ CppMetricsServiceHandler::CppMetricsServiceHandler(
...
+    typedef odb::result<model::CppAstNodeMetrics> CppAstNodeMetricsResult;
+    typedef odb::query<model::CppAstNodeMetrics> CppAstNodeMetricsQuery;
+
+    auto nodeMetric = _db->query<model::CppAstNodeMetrics>(
+      CppAstNodeMetricsQuery::astNodeId == std::stoull(astNodeId_) &&
+      CppAstNodeMetricsQuery::type ==
+        static_cast<model::CppAstNodeMetrics::Type>(type_));
+
+    if (nodeMetric.empty())
+      return -1.0;
```
mcserep, line 42: -1.0 could be a valid value for some metrics, do not return it as an extremal value.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: I added a function for this.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
intjftw: I changed it to `-DBL_MAX` as I read it on Stackoverflow. :)

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: I left this in the code in case it comes handy some time but I also added a function to collect all available metrics for an AST node.

plugins/cpp_metrics/service/CMakeLists.txt
intjftw: I was struggling with an error in CMake and this was a trial. I put it back.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: I still think no double value should be used here as an extremal value. Either `null` or e.g. a 404 status code should be returned.

(Also, in C++ code, the `std::numeric_limits<double>::max()` const function should be preferred over the C macro `DBL_MAX`.)

intjftw: Changed it to throwing an error, as we discussed today.

mcserep (approved):

Issue #673, opened 2023-11-08
Efferent Coupling at Type Level for C++


mcserep:
*Coupling* is metric which can be computed for structural units at different levels (e.g. classes, namespaces, modules).
It measures how many other entities an entity depends on; and how many dependants it has.

The *Efferent Coupling* for a particular type is the number of types it directly depends on. Types declared in framework projects are also taken into account.

CppDepend's recommendation is that types where efferent coupling value is larger than 50, are types that depends on too many other types. They are complex and have more than one responsibility. They are good candidate for refactoring.

mcserep:
Closed by #711.

PR #711, opened 2024-02-18
Efferent coupling of types

intjftw:
This PR solves #673. I'd like to include multiple related tasks, hence the WIP label.

- [x] Efferent coupling at type level.
- ~Refactoring of the C++ metrics plugin: organize metrics calculating functions into separate source files.~\
   Should be separate PR.
- ~Test cases for efferent coupling.~


plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -1,4 +1,5 @@
 #include <cppmetricsparser/cppmetricsparser.h>
+#include "efferent.h"
```
mcserep, line 2: Incorrect inclusion order. Please refer to the [coding conventions](https://github.com/Ericsson/CodeCompass/blob/master/doc/coding_conventions.md#headers).

plugins/cpp_metrics/parser/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ include_directories(
   ${PLUGIN_DIR}/model/include)
 
 add_library(cxxmetricsparser SHARED
+  src/efferent.cpp
```
mcserep, line 11: `efferent.h` should also be added here, so CMake can recognize that in case the header file changes, recompilation of the target is required.

mcserep (changes requested):
Currently this metric calculates the number of aggregated types for each type. Shouldn't inheritance or e.g. function parameter types be considered as well? (Question, should be discussed.)

intjftw:
I complemented the calculation with inherited types, member function parameter types and member function local variable types, aside from the already calculated member field types.
I also found and fixed a bug in the C++ metrics service: the first found metric was not returned for any AST node in `getCppAstNodeMetricsDetailedForPath`.
Tests are TBA.

PR #718, opened 2024-03-08
Endpoints for module-level C++ metric types

intjftw:
As we discussed, endpoints for module-level C++ metric types are also required. I added a new endpoint to query all available module-level metric types and all stored metrics for one module. There are no module-level metrics implemented for now, so testing is difficult.
@mcserep please help me out on what else we might need, and how should I modify the endpoint if needed.

### DIFF
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -94,7 +94,7 @@ service CppMetricsService
    * This function returns all available C++ metrics
    * (file-level) for a particular path.
    */
-  list<CppMetricsModuleAll> getCppFileMetricsForPath(
+  map<common.FileId, list<CppMetricsModuleSingle>> getCppFileMetricsForPath(
     1:string path)
 
   /**
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -4,12 +4,6 @@ include "project/project.thrift"
 namespace cpp cc.service.cppmetrics
 namespace java cc.service.cppmetrics
 
-enum CppUnitType
-{
-  AstNodeUnit = 1,
-  FileUnit = 2
-}
-
 enum CppAstNodeMetricsType
 {
   ParameterCount = 1,
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -37,8 +37,9 @@ struct CppModuleMetricsTypeName
 
 struct CppMetricsAstNodeSingle
 {
-  1:CppAstNodeMetricsType type,
-  2:double value
+  1:string path,
+  2:CppAstNodeMetricsType type,
+  3:double value
 }
 
 struct CppMetricsAstNodeAll
@@ -49,8 +50,9 @@ struct CppMetricsAstNodeAll
 
 struct CppMetricsModuleSingle
 {
-  1:CppModuleMetricsType type,
-  2:double value
+  1:string path,
+  2:CppModuleMetricsType type,
+  3:double value
 }
 
 struct CppMetricsModuleAll
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -4,7 +4,7 @@ include "project/project.thrift"
 namespace cpp cc.service.cppmetrics
 namespace java cc.service.cppmetrics
 
-enum CppMetricsType
+enum CppAstNodeMetricsType
 {
   ParameterCount = 1,
   McCabe = 2,
@@ -12,15 +12,32 @@ enum CppMetricsType
   LackOfCohesionHS = 4
 }
 
-struct CppMetricsTypeName
+enum CppModuleMetricsType
 {
-  1:CppMetricsType type,
+  Placeholder = 1
+}
+
+struct CppAstNodeMetricsTypeName
+{
+  1:CppAstNodeMetricsType type,
+  2:string name
+}
+
+struct CppModuleMetricsTypeName
+{
+  1:CppModuleMetricsType type,
   2:string name
 }
 
 struct CppMetricsAstNode
 {
-  1:CppMetricsType type,
+  1:CppAstNodeMetricsType type,
+  2:double value
+}
+
+struct CppMetricsModule
+{
+  1:CppModuleMetricsType type,
   2:double value
 }
 
@@ -32,7 +49,7 @@ service CppMetricsService
    */
   double getSingleCppMetricForAstNode(
     1:common.AstNodeId astNodeId,
-    2:CppMetricsType metric)
+    2:CppAstNodeMetricsType metric)
 
   /**
    * This function returns all available C++ metrics
@@ -42,7 +59,19 @@ service CppMetricsService
     1:common.AstNodeId astNodeId)
 
   /**
-   * This function returns the names of metrics.
+   * This function returns all available C++ metrics
+   * for a particular module.
    */
-  list<CppMetricsTypeName> getCppMetricsTypeNames()
+  list<CppMetricsModule> getCppMetricsForModule(
+    1:common.FileId fileId)
+
+  /**
+   * This function returns the names of AST node metrics.
+   */
+  list<CppAstNodeMetricsTypeName> getCppAstNodeMetricsTypeNames()
+
+  /**
+   * This function returns the names of module-level metrics.
+   */
+   list<CppModuleMetricsTypeName> getCppModuleMetricsTypeNames()
 }
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -4,6 +4,12 @@ include "project/project.thrift"
 namespace cpp cc.service.cppmetrics
 namespace java cc.service.cppmetrics
 
+enum CppUnitType
+{
+  AstNodeUnit = 1,
+  FileUnit = 2
+}
+
 enum CppAstNodeMetricsType
 {
   ParameterCount = 1,
@@ -35,10 +41,33 @@ struct CppMetricsAstNode
   2:double value
 }
 
+struct CppAllMetricsAstNode
+{
+  1:common.AstNodeId id,
+  2:list<CppMetricsAstNode> metrics
+}
+
 struct CppMetricsModule
 {
   1:CppModuleMetricsType type,
   2:double value
+}
+
+struct CppAllMetricsModule
+{
+  1:common.FileId id,
+  2:list<CppMetricsModule> metrics
+}
+
+// Thrift does not provide a union type,
+// the ids can remain empty.
+struct CppMetricsPath
+{
+  1:CppUnitType type
+  2:common.AstNodeId astNodeId
+  3:list<CppMetricsAstNode> astNodeIdMetrics
+  4:common.FileId fileId
+  5:list<CppMetricsModule> fileMetrics
 }
 
 service CppMetricsService
@@ -64,6 +93,20 @@ service CppMetricsService
    */
   list<CppMetricsModule> getCppMetricsForModule(
     1:common.FileId fileId)
+
+  /**
+   * This function returns all available C++ metrics
+   * (AST node-level) for a particular path.
+   */
+  list<CppAllMetricsAstNode> getCppAstNodeMetricsForPath(
+    1:string path)
+
+  /**
+   * This function returns all available C++ metrics
+   * (file-level) for a particular path.
+   */
+  list<CppAllMetricsModule> getCppFileMetricsForPath(
+    1:string path)
 
   /**
    * This function returns the names of AST node metrics.
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -87,7 +87,7 @@ service CppMetricsService
    * This function returns all available C++ metrics
    * (AST node-level) for a particular path.
    */
-  list<CppMetricsAstNodeAll> getCppAstNodeMetricsForPath(
+  map<common.AstNodeId, list<CppMetricsAstNodeSingle>> getCppAstNodeMetricsForPath(
     1:string path)
 
   /**
--- a/plugins/cpp_metrics/service/cxxmetrics.thrift
+++ b/plugins/cpp_metrics/service/cxxmetrics.thrift
@@ -35,39 +35,28 @@ struct CppModuleMetricsTypeName
   2:string name
 }
 
-struct CppMetricsAstNode
+struct CppMetricsAstNodeSingle
 {
   1:CppAstNodeMetricsType type,
   2:double value
 }
 
-struct CppAllMetricsAstNode
+struct CppMetricsAstNodeAll
 {
   1:common.AstNodeId id,
-  2:list<CppMetricsAstNode> metrics
+  2:list<CppMetricsAstNodeSingle> metrics
 }
 
-struct CppMetricsModule
+struct CppMetricsModuleSingle
 {
   1:CppModuleMetricsType type,
   2:double value
 }
 
-struct CppAllMetricsModule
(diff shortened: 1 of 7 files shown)


mcserep:
@intjftw In my view (and as discussed on the previous meetings), there are 3 type of metrics, but 2 of them are really similar, so they can be merged.
- AST node related metrics (e.g. function or class related metrics);
- File related metrics;
- Module (directory) related metrics.

File and module related metrics are basically the same, as CodeCompass identifies both of them with a file ID. I see no reason for differentiate between them, and this would only introduce code redundancy.  
**Thus, endpoints for querying ASTnode and File level metrics should be sufficient.** (The latter including support for module metrics.)

On the other hand, we discussed the additional need for a higher abstraction level endpoint which does not receive an ASTNodeID or a FileID as an input parameter, but a file / directory path instead. Then, all file and AST node related metrics under that path should be returned. So e.g. if the path is module path, I would like to get:
- all metrics for that module;
- all metrics for the files inside that module;
- all metrics for functions, classes, etc. inside those files.

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
 }
 
 struct CppMetricsAstNode
 {
-  1:CppMetricsType type,
+  1:CppAstNodeMetricsType type,
   2:double value
 }
 
+struct CppAllMetricsAstNode
```
mcserep, line 44: Putting `All` inbetween `Cpp` and `Metrics`, while the module's name is `cppmetrics` is highly odd, I suggest naming it one of the following:
- `AllCppMetricsAstNode`
- `CppMetricsAstNodeAll`
- `MultiCppMetricsAstNode`
- `CppMetricsAstNodeMulti`

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
+{
+  1:CppModuleMetricsType type,
+  2:double value
+}
+
+struct CppAllMetricsModule
+{
+  1:common.FileId id,
+  2:list<CppMetricsModule> metrics
+}
```
mcserep, line 60: Same remarks like for the `CppMetricsAstNode` and the `CppAllMetricsAstNode` types.

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
+
+// Thrift does not provide a union type,
+// the ids can remain empty.
+struct CppMetricsPath
+{
+  1:CppUnitType type
+  2:common.AstNodeId astNodeId
+  3:list<CppMetricsAstNode> astNodeIdMetrics
```
mcserep, line 68: `CppMetricsAstNode` already contains an ID + value pair. Why do we need a separate `astNodeId` field? Which AST node ID will it contain for the many AST node IDs under this path?

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
+  4:common.FileId fileId
+  5:list<CppMetricsModule> fileMetrics
```
mcserep, line 70: `CppMetricsModule` already contains an ID + value pair. Why do we need a separate `fileId` field? Which file's ID will it contain for the many files under this path?

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
+
+struct CppAllMetricsModule
+{
+  1:common.FileId id,
+  2:list<CppMetricsModule> metrics
+}
+
+// Thrift does not provide a union type,
+// the ids can remain empty.
+struct CppMetricsPath
```
mcserep, line 64: This struct is not used at all. Maybe it was just left here from a previous attempt and can be removed?

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -42,7 +88,33 @@ service CppMetricsService
...
+  /**
+   * This function returns the names of AST node metrics.
+   */
+  list<CppAstNodeMetricsTypeName> getCppAstNodeMetricsTypeNames()
+
+  /**
+   * This function returns the names of module-level metrics.
    */
-  list<CppMetricsTypeName> getCppMetricsTypeNames()
+   list<CppModuleMetricsTypeName> getCppModuleMetricsTypeNames()
```
mcserep, line 119: Indentation error. (Extra space at the beginning of the line.)

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+    typedef odb::query<model::CppAstNodeFilePath> CppAstNodeFilePathQuery;
+    typedef odb::result<model::CppAstNodeFilePath> CppAstNodeFilePathResult;
+
+    CppAstNodeFilePathResult nodes = _db->query<model::CppAstNodeFilePath>(
+      CppAstNodeFilePathQuery::LocFile::path.like(path_ + '%'));
+
+    if (nodes.empty())
+    {
+      core::InvalidInput ex;
+      ex.__set_msg("Invalid metric type for path: " + path_);
```
mcserep, line 137: Incorrect error message: it was not the metric type, which was invalid, but no metrics were found for the given path.

Also, maybe returning an empty list, instead of raising an exception would be more fitting here? I mean, a path could be valid, but maybe there are no metrics calculated for it?

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+    typedef odb::query<model::CppFileMetrics> CppFileMetricsQuery;
+    typedef odb::result<model::CppFileMetrics> CppFileMetricsResult;
+
+    FileResult descendants = _db->query<model::File>(
+      FileQuery::path.like(path_ + '%'));
+
+    if (descendants.empty())
+    {
+      core::InvalidInput ex;
+      ex.__set_msg("Invalid metric type for path: " + path_);
```
mcserep, line 179: Incorrect error message.

mcserep (changes requested):
Generally looks okay, with some refactoring required.

Will start functional testing now.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+      auto metricsQuery = _db->query<model::CppAstNodeMetrics>(
+        CppAstNodeMetricsQuery::astNodeId == node.id);
+      std::vector<CppMetricsAstNode> metrics;
+
+      CppMetricsAstNode metricsAstNode;
+      for (const auto& metric : metricsQuery)
+      {
+        metricsAstNode.type = static_cast<CppAstNodeMetricsType::type>(metric.type);
+        metricsAstNode.value = metric.value;
+      }
```
mcserep, line 152: A new `metricsAstNode` is computed, but not added to the `metrics` vector, therefore the result is always an empty list.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+      CppFileMetricsResult metricsQuery = _db->query<model::CppFileMetrics>(
+        CppFileMetricsQuery::file == file.id);
+      std::vector<CppMetricsModule> metrics;
+
+      CppMetricsModule metricsModule;
+      for (const auto& metric : metricsQuery)
+      {
+        metricsModule.type = static_cast<CppModuleMetricsType::type>(metric.type);
+        metricsModule.value = metric.value;
+      }
```
mcserep, line 194: A new `metricsModule` is computed, but not added to the `metrics` vector, therefore the result is always empty.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+{
+  _transaction([&, this]()
+  {
+    typedef odb::query<model::File> FileQuery;
+    typedef odb::result<model::File> FileResult;
+    typedef odb::query<model::CppFileMetrics> CppFileMetricsQuery;
+    typedef odb::result<model::CppFileMetrics> CppFileMetricsResult;
+
+    FileResult descendants = _db->query<model::File>(
+      FileQuery::path.like(path_ + '%'));
```
mcserep, line 174: Files with 0 zero metrics should not be returned in my opinion. If we put this restriction into the query, even performance can be boosted.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
```diff
@@ -81,6 +96,111 @@ double CppMetricsServiceHandler::getSingleCppMetricForAstNode(
...
+{
+  _transaction([&, this]()
+  {
+    typedef odb::query<model::CppAstNodeMetrics> CppAstNodeMetricsQuery;
+    typedef odb::result<model::CppAstNodeMetrics> CppAstNodeMetricsResult;
+    typedef odb::query<model::CppAstNodeFilePath> CppAstNodeFilePathQuery;
+    typedef odb::result<model::CppAstNodeFilePath> CppAstNodeFilePathResult;
+
+    CppAstNodeFilePathResult nodes = _db->query<model::CppAstNodeFilePath>(
+      CppAstNodeFilePathQuery::LocFile::path.like(path_ + '%'));
```
mcserep, line 132: AST nodes with 0 zero metrics should not be returned in my opinion. If we put this restriction into the query, even performance can be boosted.

mcserep (changes requested):
During functional testing, I discovered some further issues.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: Fixed this in 0936a195ad87688138a03977d8fcafa5ab62eef5.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: Fixed this in 0936a195ad87688138a03977d8fcafa5ab62eef5.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: I have added a quick temporal solution for this in 0936a195ad87688138a03977d8fcafa5ab62eef5.  
However, this does not modify the query, which would also add the expected performance boost, so it should be refactored.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: I have added a quick temporal solution for this in 0936a195ad87688138a03977d8fcafa5ab62eef5.  
However, this does not modify the query, which would also add the expected performance boost, so it should be refactored.

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,26 +4,72 @@ include "project/project.thrift"
...
+}
+
+struct CppModuleMetricsTypeName
 {
-  1:CppMetricsType type,
+  1:CppModuleMetricsType type,
   2:string name
 }
 
 struct CppMetricsAstNode
```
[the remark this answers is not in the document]
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c by deleteing that redundant struct.

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Fixed in e09e8ff80ab891cd8da090bdd74b5fe870ab6c1c by deleteing that redundant struct.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
intjftw: I have yet to figure out how to put a subquery into an ODB query, nevertheless, I transformed the query as you asked. I could only do it with another query - which creates some more technical debt.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
intjftw: I have yet to figure out how to put a subquery into an ODB query, nevertheless, I transformed the query as you asked. I could only do it with another query - which creates some more technical debt. So "fixed" in https://github.com/Ericsson/CodeCompass/commit/eb36b1b2b60fbf6e7909cbba169bc16ba37719ff.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
intjftw: Fixed in eb36b1b2b60fbf6e7909cbba169bc16ba37719ff.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
intjftw: Fixed in eb36b1b2b60fbf6e7909cbba169bc16ba37719ff.

plugins/cpp_metrics/service/src/cppmetricsservice.cpp
mcserep: I have performance concerns with this approach, as this will load all AST nodes into memory, which have a metric assigned. Even temporary merge into `master` (with a later fix) should be supported by a performance evaluation on e.g. CodeCompass itself.

The proper solution could be either a subquery or a JOIN statement. I guess a new ODB view will be required for that.

plugins/cpp_metrics/service/cxxmetrics.thrift
```diff
@@ -4,24 +4,61 @@ include "project/project.thrift"
 namespace cpp cc.service.cppmetrics
 namespace java cc.service.cppmetrics
 
-enum CppMetricsType
+enum CppUnitType
```
mcserep, line 7: This is not used anywhere. Should be removed?

mcserep (changes requested):
Apart from the mentioned issues, looks good, ready to merge after fix :rocket:

plugins/cpp_metrics/service/cxxmetrics.thrift
intjftw: Yes.

mcserep (approved):

PR #716, opened 2024-03-08
Create scorecard.yml

gkunz:
This PR enables the OpenSSF Scorecard action to be run periodically and on pushes to master. This will keep the Scorecard score shown in the badge up-to-date.

PR #730, opened 2024-04-01
Update Scorecard Action version

mcserep:
The Scorecard job added in #716 started to fail a week after, stating:
> error signing payload: getting key from Fulcio: verifying SCT: updating local metadata and targets: error updating to TUF remote mirror: invalid key

See e.g. https://github.com/Ericsson/CodeCompass/actions/runs/8411925535/job/23032110561 for reference.

It is discussed in https://github.com/ossf/scorecard-action/issues/997, that the Scorecard Action should be updated to v2.3.1.  
I have replaced the pinned versions with semantic version requirements, so bugfixes and other non-breaking improvements are added automatically to newer pipeline runs. We do not use hash pinning in other CI pipelines as well.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -32,12 +32,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+        uses: actions/checkout@v3
         with:
           persist-credentials: false
 
       - name: "Run analysis"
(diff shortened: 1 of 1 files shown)


mcserep:
@intjftw Can you please take a look on this and merge it? Not super important, but the CI job gets failing, and I get notifications.

intjftw (approved):

mcserep:
After the merge, the job still fails, as for the `ossf/scorecard-action` action, there is no support for semantic version tags (e.g. `v2`), like for any other GitHub Action. Instead, explicit version tags have to be used.

Fixed 8e84d84e29a0cec6cb0af9f6dcc587ea9ff34480.

PR #734, opened 2024-04-12
Implementation of C++ metrics parallelization

### LINKED ISSUE #733: Parallel parsing in C++ metrics
As the title says. Currently C++ metrics calculation runs sequentially. As the other parsers, C++ metrics parsing should be run in parallel.

dbukki:
Fixes #733

Added parallelization support to the C++ metrics parser plugin:
- Using `getFilterPathsQuery`, we can now perform the same check as what `cc::util::isRootedUnderAnyOf` used to do, but on the ODB-side. This is good because:
  - The number of records retrieved for manual processing by the client's handler is now drastically reduced.
  - The `filePath` members can now be removed from views that only needed them for this purpose. This further reduces the amount of data needed to be transferred per record as well.
- The `parallelCalcMetric` method of the metrics parser now serves as the backbone of all parallel computations.
- Existing metrics have been refactored to perform their outermost loop using the new parallelization mechanism.

Further considerations:
- [x] Buffering: Preload all records to already know the size on startup. (This is a memory overhead, but it might be worth it.)
- [x] Progress reporting: Building on the idea of buffering, report the current progress of each metric onto the log to increase interactivity. (Reporting at regular intervals in time and/or in steps.)
- [ ] Caching & Flushing: Collect all computed metrics into a single shared thread-safe container and persist everything at once at the end. This could eliminate the "one transaction for each worker" phenomenon currently observed. (Depending on the size of the overhead of opening and closing transactions, this could potentially give us a speed boost.)

### DIFF
--- a/plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
+++ b/plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
@@ -49,6 +49,32 @@ odb::query<TQueryParam> getFilterPathsQuery(
   return query;
 }
 
+template<typename TTask>
+class MetricsTasks
+{
+public:
+  typedef typename std::vector<TTask>::const_iterator TTaskIter;
+
+  const TTaskIter& begin() const { return _begin; }
+  const TTaskIter& end() const { return _end; }
+  std::size_t size() const { return _size; }
+
+  MetricsTasks(
+    const TTaskIter& begin_,
+    const TTaskIter& end_,
+    std::size_t size_
+  ) :
+    _begin(begin_),
+    _end(end_),
+    _size(size_)
+  {}
+
+private:
+  TTaskIter _begin;
+  TTaskIter _end;
+  std::size_t _size;
+};
+
 
 class CppMetricsParser : public AbstractParser
 {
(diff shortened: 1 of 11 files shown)


mcserep:
For some reason the diff contains several changes from unrelated PRs, which were already merged. This makes quite hard to review this PR, and the changes should be rebased to `master`.

mcserep:
> For some reason the diff contains several changes from unrelated PRs, which were already merged. This makes quite hard to review this PR, and the changes should be rebased to `master`.

Any insights on this @dbukki? Could you refactor this PR to `master`? I would love to merge this, but it is quite hard to review it like this.

dbukki:
> Any insights on this @dbukki? Could you refactor this PR to `master`? I would love to merge this, but it is quite hard to review it like this.

Sorry for not noticing your comment earlier. My branches must have been out of sync by the time I rebased this branch onto my latest version of master to add the parallelization of bumpy road. The funky diff behavior is probably a result of that. I synced my fork, the diff seems alright now.

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
```diff
@@ -15,12 +15,67 @@
...
+/// @tparam TQueryParam The type of database records to query.
+/// This type must represent an ODB view that has access to
+/// (i.e. is also joined with) the File table.
+/// @tparam TIter The iterator type of the filter paths.
+/// @tparam TSentinel The type of the end of the filter paths.
+/// @param begin_ The iterator referring to the first filter path.
+/// @param end_ The sentinel for the end of the filter paths.
+/// @return A query containing the disjunction of filters.
+template<typename TQueryParam, typename TIter, typename TSentinel>
+odb::query<TQueryParam> getFilterPathsQuery(
```
mcserep, line 37: This method looks quite independent from C++ metrics and should better be place in `dbutil`.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -101,170 +100,179 @@ bool CppMetricsParser::cleanupDatabase()
...
   });
 }
 
 void CppMetricsParser::lackOfCohesion()
 {
-  util::OdbTransaction {_ctx.db} ([&, this]
+  // Calculate the cohesion metric for all types on parallel threads.
+  parallelCalcMetric<model::CohesionCppRecordView>(
+    "Lack of cohesion",
+    _threadCount * 25,// number of jobs; adjust for granularity
```
mcserep, line 177: How were these magic numbers decided?

mcserep (changes requested):

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
mcserep: Refactored in 958d418e2861963e68475189fea64b54bd80f9ef.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
mcserep: Extracted them at least to const members. See fc86a64c0fa4e4299814b9d48a98378c00522309.

mcserep (approved):

PR #736, opened 2024-04-26
Incremental parsing issue fixes

### LINKED ISSUE #735: Incremental parsing fails with segmentation fault
As the name suggests, parsing anything in incremental mode (without a full re-parse) crashes CodeCompass with a segmentation fault.
The issue was first signaled in https://github.com/Ericsson/CodeCompass/pull/714 but was found to be present in the current master (https://github.com/Ericsson/CodeCompass/commit/8e84d84e29a0cec6cb0af9f6dcc587ea9ff34480) as well.

dbukki:
Fixes #735 

In accordance with the diagnosis outlined in https://github.com/Ericsson/CodeCompass/issues/735#issuecomment-2078918431 , the following fixes have been made:

1. A new `CppAstNodeMetricsFileView` db view has been added to manage the AST node ID <-> File ID association (`_astNodeIdCache`) in `CppMetricsParser::CppMetricsParser`.
2. The `.size()` call in `SourceManager::removeFile` has been replaced by a new `isSingletonResult` helper function in dbutil.

### DIFF
--- a/util/include/util/dbutil.h
+++ b/util/include/util/dbutil.h
@@ -98,20 +98,20 @@ inline std::string getDbDriver()
 }
 
 /// @brief Determines if the specified ODB query result only contains
-/// a single entity. That single entity is then stored in 'singleton_'.
+/// a single entity. That single entity is then stored in 'single_'.
 /// @tparam TEntity The type of entities in the query result.
 /// @param result_ The ODB query result in question.
-/// @param singleton_ The variable that receives the first entity (if any).
-/// @return Returns true if 'result_' only contained 'singleton_';
+/// @param single_ The variable that receives the first entity (if any).
+/// @return Returns true if 'result_' only contained 'single_';
 /// otherwise false.
 template<typename TEntity>
-bool isSingletonResult(odb::result<TEntity>& result_, TEntity& singleton_)
+bool isSingleResult(odb::result<TEntity>& result_, TEntity& single_)
 {
   auto it_b = result_.begin();
   const auto it_e = result_.end();
   if (it_b != it_e)
   {
-    singleton_ = *it_b;
+    single_ = *it_b;
     return ++it_b == it_e;
   }
   else return false;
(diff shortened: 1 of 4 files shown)


util/include/util/dbutil.h
```diff
@@ -97,6 +97,40 @@ inline std::string getDbDriver()
...
 
+/// @brief Determines if the specified ODB query result only contains
+/// a single entity. That single entity is then stored in 'singleton_'.
+/// @tparam TEntity The type of entities in the query result.
+/// @param result_ The ODB query result in question.
+/// @param singleton_ The variable that receives the first entity (if any).
+/// @return Returns true if 'result_' only contained 'singleton_';
+/// otherwise false.
+template<typename TEntity>
+bool isSingletonResult(odb::result<TEntity>& result_, TEntity& singleton_)
```
mcserep, line 108: I think *singleton* and *single result* are two different concepts. This naming is confusing for me.

mcserep (changes requested):
In case this `size() == 1` is the only conflicting code part with SQLite, then we can keep its support with your proposed solution. However, the C++ LSP service plugin also contains usages of the `size()` method in ODB query results. Probably those should be fixed, too.

It should be documented in the code why not to use the `size()` method in these cases.

util/include/util/dbutil.h
dbukki: The name originally came from the idea of a 'singleton list' as in a list with only one element.
But it's best to avoid confusion, and `isSingleResult` is also shorter. So I changed this in https://github.com/Ericsson/CodeCompass/pull/736/commits/72c6a6b441d4092364f480c32fe43cb69a23e18e

mcserep (dismissed):

mcserep:
Looks good to me! I had to resolve a merge conflict, so now my approval is not enough as the last commiter.

@intjftw Please approve it as well (review it too of course, if you would like 😄 )!

intjftw (approved):

PR #743, opened 2024-05-01
Build CodeCompass in Release mode in CI

mcserep:
Build CodeCompass in `RelWithDebInfo` mode in the CI, which results in a 15-25% better performance during the parsing.

Build CodeCompass in `Release` mode during tarball packaging to shrink the output size.


mcserep:

PR #744, opened 2024-05-01
Log the start of each C++ metric computation

mcserep:
Required for performance analysis of the C++ metric computation


mcserep:

intjftw (approved):

PR #746, opened 2024-05-02
Update Tarball generation for Ubuntu 20.04 Focal Fossa.

### LINKED ISSUE #725: Newer NodeJS versions are not compatible with Ubuntu 16.04 / SUSE 42.1
#706 upgraded NextJS to v14.1, which requires NodeJS v18. Currently the tarball build in the GitLab CI pipeline uses NodeJS v16.
As it was already mentioned in #624, NodeJS v18 is not compatible with Ubuntu 16.04, due to the GLIBC mismatch. The same stands for SUSE 42.1. Therefore the tarball building on *master* fails.

While we could downgrade to NextJS v13.5, as the security vulnerabilities mentioned in in #667 have been fixed there as well, in the long run the only considerable solution seems like dropping support for such old operating systems like Ubuntu 16.04 and SUSE 42.1.

mcserep:
NodeJS v18 is not compatible with Ubuntu 16.04, due to the GLIBC mismatch, as described in #725.

I have upgraded the GitLab CI configuration to drop support for Ubuntu 16.04 and SUSE 42.1. Instead, tarball builds will be provided for Ubuntu 20.04 only, the lowest currently supported Ubuntu LTS version.

The Gershwin Release will keep its dependency on NextJS v13 and thus still provider tarball builds for Ubuntu 16.04 and SUSE 42.1.

Fixes #725.


mcserep:
Proof of working CI pipeline:  
https://gitlab.inf.elte.hu/codecompass/CodeCompass/-/pipelines/16743

mcserep:

intjftw (approved):

PR #747, opened 2024-05-08
Added calculation for relational cohesion metric

### LINKED ISSUE #680: Relational Cohesion at Module Level for C++
*Relational Cohesion* is the average number of internal relationships per type. The relational cohesion represents the relationship that a module has to all its types.

While coupling takes into account all kinds of dependencies (e.g. types of function parameters) and examines the dependencies of different layers in a top-down direction, cohesion examines the relationships between things on the same level.

**Possible computation** (defined by [CppDepend](https://www.cppdepend.com/documentation/code-metrics#RelationalCohesion)):  
Let *R* be the number of type relationships that are internal to a module (i.e that do not connect to types outside the module). Let *N* be the number of types within the project. 

Then *H* is the relational cohesion, computed as:  
*H = (R + 1)/ N*

Note: the extra 1 in the formula prevents *H=0* when *N=1*.

**Recommendation:**   
As classes inside a module should be strongly related, the cohesion should be high. On the other hand, too high values may indicate over-coupling. A good range for relational cohesion is 1.5 - 4.0.

kmkristof:
Closes #680

### DIFF
--- a/plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
+++ b/plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
@@ -1,0 +1,111 @@
+#ifndef CC_MODEL_CPPRELATIONALCOHESION_H
+#define CC_MODEL_CPPRELATIONALCOHESION_H
+
+#include <model/cppentity.h>
+#include <model/cpprecord.h>
+#include <model/cppastnode.h>
+#include <model/cppfunction.h>
+#include <model/cppvariable.h>
+
+namespace cc
+{
+    namespace model
+    {
+
+        #pragma db view \
+        object(File)
+        struct RelationalCohesionFileView
+        {
+            #pragma db column(File::path)
+            std::string filePath;
+
+            #pragma db column(File::type)
+            std::string fileType;
+
+        };
+
+        #pragma db view \
+        object(CppRecord) \
+        object(CppAstNode : CppRecord::astNodeId == CppAstNode::id) \
+        object(File : CppAstNode::location.file) \
+        object(CppMemberType : CppMemberType::memberAstNode) 
+        struct RelationalCohesionRecordView
+        {
+            #pragma db column(CppEntity::entityHash)
+            std::size_t entityHash;
+
+            #pragma db column(CppMemberType::typeHash)
+            std::size_t typeHash;
+
+            #pragma db column(CppEntity::qualifiedName)
+            std::string qualifiedName;
+
+            #pragma db column(CppEntity::astNodeId)
+            CppAstNodeId astNodeId;
+            
+            #pragma db column(File::path)
+            std::string filePath;
+        };
+
+        #pragma db view \
+        object(CppFunction) \
+        object(CppAstNode : CppFunction::astNodeId == CppAstNode::id) \
+        object(File : CppAstNode::location.file)
+        struct RelationalCohesionFunctionView
+        {
+            #pragma db column(CppEntity::entityHash)
+            std::size_t entityHash;
+            
+            #pragma db column(CppTypedEntity::typeHash)
+            std::size_t returnType;
+
+            #pragma db column(File::path)
+            std::string filePath;
+        };
+
+        #pragma db view \
+        object(CppEntity) \
+        object(CppVariable = Parameters : CppFunction::parameters) \
+        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(File : CppAstNode::location.file)
+        struct RelationalCohessionFunctionParameterView
+        {
+            #pragma db column(Parameters::typeHash)
+            std::size_t typeHash;
+
+            #pragma db column(File::path)
+            std::string filePath;
+        };
+
+
+        #pragma db view \
+        object(CppVariable = Locals : CppFunction::locals) \
+        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(File : CppAstNode::location.file)
+        struct RelationalCohessionFunctionLocalView
+        {
+            #pragma db column(Locals::typeHash)
+            std::size_t typeHash;
+
+            #pragma db column(File::path)
+            std::string filePath;
+        };
+
+        #pragma db view \
+        object(CppVariable) \
+        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(File : CppAstNode::location.file)
+        struct RelationalCohesionVariableView
+        {
+            #pragma db column(CppVariable::typeHash)
+            std::size_t typeHash;
+
+            #pragma db column(File::path)
+            std::string filePath;
+
+        };
+        
+    }
+}
+
+#endif
--- a/plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
+++ b/plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
@@ -63,10 +63,12 @@ namespace cc
             std::string filePath;
         };
 
+        // object(CppEntity) \
+
         #pragma db view \
-        object(CppEntity) \
+        object(CppFunction) \
         object(CppVariable = Parameters : CppFunction::parameters) \
-        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(CppAstNode : CppAstNode::id == CppFunction::astNodeId) \
         object(File : CppAstNode::location.file)
         struct RelationalCohessionFunctionParameterView
         {
@@ -79,8 +81,9 @@ namespace cc
 
 
         #pragma db view \
+        object(CppFunction) \
         object(CppVariable = Locals : CppFunction::locals) \
-        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(CppAstNode : CppAstNode::id == CppFunction::astNodeId) \
         object(File : CppAstNode::location.file)
         struct RelationalCohessionFunctionLocalView
         {
@@ -93,7 +96,7 @@ namespace cc
 
         #pragma db view \
         object(CppVariable) \
-        object(CppAstNode : CppAstNode.id == CppEntity.astNodeId) \
+        object(CppAstNode : CppAstNode::id == CppEntity::astNodeId) \
         object(File : CppAstNode::location.file)
         struct RelationalCohesionVariableView
         {
--- a/plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
+++ b/plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
@@ -38,6 +38,8 @@ private:
   // Calculate the lack of cohesion between member variables
   // and member functions for every type.
   void lackOfCohesion();
+  // Calculate the cohesion within modules
+  void relationalCohesion();
 
   std::vector<std::string> _inputPaths;
   std::unordered_set<model::FileId> _fileIdCache;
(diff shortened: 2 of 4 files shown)


mcserep:
@kmkristof  The PR was marked ready, but the description says the tests are still missing. (And there are indeed no tests in the changeset.) Please add the missing unit tests.

mcserep:
@kmkristof The PR contains various commits unrelated to this PR for some reason. This makes it hard to review the PR.  
Please rebase your branch onto `master` and force push it.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -18,6 +20,7 @@
 #include <util/odbtransaction.h>
 
 #include <memory>
+#include <iostream>
```
mcserep, line 23: Why was this required? `fstream` should be enough for file management.

plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
```diff
@@ -0,0 +1,117 @@
...
+            std::size_t fileId;
+
+        };
+
+        #pragma db view \
+        object(CppRecord) \
+        object(CppAstNode : CppRecord::astNodeId == CppAstNode::id) \
+        object(File : CppAstNode::location.file) \
+        object(CppMemberType : CppMemberType::memberAstNode) 
+        struct RelationalCohesionRecordView
```
mcserep, line 35: I think the same view is used in #757. In this case a more generic name should be used and code duplication should be avoided.

plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
```diff
@@ -0,0 +1,117 @@
...
+            std::size_t entityHash;
+            
+            #pragma db column(CppTypedEntity::typeHash)
+            std::size_t typeHash;
+
+            #pragma db column(File::path)
+            std::string filePath;
+        };
+
+        // object(CppEntity) \
```
mcserep, line 69: Commented out code should be removed.

plugins/cpp_metrics/parser/include/cppmetricsparser/cppmetricsparser.h
```diff
@@ -40,8 +40,21 @@ class CppMetricsParser : public AbstractParser
   // Calculate the lack of cohesion between member variables
   // and member functions for every type.
   void lackOfCohesion();
+  // Calculate the cohesion within modules
+  void relationalCohesion();
+  // Check type relations in template parameter view.
+  // Used in relational cohesion metric.
+  template <typename T>
+  void checkTypes(
+    const std::string path,
```
mcserep, line 49: Why not const reference?

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+        {
+          relationsFoundInFile.insert(std::make_pair(item.filePath,std::unordered_set<std::uint64_t>{item.typeHash}));
+        }
+      }
+    }
+}
+
+void CppMetricsParser::relationalCohesion()
+{
+  std::unordered_set<std::string> filepaths;
```
mcserep, line 212: Should be `filePaths`, following naming guidelines.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+      std::string line;
+      while (std::getline(file,line))
+      {
+        filepaths.insert(line);
+      }
+    }
+    //if no modules list is specified every directory under the input paths is considered one
+    else
+    {
+      for (const auto &inputPath : _inputPaths)
```
mcserep, line 234:
```suggestion
      for (const auto& inputPath : _inputPaths)
```

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+      {
+        for (const model::RelationalCohesionFileView& file
+        : _ctx.db->query<model::RelationalCohesionFileView>(
+          odb::query<model::RelationalCohesionFileView>::query_columns::path.like(inputPath + "%") &&
+          odb::query<model::RelationalCohesionFileView>::query_columns::type.equal("DIR")
```
mcserep, line 239: `File::DIRECTORY_TYPE` should be used.  
https://github.com/Ericsson/CodeCompass/blob/master/model/include/model/file.h#L35

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+        ))
+        {
+          filepaths.insert(file.filePath);
+        } 
+      }
+    }
+
+    //get the file IDs for each module
+    for (auto &path : filepaths)
```
mcserep, line 248:
```suggestion
    for (const auto& path : filepaths)
```

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+
+    std::ifstream file(_modulesPath);
+
+    //read the specified module list if given
+    if (file.is_open())
+    {
+      std::string line;
+      while (std::getline(file,line))
+      {
+        filepaths.insert(line);
```
mcserep, line 228: This file is intended to contain module paths, which are directories. Naming the variable `filepaths` is misleading.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+        } 
+      }
+    }
+
+    //get the file IDs for each module
+    for (auto &path : filepaths)
+    {
+      for (const model::RelationalCohesionFileView& file
+        : _ctx.db->query<model::RelationalCohesionFileView>(
+          odb::query<model::RelationalCohesionFileView>::query_columns::path.like(path)
```
mcserep, line 252: Instead of `like`, we are looking for equality here, aren't we?

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+        ))
+        {
+          moduleIds[file.filePath] = file.fileId;
+        }
+    }
+
+    for (const auto &path : filepaths)
```
mcserep, line 259:
```suggestion
    for (const auto& path : filepaths)
```

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+      for (const model::RelationalCohesionRecordView& record
+      : _ctx.db->query<model::RelationalCohesionRecordView>(
+        odb::query<model::RelationalCohesionRecordView>::query_columns::File::path.like(path + "%")
+      ))
+      {
+        typesFound.insert(record.typeHash); //save types defined inside the module
+        typeDefinitionPaths.insert(std::make_pair(record.typeHash,record.filePath)); //save where the type is defined to avoid self relation
+      }
+      
+      std::unordered_map<std::string,std::unordered_set<std::uint64_t>> relationsFoundInFile; //store the type relations already found for each file
```
mcserep, line 274: This is what `unordered_multimap` is for. Wouldn't that simplify the code?

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
 }
 
+template <typename T>
+void CppMetricsParser::checkTypes(
+  const std::string path,
+  const std::unordered_set<std::uint64_t>& typesFound,
+  const std::unordered_map<std::uint64_t,std::string>& typeDefinitionPaths,
+  std::unordered_map<std::string,std::unordered_set<std::uint64_t>>& relationsFoundInFile,
+  int& relationsInModule
+)
```
mcserep, line 178: Remark: once different C++ code metrics have been separated into different classes, this should be refactored to use class fields instead of passing all these structures are references.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,170 @@ void CppMetricsParser::functionBumpyRoad()
...
+  std::unordered_map<std::uint64_t,std::string> typeDefinitionPaths;
+  std::unordered_set<std::uint64_t> typesFound;
+  std::unordered_map<std::string,std::uint64_t> moduleIds;
+
+  util::OdbTransaction {_ctx.db} ([&, this]
+  {
+
+    std::ifstream file(_modulesPath);
+
+    //read the specified module list if given
```
mcserep, line 222: Coding convention for comments is the following:
```
// Read the specified module list if given
```
(Note the space and the capitalization.)

Please apply this to all your in-code comments.

mcserep (changes requested):

plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
kmkristof: A common view could probably be used however the same applies to "CohesionCppRecordView" for example ("RelationalCohesionRecordView" contains this + typeHash). 

Maybe this should be looked at separately to unify the commonly used views?

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -165,6 +168,165 @@ void CppMetricsParser::functionBumpyRoad()
...
+      std::none_of(
+        relationsFoundInFile.equal_range(item.filePath).first,
+        relationsFoundInFile.equal_range(item.filePath).second,
+        [&item](const auto& pair) { return pair.second == item.typeHash; }
+      )
+    )
+    {
+      ++relationsInModule;
+      relationsFoundInFile.emplace(item.filePath, item.typeHash);
+      LOG(info) << item.filePath;
```
kmkristof, line 201: This was left here by mistake, it will be removed

kmkristof:
@mcserep I added the necessary data and definitions for unit testing however i ran into a problem. 
For example when looking at _plugins/cpp_metrics/test/sources/parser/RelationalCohesion/MultipleTypesInModuleWithRelations/RelationByFunctionParameter/relatedByFPA_ no typeHash is generated for the class itself (as it doesn't become a CppTypedEntity). 
This means that when i find a reference to this type i cannot connect it to the definition. How do you suggest i solve this issue?

plugins/cpp_metrics/model/include/model/cpprelationalcohesion.h
mcserep: You started working concurrently on a second issue before finishing this one. I don't advise to merge in duplicated code segments so close to each other in time, and then refactor them in a follow-up issue.

PR #750, opened 2024-06-18
Tag runtime postgresql Docker image on master with latest

mcserep:
For some time the `latest` image on the `master` branch is not tagged properly. (Therefore the `latest` tag is 9 months old on DockerHub.)

This PR fixes it.


intjftw (approved):
Seems okay to me.

PR #756, opened 2024-06-30
Fix ODB installation in web Docker image

### LINKED ISSUE #742: Incorrect ODB installation to webserver Docker images
Running `CodeCompass_webserver` in the `web-pgsql` image yields the following error:  
> CodeCompass_webserver: error while loading shared libraries: libodb-2.5.0-b.25.so: cannot open shared object file: No such file or directory

The `runtime-pgsql` image is not affected by the bug and works as expected.

mcserep:
Fixes #742.

The ODB installation in the web only docker images were incorrect. While fixing this I also improved the previous approach, which made a full ODB custom compilation both for the `dev`, the `runtime` and the `web` images.

With this PR the ODB library will be only compiled for the `dev` image, and then the installation directory will be copied to the `runtime` and the `web` image without recompilation. This should speed up building the `runtime` and the `web` images significantly (by 20+ minutes on my machine).

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -71,6 +71,7 @@ ENV DATABASE=sqlite \
     TEST_DB="sqlite:database=$TEST_WORKSPACE/cc_test.sqlite" \
     WITH_AUTH="plain;ldap" \
     LLVM_DIR=/usr/lib/llvm-11/cmake \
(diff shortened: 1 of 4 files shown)


intjftw (approved):
Looks good to me, let's check it in the CI, too. 🚀

PR #757, opened 2024-07-01
Added calculation for afferent coupling of types

### LINKED ISSUE #671: Afferent Coupling at Type Level for C++
*Coupling* is metric which can be computed for structural units at different levels (e.g. classes, namespaces, modules).
It measures how many other entities an entity depends on; and how many dependants it has.

The *Afferent Coupling* for a particular type is the number of types that depends directly on it.

High afferent coupling indicates that the concerned types have many responsibilities, while a zero value could indicate unused code.

kmkristof:
Closes #671

I added a calculation for the type level afferent coupling metric based on @intjftw's solution in #711.

TODO:

- [ ] tests

### DIFF
--- a/plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
+++ b/plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
@@ -103,46 +103,55 @@ bool CppMetricsParser::cleanupDatabase()
 
 void CppMetricsParser::afferentCouplingTypeLevel()
 {
-  util::OdbTransaction{_ctx.db}([&,this]
+
+  // Calculate the cohesion metric for all types on parallel threads.
+  parallelCalcMetric<model::AfferentRecordView>(
+    "Afferent coupling",
+    _threadCount * afferentCouplingPartitionMultiplier, // number of jobs; adjust for granularity
+    getFilterPathsQuery<model::AfferentRecordView>(),
+    [&, this](const MetricsTasks<model::AfferentRecordView>& tasks)
   {
-    std::set<std::uint64_t> typesFound;
-    std::unordered_map<std::uint64_t, int> typeFoundCnt;
-    std::unordered_map<std::uint64_t,std::uint64_t> astNodeIdOfType;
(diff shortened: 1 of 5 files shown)


mcserep:
@intjftw Since you have worked on efferent coupling in #711, I suggest you to review it, when this PR will be ready.

mcserep:
@kmkristof Since the development of this new metric was started after *Implementation of C++ metrics parallelization* (see #734) was merged, I suggest to implement this metric with multithread computation on the first attempt, based on the already implemented pattern.

We can consider this task as https://github.com/Ericsson/CodeCompass/labels/Level%3A%20Intermediate%20%283%29 with this added complexity.

plugins/cpp_metrics/parser/src/cppmetricsparser.cpp
```diff
@@ -98,6 +100,52 @@ bool CppMetricsParser::cleanupDatabase()
...
+        continue;
+      }
+
+      typesFound.clear();
+      for (const model::CppMemberType& member : _ctx.db->query<model::CppMemberType>(
+        odb::query<cc::model::CppMemberType>::typeHash == type.entityHash &&
+        odb::query<cc::model::CppMemberType>::kind == model::CppMemberType::Kind::Field))
+      {
+        typesFound.insert(member.memberTypeHash);
+      }
```
intjftw, line 126: My current solution is not complete. This coupling is not only defined by member fields but also by function parameters, inheritance relations etc. Please complement this calculation accordingly.

intjftw (changes requested):

intjftw (changes requested):
Testing with TinyXML2, the records in the db contain 0 for the AST node ID for afferent coupling. Another issue is that all afferent coupling records are persisted multiple times (or at least I think I'm discovering a pattern in the data, but the same series of stored metrics seem to be duplicated several times). Also, the afferent metrics are not present in the service component.

mcserep:
Superseded by #780.

PR #759, opened 2024-08-26
Webgui-new language service

wbqpk3:
The current implementation of `webgui-new` only calls the `CppService` as a Language Service. 
This is an issue when requesting information such as `getAstNodeInfoByPosition` since the `CppService` gets called even if we are not viewing C++ files.

In this patch, I added a universal `language-service` for `webgui-new`,  which decides the service to call based on file type.
Before using this `language-service`, one should call `createClient` with a file type and then it creates the appropriate Thrift connection.

I also replaced existing `CppService` calls with the new `language-service`.

For the codemirror-editor, I also added a `languageExtension` function, which selects the correct CodeMirror extension for syntax highlight.

### DIFF
--- a/webgui-new/package-lock.json
+++ b/webgui-new/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@codemirror/lang-cpp": "^6.0.2",
+        "@codemirror/lang-python": "^6.1.6",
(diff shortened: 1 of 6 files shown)


mcserep:
Thanks for refactoring this part. The code looks good, needs functional testing before getting merged.

mcserep (approved):

Issue #762, opened 2024-09-05
Webgui gitblame tooltip display bug


wbqpk3:
The old webgui has a git blame feature which displays the entire commit message as a tooltip.

Image of intended behaviour:

![image](https://github.com/user-attachments/assets/a1a946d0-1639-46cc-aea2-7bc6b06ca975)

However, when the commit message is too long, the text overflows and there is not enough space above the element to display the tooltip.

Example of this bug:
https://codecompass.net/demo/#wsid=CodeCompass&fid=13118785326581343631&select=1|1|1|1&center=gitblame
(example file is: /projects/CodeCompass/webserver/src/session.cpp)

If the bug does not occur, try to zoom into the page.

The code responsible for displaying tooltip is in file `plugins/git/webgui/js/gitBlame.js`

```js
  function registerCommitTooltip(repoId, commitId, domNode) {
    if (commitId === '')
      return;

    on(domNode, 'mouseenter', function (evt) {
      var tooltipId = repoId + commitId;
      if (!commitCache[tooltipId]) {
        var commit = model.gitservice.getCommit(repoId, commitId);
        commitCache[tooltipId] = gitUtil.createTooltip(commit);
      }

      Tooltip.show(commitCache[tooltipId], domNode, ['above']);
      on.once(domNode, 'mouseleave', function(){
        Tooltip.hide(domNode);
      })
    });
  }

```

Note the `Tooltip.show()` function and the position `above`.

The dojo documentation recommends using against it:
https://dojotoolkit.org/reference-guide/1.8/dijit/Tooltip.html#usage
> Be careful setting this parameter. A value of “above” may work fine until the user scrolls the screen so that there’s no room above the target node.

PR #763, opened 2024-09-05
Fix: Move gitblame tooltip to the left

wbqpk3:
This is a possible solution to fix: https://github.com/Ericsson/CodeCompass/issues/762

In this patch, the tooltip is moved to the left:

image

A CSS change is also needed to fix cases where the words are too long and overflow.


mcserep (approved):
Looks good :clap:

Issue #478, opened 2020-10-14
Snap packaging


mcserep:
To ease the installation procedure of CodeCompass, we shall support a package manager system, so end-users can easily install CodeCompass with a single command, without the trouble to compile or install (system-wide) its dependencies and CodeCompass itself. (Without superuser permissions installing the dependencies from the standard package manager is not even possible.)

I suggest to support the [Snap package manager](https://snapcraft.io/), as it would have multiple benefits for our use case:
 - Dependencies are bundled within the package, meaning we can include the specific and working versions and collection of dependencies.
 - Snaps are isolated and can be easily installed and removed, making it easier for end-users to try out without fearing from a troublesome removal procedure if they do not want to use CodeCompass (on that computer) further.
 - As the maintainer of the package we could easily update it, as we have an agile release cycle.
 - Snap is [supported](https://snapcraft.io/docs/installing-snapd) on all major Linux distributions.

zporky:
I support the idea, but I am afraid tar ball will be the main deployment technique for many of our customers, especially internal ones.

mcserep:
> I support the idea, but I am afraid tar ball will be the main deployment technique for many of our customers, especially internal ones.

The main problem with tar balls (in my opinion) that we have to deliver them to every operating system and architecture we plan to support. As far as I see even our internal customers use operating systems which are supported by snap. Is there no way for them to use snap?

zporky:
Every platform: Last time I created tarball for SUSE 11.4 which runs perfectly on all Ubuntu distros (likely all 64 bit Linux platforms younger than 8 years). All dependences carried by the tarball except Linux base.

Typical customer who wants to try out CodeCompass has no sudo right. Can we use snap package in that situation?

Meanwhile I strongly support the idea of a CodeCompass package. That is the "right" solution, and likely the one we suggest to use in "official" installations. But we also have to do a "-devel" package too.

mcserep:
> Typical customer who wants to try out CodeCompass has no sudo right. Can we use snap package in that situation?

I don't think so. I get your point then.

whisperity:
[...]

PR #773, opened 2024-10-24
Linux deb packaging

wbqpk3:
The goal of this patch is to allow users to easily install CodeCompass without compiling from source.

See `doc/packaging.md` for documentation.

The resulting `.deb` package is about `150MB` in size. Runtime dependencies are always included in the package (such as `ODB`), but in total these only take up about `20MB` in space uncompressed.

The package has been tested on a fresh `Ubuntu 22.04 server` image. First, the required dependencies were installed (from `doc/deps.md` guide), and a built `codecompass.deb` package. Then, it successfully parsed a C++ project and ran the webserver.
It was not needed to compile `ODB` since it's included in the `.deb` package.

It's also possible to create a Linux distribution specific package such as for `Ubuntu 22.04` which also pulls all the required dependencies specific to the distribution. The dependencies can be specified using `PACKAGE_DEPENDS` (see `doc/packaging.md`).


zporky:
Hi 
This is a great achievement. Is there an installation description somewhere?
gsd

wbqpk3:
Yes, it's documented here: https://github.com/wbqpk3/CodeCompass/blob/packaging/doc/packaging.md

mcserep:
Related to #478, but the conclusion there was to implement either a Snapcraft or an AppImage approach as a solution. I will keep this PR open for now, but #774 seems to supersede it, so we will might not need this in the future.

PR #637, opened 2023-10-02
Making codebase macOS compatible

mcserep:
In progress.

PR #775, opened 2024-11-22
LLVM 19 migration

wbqpk3:
This patch implements LLVM 19, clang-19 support. The changes are **not** backwards compatible, meaning it no longer compiles with older LLVM.

### Upgraded dependencies:

- Boost: version `1.86.0` (current Boost version 1.74.0 does not compile with clang-19)
- C++ standard upgraded to `17` (required by clang-19)

### Prerequisites
The new Boost version is not in the Ubuntu 22.04 repositories, it needs to be compiled from source.
Download from https://www.boost.org/users/download/, steps:
```
./bootstrap.sh --prefix=path/to/installation/prefix
./b2 install
```

LLVM 19 is also not in the repositories.
Download from https://apt.llvm.org/, steps:
```
wget https://apt.llvm.org/llvm.sh
chmod +x llvm.sh
sudo ./llvm.sh 19 all
```
### Compiling CodeCompass
Configure CMake with these flags:
```
-DLLVM_DIR=/usr/lib/llvm-19/cmake
-DClang_DIR=/usr/lib/cmake/clang-19 
-DCMAKE_C_COMPILER=clang-19 
-DCMAKE_CXX_COMPILER=clang++-19
```
We may encounter the following error:
```
In file included from CodeCompass/plugins/cpp/parser/src/cppparser.cpp:26:
In file included from CodeCompass/util/include/util/logutil.h:4:
In file included from /usr/local/include/boost/log/trivial.hpp:23:
In file included from /usr/local/include/boost/log/sources/severity_logger.hpp:23:
In file included from /usr/local/include/boost/log/sources/basic_logger.hpp:36:
In file included from /usr/local/include/boost/log/core/core.hpp:23:
In file included from /usr/local/include/boost/log/core/record.hpp:21:
In file included from /usr/local/include/boost/log/attributes/attribute_value_set.hpp:27:
In file included from /usr/local/include/boost/log/attributes/attribute_value.hpp:18:
In file included from /usr/local/include/boost/type_index.hpp:29:
In file included from /usr/local/include/boost/type_index/stl_type_index.hpp:34:
In file included from /usr/local/include/boost/core/demangle.hpp:32:
/usr/lib/llvm-19/include/cxxabi.h:55:1: error: functions that differ only in their return type cannot be overloaded
   50 | extern _LIBCXXABI_FUNC_VIS __cxa_exception*
      |                            ~~~~~~~~~~~~~~~~
   51 | #ifdef __wasm__
   52 | // In Wasm, a destructor returns its argument
   53 | __cxa_init_primary_exception(void* object, std::type_info* tinfo, void*(_LIBCXXABI_DTOR_FUNC* dest)(void*)) throw();
   54 | #else
   55 | __cxa_init_primary_exception(void* object, std::type_info* tinfo, void(_LIBCXXABI_DTOR_FUNC* dest)(void*)) throw();
      | ^
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/cxxabi_init_exception.h:70:7: note: previous declaration is here
   69 |       __cxa_refcounted_exception*
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~
   70 |       __cxa_init_primary_exception(void *object, std::type_info *tinfo,
      |       ^
```

This can be resolved by ~~installing LLVM-18~~ removing the following packages:
```
sudo apt purge libc++1-19 libc++-19-dev libc++abi1-19 libc++abi-19-dev
```

Finally, CodeCompass should build without errors.
I also successfully parsed the `gtest` C++ project.

### DIFF
--- a/plugins/cpp/parser/include/cppparser/filelocutil.h
+++ b/plugins/cpp/parser/include/cppparser/filelocutil.h
@@ -91,7 +91,7 @@ public:
     if (fid.isInvalid())
       return std::string();
 
-    const clang::FileEntry* fileEntry = _clangSrcMan.getFileEntryForID(fid);
+    const clang::OptionalFileEntryRef fileEntry = _clangSrcMan.getFileEntryRefForID(fid);
     if (!fileEntry)
       return std::string();
 
--- a/plugins/cpp/parser/src/cppparser.cpp
+++ b/plugins/cpp/parser/src/cppparser.cpp
@@ -232,7 +232,7 @@ model::BuildActionPtr CppParser::addBuildAction(
 
   model::BuildActionPtr buildAction(new model::BuildAction);
 
-  std::string extension = fs::extension(command_.Filename);
+  std::string extension = fs::path(command_.Filename).extension().string();
 
   buildAction->command = boost::algorithm::join(command_.CommandLine, " ");
   buildAction->type
--- a/plugins/cpp/parser/src/doccommentformatter.cpp
+++ b/plugins/cpp/parser/src/doccommentformatter.cpp
@@ -58,10 +58,10 @@ FullCommentParts::FullCommentParts(
 
     switch (child->getCommentKind())
     {
-      case Comment::NoCommentKind:
+      case CommentKind::None:
         continue;
 
-      case Comment::ParagraphCommentKind:
+      case CommentKind::ParagraphComment:
       {
         const ParagraphComment* pc = llvm::cast<ParagraphComment>(child);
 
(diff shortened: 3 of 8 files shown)


mcserep:
Support for Ubuntu 20.04 has been dropped recently (see #792) and LLVM was upgraded to v15 (see #791), as it was already implemented as part of #637.

**Now I have rebased and updated this PR accordingly to update the codebase for LLVM v19.**

I don't think we will merge this in the near future, as even Ubuntu 24.04 does not have LLVM v19 in its standard package repository and the original intention to upgrade LLVM to the newest available version was hunting a macOS specific bug, but it was not solved with LLVM v19 either.

Therefore I will close this PR now, but we can look this up once we upgrade to LLVM v19, if that makes the migration easier.

**Note: documentation and CI upgrades are missing from this PR.**

PR #782, opened 2025-03-10
Add module configuration support for C++ metrics

wbqpk3:
This patch implements `--modules` or `-m` flag for metrics calculations.

### DIFF
--- a/parser/src/parsercontext.cpp
+++ b/parser/src/parsercontext.cpp
@@ -1,3 +1,4 @@
+#include <boost/filesystem/exception.hpp>
 #include <fstream>
 
 #include <boost/filesystem.hpp>
@@ -86,7 +87,7 @@ ParserContext::ParserContext(
 
     if (!fileStream.good()) {
       LOG(error) << "Failed to open modules file: " << modulesFilePath;
-      exit(1);
+      return;
     }
 
     LOG(info) << "Processing modules file: " << modulesFilePath;
@@ -96,9 +97,9 @@ ParserContext::ParserContext(
       try {
         const fs::path p = fs::canonical(line);
         moduleDirectories.push_back(p.string());
-      } catch (...) {
+      } catch (fs::filesystem_error& err) {
         LOG(error) << "Failed to process path from modules file: " << line;
-        exit(1);
+        LOG(error) << err.what();
       }
     }
   }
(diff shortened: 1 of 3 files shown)


parser/src/parsercontext.cpp
```diff
@@ -76,6 +77,31 @@ ParserContext::ParserContext(
...
+
+  // Fill moduleDirectories vector
+  if (options.count("modules")) {
+    const std::string& modulesFilePath = options["modules"].as<std::string>();
+
+    std::ifstream fileStream(modulesFilePath);
+
+    if (!fileStream.good()) {
+      LOG(error) << "Failed to open modules file: " << modulesFilePath;
+      exit(1);
```
mcserep, line 89: This would mean quite an ungraceful halt of the parser, we usually do not want that. In this case, beside logging the error message, I suggest to ignore the module file.

parser/src/parsercontext.cpp
```diff
@@ -76,6 +77,31 @@ ParserContext::ParserContext(
...
+    LOG(info) << "Processing modules file: " << modulesFilePath;
+
+    std::string line;
+    while (std::getline(fileStream, line)) {
+      try {
+        const fs::path p = fs::canonical(line);
+        moduleDirectories.push_back(p.string());
+      } catch (...) {
+        LOG(error) << "Failed to process path from modules file: " << line;
+        exit(1);
```
mcserep, line 101: Same here, instead of an ungraceful exit, the improper line should be ignored. Also, maybe some information about the catched exception could be added to the log message, e.g. the exception message. Maybe try to catch a ` std::exception` first, and only keep the `catch (...)` as final handler.

mcserep (changes requested):

parser/src/parsercontext.cpp
wbqpk3: Fixed in https://github.com/Ericsson/CodeCompass/pull/782/commits/d0c2e9497c8e2a3164da4d58b98b7d72322ba511

parser/src/parsercontext.cpp
mcserep: I think we can omit handling the out of memory issue for `push_back()` at this point, since we are preprocessing the input parameters here. If we run out of memory at this point, we can crash the app, we don't really have any solution for the problem in the first place.

mcserep (approved):

PR #784, opened 2025-04-03
Add paging support for C++ metrics

wbqpk3:
This PR implements paged versions of existing functions in `CppMetricsService`:
- `getPagedCppAstNodeMetricsForPath`
- `getPagedCppAstNodeMetricsDetailedForPath`
- `getPagedCppFileMetricsForPath`

I also upgraded the C++ standard to 17, since I used `if constexpr` in a function template.

The Thrift `CppMetricsService` was missing JavaScript bindings, I think this is a mistake so I added that as well to `plugins/metrics/service/CMakeLists.txt`.

On a high level, the paging is implemented as follows:
1. We select `DISTINCT` `astNodeId` from table `CppAstNodeMetrics`.
2. We filter for `path` (using `LIKE`) and also for `pageSize` and `pageNumber` (using `LIMIT` and `OFFSET`).
3. Finally, we have a list of `astNodeId` with exact size as `pageSize`. We run `SELECT` on `CppAstNodeMetrics` once again, selecting all rows where the `astNodeId` is in the list (using SQL `IN` keyword).

The PR was also tested using Thrift JS bindings.

### DIFF
--- a/plugins/cpp_metrics/service/CMakeLists.txt
+++ b/plugins/cpp_metrics/service/CMakeLists.txt
@@ -19,8 +19,9 @@ add_custom_command(
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp/CppMetricsService.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp/CppMetricsService.h
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/gen-js
   COMMAND
-    ${THRIFT_EXECUTABLE} --gen cpp --gen js:jquery
+    ${THRIFT_EXECUTABLE} --gen cpp --gen js
       -o ${CMAKE_CURRENT_BINARY_DIR}
       -I ${PROJECT_SOURCE_DIR}/service
       ${CMAKE_CURRENT_SOURCE_DIR}/cxxmetrics.thrift
--- a/plugins/cpp_metrics/service/CMakeLists.txt
+++ b/plugins/cpp_metrics/service/CMakeLists.txt
@@ -20,7 +20,7 @@ add_custom_command(
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp/CppMetricsService.h
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp
   COMMAND
-    ${THRIFT_EXECUTABLE} --gen cpp
+    ${THRIFT_EXECUTABLE} --gen cpp --gen js:jquery
       -o ${CMAKE_CURRENT_BINARY_DIR}
       -I ${PROJECT_SOURCE_DIR}/service
       ${CMAKE_CURRENT_SOURCE_DIR}/cxxmetrics.thrift
--- a/Config.cmake
+++ b/Config.cmake
@@ -87,7 +87,7 @@ set(ODBFLAGS
   --default-pointer "std::shared_ptr")
 
 # Set CXX standard
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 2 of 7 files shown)


plugins/cpp_metrics/service/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ add_custom_command(
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp/CppMetricsService.h
     ${CMAKE_CURRENT_BINARY_DIR}/gen-cpp
```
mcserep, line 21: The JS output dir should be added here as well.

plugins/cpp_metrics/service/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ add_custom_command(
...
   COMMAND
-    ${THRIFT_EXECUTABLE} --gen cpp
+    ${THRIFT_EXECUTABLE} --gen cpp --gen js:jquery
```
mcserep, line 23: Usually we use `--gen js` in CodeCompass. Is there a reason for generating JQuery-based code here?

mcserep (changes requested):
Nice approach on avoiding code redundancy in the service implementation! 👏 

> The Thrift CppMetricsService was missing JavaScript bindings, I think this is a mistake so I added that as well to plugins/metrics/service/CMakeLists.txt.

I think it was simply missing, because the frontend does not support metrics yet. It can be added though.

mcserep:
@wbqpk3 The CI fails to build on Ubuntu 20.04. While we should drop support for Ubuntu 20.04 soon, now it should be checked why the build fails there, otherwise the Docker images won't be build too in the pipeline.

plugins/cpp_metrics/service/CMakeLists.txt
wbqpk3: Thanks for pointing this out! Now fixed in https://github.com/Ericsson/CodeCompass/pull/784/commits/30f6f4688bbc3ea4d221a67a0682148df9994a34

plugins/cpp_metrics/service/CMakeLists.txt
wbqpk3: I copied the Thrift JS generation code from `cpp_reparse`, where `jquery` was also used. But in this case, `--gen js` is sufficient.

wbqpk3:
@mcserep

Looking at the build logs, it seems the error is caused by the switch to the C++ 17 standard.
On Ubuntu 20.04, we install ODB from package repositories resulting in an older ODB version which has issues with C++ 17 changes.
In https://github.com/Ericsson/CodeCompass/pull/784/commits/77566fec8eff2abf4a149fbe40ac9a49e53fec72 I attempted to fix this issue by compiling ODB from source.

wbqpk3:
I just also noticed a warning message that Ubuntu 20.04 runner images on GitHub will no longer be supported after 2025-04-15.
https://github.com/actions/runner-images/issues/11101

mcserep:
> I just also noticed a warning message that Ubuntu 20.04 runner images on GitHub will no longer be supported after 2025-04-15. [actions/runner-images#11101](https://github.com/actions/runner-images/issues/11101)

Yes, and while it is not April 15th yet, CI jobs for Ubuntu 20.04 already started to skip, due to image unavailability. I have disabled the CI build for Ubuntu 20.04 in b9e67335ce73025df49bb84bc84796b2cc9b32b8, and added an issue #789 to deprecate it completely.

> Looking at the build logs, it seems the error is caused by the switch to the C++ 17 standard.
On Ubuntu 20.04, we install ODB from package repositories resulting in an older ODB version which has issues with C++ 17 changes.
In https://github.com/Ericsson/CodeCompass/commit/77566fec8eff2abf4a149fbe40ac9a49e53fec72 I attempted to fix this issue by compiling ODB from source.

Thanks for attempting to resolve it, but in this case it is not necessary anymore and simply reverting 77566fec8eff2abf4a149fbe40ac9a49e53fec72 would be better.

mcserep (approved):

PR #791, opened 2025-05-05
Upgrade LLVM to v15

mcserep:
Since we dropped Ubuntu 20.04 support recently, this PR cherry-picks LLVM15 upgrade from #637, as Ubuntu 22.04 has native support for it.


mcserep:
Resolved merge conflicts with #792.

PR #793, opened 2025-05-05
Make codebase compatible with Boost 1.85 deprecation removals

mcserep:
Cherry-picked from  #637.

### DIFF
--- a/plugins/cpp/parser/src/cppparser.cpp
+++ b/plugins/cpp/parser/src/cppparser.cpp
@@ -159,7 +159,7 @@ bool CppParser::isSourceFile(const std::string& file_) const
   const std::vector<std::string> cppExts{
     ".c", ".cc", ".cpp", ".cxx", ".o", ".so", ".a"};
 
-  std::string ext = fs::extension(file_);
+  std::string ext = fs::path(file_).extension().string();
   std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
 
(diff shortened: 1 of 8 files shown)


plugins/cpp/parser/src/cppparser.cpp
```diff
@@ -232,7 +232,7 @@ model::BuildActionPtr CppParser::addBuildAction(
 
   model::BuildActionPtr buildAction(new model::BuildAction);
 
-  std::string extension = fs::extension(command_.Filename);
+  auto extension = fs::path(command_.Filename).extension();
```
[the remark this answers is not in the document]
mcserep: Valid point.

plugins/dummy/parser/src/dummyparser.cpp
```diff
@@ -17,7 +17,7 @@ DummyParser::DummyParser(ParserContext& ctx_): AbstractParser(ctx_)
 
 bool DummyParser::accept(const std::string& path_)
 {
-  std::string ext = boost::filesystem::extension(path_);
+  auto ext = boost::filesystem::path(path_).extension();
```
mcserep, line 20:
```suggestion
  auto ext = boost::filesystem::path(path_).extension().string();
```

mcserep: Similarly,  as mentioned in https://github.com/Ericsson/CodeCompass/pull/793#discussion_r2073935439.

PR #800, opened 2025-06-25
Add option to skip C++ metrics calculations for lambdas and nested classes

barnabasdomozi:
- Changed `CppRecord` parsing: we now also store an `isLambda` and `context` attribute
- Changed `CppServiceHandler::getProperties` to display the new attributes
- Added CLI options `--cppmetrics-ignore-lambdas` and `--cppmetrics-ignore-nested-classes` to skip C++ metrics calculations in those cases
- These new CLI options are considered when calculating: `lackOfCohesion`, `efferentTypeLevel`, `afferentTypeLevel`


mcserep:
Great work! :clap:

PR #804, opened 2025-07-25
Query paginated metrics efficiently

mcserep:
Paginated metric querying introduced in #784 uses `LIMIT` + `OFFSET` SQL closures to select the appropriate AST Node IDs or File IDs.
However `DISTINCT` is also applied on the projected IDs, therefore the `OFFSET` does not provide any performance benefit: the database engine still needs to process (deduplicate) **all** offseted elements.

This PR introduces a new approach: *Keyset Pagination* (a.k.a. Seek Method); where instead of using `OFFSET`, we use a last / previous ID to start querying records "after" that.


mcserep:
CI only failed for SQLite due to #803.

Issue #612, opened 2023-07-08
Configurable PUBLIC_URL when building docker images


mcserep:
On the *https://codecompass.net* website we use the `/demo/` path prefix to serve a staging instance of the CodeCompass webserver. On the server itself, CodeCompass is executed in Docker containers, where the new frontend is built in the docker images with the `PUBLIC_URL` environment set to empty string. Therefore all static resources (JS files, images, etc.) are loaded from `https://codecompass.net/new/`, instead of `https://codecompass.net/demo/new/`, resulting in 404 errors.
(Currently it only works because some manual fixing was done.)

Since we are talking about Docker images, it would be nice not to hard code the path into it like that. Can we make it a little more dynamic, configurable at runtime?

PR #811, opened 2025-09-30
Frontend source code not removed from docker build for new web GUI

mcserep:
Properly remove application source code in case of Release build for new webGUI.


mcserep:
Should only be merged after #612 is resolved, since this makes fixing production easier until then :sweat_smile:

PR #839, opened 2026-03-18
Fix Python plugin venv loading

barnabasdomozi:
Changes:
- Updated the Python parser's dependencies to the latest version since
newer Python versions require a more recent version of the Jedi
parsing library to load the Python virtual environment (venv).

- Minor fix: failing to load a virtual environment
is a major misconfiguration issue, so the parser now exits
with and error.

- Removed PYBuiltin test case: these test cases have been flaky
for some time, especially after the library version updates.
Since whether a node is built-in is only shown in the info tree,
this feature is not critical and was implemented back then as a
nice-to-have.


mcserep (approved):
The removed PYBuiltin test cases raise a concern for me, this is not the way to get around a bug which we do not understand completely.

We can merge this, so the CI won't fail on these, but I have opened a separate ticket #843 for it for later evaluation.

PR #846, opened 2026-03-29
Remove -b flag for C# parser

### LINKED ISSUE #820: Refactor the `-b` flag of the C# plugin
Instead, the general `-i` input flag should be used for all input locations, similar to the C++ plugin.

LakatosMark1:
Refactor C# parser to extract target DLLs from .csproj and populate build graph

Resolves #820

### DIFF
--- a/plugins/csharp/parser/include/csharpparser/csharpparser.h
+++ b/plugins/csharp/parser/include/csharpparser/csharpparser.h
@@ -32,9 +32,10 @@ private:
 
   bool acceptProjectBuildPath(const std::string& buildDir_);
   bool parseProjectBuildPath(
-    const std::vector<std::string>& path_,
-    const std::string& buildPath_);
-  void addSource(const std::string& filepath_, bool error_);
+    const std::vector<std::string>& path_ //,
+    ); //const std::string& buildPath_
+  //void addSource(const std::string& filepath_, bool error_);
+  void addSource(const std::string& filepath_, const std::string& targetDll_, bool error_);
 };
   
 } // parser
--- a/plugins/csharp/parser/include/csharpparser/csharpparser.h
+++ b/plugins/csharp/parser/include/csharpparser/csharpparser.h
@@ -32,9 +32,8 @@ private:
 
   bool acceptProjectBuildPath(const std::string& buildDir_);
   bool parseProjectBuildPath(
-    const std::vector<std::string>& path_ //,
-    ); //const std::string& buildPath_
-  //void addSource(const std::string& filepath_, bool error_);
+    const std::vector<std::string>& path_ 
+    ); 
   void addSource(const std::string& filepath_, const std::string& targetDll_, bool error_);
 };
   
--- a/plugins/csharp/parser/src/csharpparser.cpp
+++ b/plugins/csharp/parser/src/csharpparser.cpp
@@ -18,6 +18,9 @@
 
 #include <memory>
 
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
 namespace cc
 {
 namespace parser
@@ -104,31 +107,34 @@ bool CsharpParser::parseProjectBuildPath(
 
   while(std::getline(log_str, line, '\n'))
   {
-    if (line[0] == '+' || line[0] == '-')
-    {
-      std::string content = line.substr(1); // We cut off the +/- sign
-      
-      // Find the line (|) that separates the file and the DLL
-      size_t separatorPos = content.find('|'); 
-      
-      if (separatorPos != std::string::npos) {
-          // If it exists, we split the text along |
-          std::string filepath = content.substr(0, separatorPos);
-          std::string targetDll = content.substr(separatorPos + 1);
-          
-          // We clean up the spaces from the beginning
-          filepath.erase(0, filepath.find_first_not_of(" \t"));
-          
-          addSource(filepath, targetDll, line[0] == '-'); 
-      }
-      else {
-          // Fallback if for some reason the DLL name was not sent by C#
-          content.erase(0, content.find_first_not_of(" \t"));
-          addSource(content, "Unknown.dll", line[0] == '-');
-      }
+    // Skip empty lines or non-JSON lines (like debug info "ParallelRun ...")
+    if (line.empty() || line[0] != '{') continue;
 
-      if (line[0] == '+') { countFull++; }
+    try 
+    {
+      std::stringstream jsonStream(line);
+      boost::property_tree::ptree pt;
+      boost::property_tree::read_json(jsonStream, pt);
+
+      bool fullyParsed = pt.get<bool>("fullyParsed");
+      std::string filepath = pt.get<std::string>("filePath");
+      std::string targetDll = pt.get<std::string>("targetDll");
+
+      // Check if it's an error based on 'fullyParsed'
+      bool isError = !fullyParsed;
+
+      addSource(filepath, targetDll, isError);
+
+      if (fullyParsed) { countFull++; }
       else { countPart++; }
+    } 
+    catch (const boost::property_tree::json_parser::json_parser_error& e) 
+    {
+      LOG(warning) << "Failed to parse JSON output from C# parser: " << e.what() << " | Line: " << line;
+    }
+    catch (const boost::property_tree::ptree_error& e)
+    {
+       LOG(warning) << "Missing expected JSON field from C# parser: " << e.what() << " | Line: " << line;
     }
   }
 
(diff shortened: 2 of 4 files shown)


plugins/csharp/parser/include/csharpparser/csharpparser.h
```diff
@@ -32,9 +32,10 @@ class CsharpParser : public AbstractParser
 
   bool acceptProjectBuildPath(const std::string& buildDir_);
   bool parseProjectBuildPath(
-    const std::vector<std::string>& path_,
-    const std::string& buildPath_);
-  void addSource(const std::string& filepath_, bool error_);
+    const std::vector<std::string>& path_ //,
+    ); //const std::string& buildPath_
+  //void addSource(const std::string& filepath_, bool error_);
```
mcserep: Remove commented out code from codebase.

plugins/csharp/parser/src/csharpparser.cpp
```diff
@@ -42,26 +38,24 @@ bool CsharpParser::parse()
...
       success = false;
     }
   
   return success;
 }
 
 bool CsharpParser::parseProjectBuildPath(
-  const std::vector<std::string>& paths_,
-  const std::string& buildPath_)
+  const std::vector<std::string>& paths_ //,
```
mcserep, line 60: Remove commented out comma.

plugins/csharp/parser/src_csharp/Program.cs
```diff
@@ -91,21 +51,67 @@ static int Main(string[] args)
...
             List<string> allFiles = new List<string>();
+            // This dictionary will remember which file belongs to which DLL
+            Dictionary<string, string> fileToTargetDll = new Dictionary<string, string>();
+
             foreach (var p in _rootDir)
             {
-                Console.WriteLine(p);
-                allFiles.AddRange(GetSourceFilesFromDir(p, ".cs"));
+                // We find all .csproj files
+                var csprojFiles = Directory.GetFiles(p, "*.csproj", SearchOption.AllDirectories);
```
mcserep: Project file processing should be solved with the proper tooling, the `Microsoft.CodeAnalysis.CSharp.Workspaces` NuGet package exists for that.

We can reuse this from #847.

plugins/csharp/parser/src_csharp/Program.cs
```diff
@@ -91,21 +51,67 @@ static int Main(string[] args)
...
             {
                 WriteLine(f);
             }
-            IEnumerable<string> assemblies = GetSourceFilesFromDir(_buildDir, ".dll");
-            IEnumerable<string> assemblies_base = assemblies;
-            if (args.Length == 5)
-                assemblies_base = GetSourceFilesFromDir(_buildDirBase, ".dll");
+
+
+            IEnumerable<string> assemblies_base = GetSourceFilesFromDir(_buildDirBase, ".dll"); //loading basic dlls
```
mcserep, line 72: Do we really need to collect the DLL paths here?

plugins/csharp/parser/src_csharp/Program.cs
```diff
@@ -183,15 +189,20 @@ private static async Task<int> ParalellRun(string csharpConnectionString, int th
...
                 WriteLine("ParallelRun " + tree.FilePath);
                 SemanticModel model = compilation.GetSemanticModel(tree);
                 var visitor = new AstVisitor(context, model, tree);
-                visitor.Visit(tree.GetCompilationUnitRoot());                
-                WriteLine((visitor.FullyParsed ? "+" : "-") + tree.FilePath);
+                visitor.Visit(tree.GetCompilationUnitRoot());      
+
+                // Find the DLL name and append a | to the filename.
+                string target = fileToTargetDll.ContainsKey(tree.FilePath) ? fileToTargetDll[tree.FilePath] : "Unknown.dll";
+                WriteLine((visitor.FullyParsed ? "+" : "-") + tree.FilePath + "|" + target);
```
mcserep: Why do we print the results in such an odd format?

mcserep (changes requested):

plugins/csharp/parser/src_csharp/Program.cs
LakatosMark1: It is necessary because without them, the parser cannot resolve basic .NET framework types. This logic was already in the previous version, , I just updated it because of  the changes to the command-line parameters.

plugins/csharp/parser/src_csharp/Program.cs
LakatosMark1: The + equals fully parsed, - partially parsed and the | separates the dll name and the filepath. The csharpparser.cpp separates them by this format. Should i change it to something else?

plugins/csharp/parser/src_csharp/Program.cs
mcserep: OK, we shall keep the current approach for now, meaning it is expected that the parsed .NET project was built previously and the build folder is an input folder, and adding all found DLLs as reference to the `CSharpCompilation` object will enable creating a complete semantic model of the parsed program.

In a later, separate issue, this could be significantly simplified. Instead of manually collecting files and DLLs, let Roslyn's `MSBuildWorkspace` open the .csproj or .sln directly — it resolves references automatically via NuGet/SDK:

```csharp
using Microsoft.CodeAnalysis.MSBuild;

var workspace = MSBuildWorkspace.Create();
var solution = await workspace.OpenSolutionAsync("MyApp.sln");

foreach (var project in solution.Projects)
{
    var compilation = await project.GetCompilationAsync();
    foreach (var tree in compilation.SyntaxTrees)
    {
        var model = compilation.GetSemanticModel(tree);
        // execute visitor
    }
}
```

This requires the `Microsoft.CodeAnalysis.Workspaces.MSBuild` NuGet package, which we already plan to use in #847.

plugins/csharp/parser/src_csharp/Program.cs
mcserep: As discussed today, instead of a custom and not so robust format, we should use a well known format for transmitting this data, e.g. JSON.

LakatosMark1:
I have made the canges requested!

plugins/csharp/parser/src/csharpparser.cpp
mcserep: @LakatosMark1 This is still a pending formatting issue.

plugins/csharp/parser/src/csharpparser.cpp
```diff
@@ -108,22 +103,38 @@ bool CsharpParser::parseProjectBuildPath(
...
   std::stringstream log_str(log.get());
-  //LOG(warning) << log_str.str();
   int countFull = 0, countPart = 0;
-  
+
   while(std::getline(log_str, line, '\n'))
   {
-    if (line[0] == '+' || line[0] == '-')
+    // Skip empty lines or non-JSON lines (like debug info "ParallelRun ...")
+    if (line.empty() || line[0] != '{') continue;
```
mcserep: I don't really get this, in a valid JSON file / content, not every line starts with a `{` char neccesearily.

plugins/csharp/parser/src_csharp/Program.cs
```diff
@@ -183,15 +158,28 @@ private static async Task<int> ParalellRun(string csharpConnectionString, int th
...
             {
                 WriteLine("ParallelRun " + tree.FilePath);
                 SemanticModel model = compilation.GetSemanticModel(tree);
                 var visitor = new AstVisitor(context, model, tree);
-                visitor.Visit(tree.GetCompilationUnitRoot());                
-                WriteLine((visitor.FullyParsed ? "+" : "-") + tree.FilePath);
+                visitor.Visit(tree.GetCompilationUnitRoot());      
+
+                // Find the DLL name and append a | to the filename.
+                string target = fileToTargetDll.ContainsKey(tree.FilePath) ? fileToTargetDll[tree.FilePath] : "Unknown.dll";
```
mcserep: As discussed on previous meeting, we shall not use a fake "Unknown.dll" filename. That could be a valid file. Use null or maybe empty string as extremal value.

mcserep (changes requested):
Looks almost ready now, some final touches and questions remain.

plugins/csharp/parser/src/csharpparser.cpp
LakatosMark1: Since it's pretty much impossible to reliably determine if a line is valid JSON on the C++ side, I was thinking about a different approach. What if we redirect all non-JSON messages (like logs and debug info) to Console.Error.WriteLine in the C# parser? This way, stdout will strictly contain the JSON payload, and we can completely remove the line filtering logic here. What do you think about this approach?

mcserep: As discussed today, the C# parser should dump its results (as JSON) into a file. A writeable path should be found, but the global TEMP folder could work for now. Onto the console output, the absolute path of this created file should be written.

Then, the C++ part of the C# parser could read that file easily.

